PR #11, opened 2020-02-07
WIP: introduce communities and community_users

pinnymz:
This branch includes:
- new Community and CommunityUser model
- refactoring `reputation`, `is_moderator`, `is_admin` out of User into CommunityUser
- RequestContext singleton to hold global request context information: includes current_user and host-mapped community
- required `community_id` on every table holding community specific data, default scope leverages RequestContext
- optional `community_id` on SiteSetting, default looks for current community, fallback to global (nil) if no specific record exists for the community
- Rails.cache namespace is expanded to be community specific, exception for persistent cached values
- default reputation removed from after_create hook, created on next request when a 'new' community is accessed
- single 'sample' Community added to seeds/fixtures, migration will backfill existing records to use this for required community_id values

All existing tests are passing at the moment, new tests still need to be written.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -50,7 +50,7 @@ class ApplicationController < ActionController::Base
     RequestContext.clear!
 
     host_name = request.raw_host_with_port   # include port to support multiple localhost instances
-    RequestContext.community = @community = Rails.cache.fetch("#{host_name}/community") do
+    RequestContext.community = @community = Rails.cache.fetch("#{host_name}/community", expires_in: 1.hour) do
       Community.find_by(host: host_name)
     end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -50,7 +50,7 @@ class ApplicationController < ActionController::Base
     RequestContext.clear!
 
     host_name = request.raw_host_with_port   # include port to support multiple localhost instances
-    RequestContext.community = @community = Rails.cache.fetch("#{host_name}/community", expires_in: 30.minutes) do
+    RequestContext.community = @community = Rails.cache.fetch("#{host_name}/community") do
       Community.find_by(host: host_name)
     end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -47,10 +47,19 @@ class ApplicationController < ActionController::Base
   private
 
   def set_globals
+    RequestContext.clear!
+
+    host_name = request.raw_host_with_port   # include port to support multiple localhost instances
+    RequestContext.community = @community = Rails.cache.fetch("#{host_name}/community", expires_in: 30.minutes) do
+      Community.find_by(host: host_name)
+    end
+
     if current_user.nil?
       Rails.logger.info 'No user signed in'
     else
       Rails.logger.info "User #{current_user.id} (#{current_user.username}) signed in"
+      RequestContext.user = current_user
+      current_user.ensure_community_user!
     end
 
     @hot_questions = Rails.cache.fetch("hot_questions", expires_in: 30.minutes) do
--- a/app/models/site_setting.rb
+++ b/app/models/site_setting.rb
@@ -6,7 +6,7 @@ class SiteSetting < ApplicationRecord
 
   scope :for_community_id, ->(community_id){ where(community_id: community_id) }
   scope :global, ->{ for_community_id(nil) }
-  scope :priority_order, -> { order(Arel::Nodes::Case.new.when(arel_table[:community_id].eq(nil), 1).else(0)) }
+  scope :priority_order, -> { order(Arel.sql('IF(site_settings.community_id IS NULL, 1, 0)')) }
 
   validates :name, uniqueness: true
 
--- a/app/models/site_setting.rb
+++ b/app/models/site_setting.rb
@@ -1,6 +1,13 @@
 # Represents a site setting. Site settings control the operation and display of most aspects of the site, such as
 # reputation awards, additional content, and site constants such as name and logo.
 class SiteSetting < ApplicationRecord
+  belongs_to :community
+  default_scope { for_community_id(RequestContext.community_id).or(global).priority_order }
+
+  scope :for_community_id, ->(community_id){ where(community_id: community_id) }
+  scope :global, ->{ for_community_id(nil) }
+  scope :priority_order, -> { order(Arel::Nodes::Case.new.when(arel_table[:community_id].eq(nil), 1).else(0)) }
+
   validates :name, uniqueness: true
 
   def self.[](name)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -18,7 +18,7 @@ class User < ApplicationRecord
   validates :username, presence: true, length: { minimum: 3 }
   validate :username_not_fake_admin
 
-  delegate :reputation, :reputation=, :is_moderator, :is_admin, to: :community_user
+  delegate :reputation, :reputation=, to: :community_user
 
   def self.list_includes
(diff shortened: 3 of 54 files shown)


pinnymz:
@ArtOfCode- I'll work on tests next, looking for feedback on the proposed changes up to this point

app/controllers/application_controller.rb
```diff
@@ -47,10 +47,19 @@ def check_your_privilege(name, post = nil, render_error = true)
   private
 
   def set_globals
+    RequestContext.clear!
+
+    host_name = request.raw_host_with_port   # include port to support multiple localhost instances
+    RequestContext.community = @community = Rails.cache.fetch("#{host_name}/community", expires_in: 30.minutes) do
```
ArtOfCode-: Could probably make this non-expiring - if the community for a particular host is going to change, it wouldn't be unreasonable to require a server restart

app/models/site_setting.rb
```diff
@@ -1,6 +1,13 @@
 # Represents a site setting. Site settings control the operation and display of most aspects of the site, such as
 # reputation awards, additional content, and site constants such as name and logo.
 class SiteSetting < ApplicationRecord
+  belongs_to :community
+  default_scope { for_community_id(RequestContext.community_id).or(global).priority_order }
+
+  scope :for_community_id, ->(community_id){ where(community_id: community_id) }
+  scope :global, ->{ for_community_id(nil) }
+  scope :priority_order, -> { order(Arel::Nodes::Case.new.when(arel_table[:community_id].eq(nil), 1).else(0)) }
```
ArtOfCode-: Might be easier to write this as raw SQL: `Arel.sql('IF(community_id IS NULL, 1, 0)')`

app/models/user.rb
```diff
@@ -79,4 +75,8 @@ def username_not_fake_admin
       end
     end
   end
+
+  def ensure_community_user!
+    community_user || create_community_user(reputation: SiteSetting['NewUserInitialRep'])
```
ArtOfCode-: Where's `create_community_user` defined for this line?

db/migrate/20200202052500_create_communities_table.rb
```diff
@@ -0,0 +1,68 @@
...
+    reversible do |dir|
+      dir.up do
+        @community = Community.create(name: 'Sample', host: 'sample.qpixel.com')
+        community_users = MigrationSafeUser.pluck(:id, :is_admin, :is_moderator, :reputation).map do |user_id, is_admin, is_moderator, reputation|
+          {community_id: @community.id, user_id: user_id, is_admin: is_admin, is_moderator: is_moderator, reputation: reputation}
+        end
+        CommunityUser.create(community_users)
+        change_table :users do |t|
+          t.remove :is_moderator
+          t.remove :is_admin
```
ArtOfCode-: Moving is_admin and is_moderator is good because it allows for per community admins/mods - but we should keep an is_admin attribute on the global user model so that we can also have global admins. We'll also need to make some changes in views and controllers so that global admins can see and change anything, but community admins can only see and change things for their community.

db/seeds.rb
```diff
@@ -2,9 +2,11 @@
 
 Rails.application.eager_load!
 
-Dir.glob(Rails.root.join('db/seeds/**/*.yml')).each do |f|
-  basename = Pathname.new(f).relative_path_from(Pathname.new(Rails.root.join('db/seeds'))).to_s
-  type = basename.gsub('.yml', '').singularize.classify.constantize
+RequestContext.community = Community.new(id: 1)
```
ArtOfCode-: This will need some more changes to make it work more widely, but I'm not entirely sure what that entails yet so we'll leave it for this PR.

app/controllers/application_controller.rb
pinnymz: Are we assuming memory cache is being used here?  If using file or Redis, it won't be removed by a restart.  I'll remove either way, just raising awareness.

app/models/user.rb
pinnymz: That's a [built in rails helper](https://guides.rubyonrails.org/association_basics.html#methods-added-by-has-one) - for creating a has_one association.

db/migrate/20200202052500_create_communities_table.rb
pinnymz: I think I'll modify that naming to is_global_moderator/is_global_admin.

app/controllers/application_controller.rb
ArtOfCode-: Aye, point - I was assuming memory cache, yes. What if we change it based on cache strategy in use, i.e. (pseudocode) `expires_in: Rails.cache.strategy.memory? ? 365.days : 1.hour`?

app/models/user.rb
ArtOfCode-: Ah, missed that there was a has_one as well as the has_many - I'd been expecting `community_users.create`. :+1:

db/migrate/20200202052500_create_communities_table.rb
ArtOfCode-: Works for me :+1:

app/controllers/application_controller.rb
pinnymz: That could work.  My personal opinion is that having a straight `1.hour` timeout should suffice.  Reasoning being that a single DB lookup per community once an hour is not going to make any discernible difference, and this approach keeps the code simple.

pinnymz:
Remaining issues here:
- Cached community expiration simple or based on cache strategy
- Tests

Regarding testing, I have a few issues here:
- Pretty much all tests are highly functional.  In my experience, this makes them much slower and considerably more brittle (it also demonstrates that a fair amount of the logic is in controllers). I'd like to start moving toward relying on model/unit tests where possible, extracting logic to service objects on the way.
- ActiveRecord::TestCase is not quite what I expected on a community project ;), any reason we shouldn't move to RSpec?

Neither of these are dealbreakers, but it would make contributing much more enjoyable!

ArtOfCode-:
> * Pretty much all tests are highly functional. In my experience, this makes them much slower and considerably more brittle (it also demonstrates that a fair amount of the logic is in controllers). I'd like to start moving toward relying on model/unit tests where possible, extracting logic to service objects on the way.

What do you mean by "service objects"? I'm not clear whether that's a pattern I've come across by a different name, or one I've not seen before.

> * ActiveRecord::TestCase is not quite what I expected on a community project ;), any reason we shouldn't move to RSpec?

It was the default :) I didn't use RSpec because I've not worked with it before, so I'm not familiar - but if you're able and you think that's better, that's fine by me. Bear in mind that I have [Coveralls](https://coveralls.io/github/ArtOfCode-/qpixel) set up for this project, and I'd like to make sure that keeps working if possible.

Those probably aren't things to do in this PR - let's get the essential tests for network support in using the current setup so that we can get this deployed, then we can look at migrating how we do tests.

pinnymz:
The most authoritative info I can find would be from Martin Fowler in describing a Service Layer. https://martinfowler.com/eaaCatalog/serviceLayer.html

Service Objects are the components that make up this layer. The idea is to allow for decoupling business logic from the controller 'traffic cops' and the lower level data model objects so neither needs to know more than it should. SRP enablers, if you will. 

The gain here is by reducing dependencies/coupling it makes testing much simpler, and therefore faster to write and run. The resulting tests are also much more stable from future breakage. I'd be glad to demonstrate if you'd like :)

ArtOfCode-:
@pinnymz I don't think I have any major objections to that. I think the only question I have is where would the service object files go?

Likewise, this isn't one to do in this PR - let's get the network support finalized and work on testing/service layer migration in separate PRs.

pinnymz:
Closing WIP, code is ready for full review

PR #14, opened 2020-02-14
Fix protocol check in update_profile

archielecroy:
is_a? returns true for subclasses. URI::HTTPS is a subclass of URI::Generic, so any URI will have 'https://' prepended.
instance_of? returns true only for that specific class


ArtOfCode-:
Shows how much I know the details of Ruby... Cheers :)

PR #17, opened 2020-02-29
Use Co-Design for QPixel and other minor improvements. 

luap42:
This PR makes QPixel use Co-Design rather than Bootstrap. There are some other minor improvements. There'll be a full change list once I'm finished. This is a Work In Progress.


app/views/posts/_image_upload.html.erb
```diff
@@ -1,8 +1,8 @@
 <%= form_tag upload_path, multipart: true, class: 'form-inline upload-form js-upload-form' do %>
   <%= hidden_field_tag :authenticity_token, form_authenticity_token %>
   <div class="form-group">
-    <%= label_tag :file, 'Insert an image' %>
-    <%= file_field_tag :file %>
-    <span class="form-text text-muted">Max file size <%= SiteSetting['MaxUploadSize'] %>.</span>
+    <%= label_tag :file, 'Insert an image', class: "form-element" %>
+    <div class="has-color-tertiary-500 has-font-size-caption">You can upload files, which are smaller than <%= SiteSetting['MaxUploadSize'] %>.</div>
```
ArtOfCode-: "Max file size" works :)

app/views/posts/_expanded.html.erb
```diff
@@ -1,147 +1,168 @@
...
-              &middot;
-            <% end %>
+          <% if is_question && post.closed %>
+            <div class="notice is-warning has-margin-2">
+              <h3 class="has-font-weight-normal">
+                <strong>closed</strong> by <%= link_to post.closed_by.username, user_path(post.closed_by) %>
+                on  <%= post.closed_at.strftime('%b %e, %Y at %H:%M') %>
+              </h3>
+              <p>
+                This question was closed, because it isn't suitable for this site.
```
ArtOfCode-: You don't need the comma here

app/views/posts/_expanded.html.erb
```diff
@@ -1,147 +1,168 @@
...
-            <a href="#" class="js-add-comment">Add a comment</a>
+  
+          <% if post.deleted %>
+            <div class="notice is-warning has-margin-2">
+              <h3 class="has-font-weight-normal">
+                <strong>deleted</strong> by  <%= link_to post.deleted_by.username, user_path(post.deleted_by) %>
+                on <%= post.deleted_at.strftime('%b %e, %Y at %H:%M') %>
+              </h3>
+              <p>
+                This post was deleted, because it violates the rules of this site. It can only be seen by users with the delete privilege.
```
ArtOfCode-: Another extra comma

app/views/questions/show.html.erb
```diff
@@ -1,17 +1,18 @@
...
-<%= tabs do %>
-  <% tab 'Score', query_url(sort: 'score'), is_active: active_search?('score') %>
-  <% tab 'Age', query_url(sort: 'age'), is_active: active_search?('created_at') %>
-<% end %>
+<h2><%= pluralize(@question.answers.where(deleted: false).count, "answer") %></h2>
+
+<div class="button-list is-gutterless has-float-right">
+  <a href="<%= query_url(sort: 'score') %>" class="button is-muted is-outlined <%= active_search?('score') ? 'is-active' : '' %>">Score</a>
+  <a href="<%= query_url(sort: 'age') %>" class="button is-muted is-outlined <%= active_search?('created_at') ? 'is-active' : '' %>">Active</a>
+</div>
```
ArtOfCode-: Here, let's change the `tabs` helper instead, so that the code in the view here can remain the same. It's defined in [tabs_helper.rb](https://github.com/codidact/qpixel/blob/master/app/helpers/tabs_helper.rb), and you can change the element classes in there.

app/views/questions/show.html.erb
```diff
@@ -1,17 +1,18 @@
...
+
+<div class="has-clear-clear"></div>
```
ArtOfCode-: This should probably be feedback on Co-Design rather than here, but still: component styles should have the clear built into them (or better, should use flexbox or grid so that they don't need it in the first place). It shouldn't be up to the calling view to add the clears in the right places.

app/views/questions/show.html.erb
luap42: Not sure about that. Co-Design has these button lists (which are intended for filtering/sorting) and it has real tabs (such as tabs on a profile page).

app/views/questions/show.html.erb
luap42: The clear is neccessary, because I floated the element a few lines above ("has-float-right"). That seems correct to me: When I add a float via a utility, I should need one to remove it again. Also, not all button lists need to be floated.

app/views/questions/show.html.erb
ArtOfCode-: Tabs feel more intuitively correct to me for this use, because clicking on them results in a full navigation to a new page, not just a client-side re-sort. Either way, though - whichever we pick, the code for it can go in `tabs_helper` and we can call `tabs` in the view.

app/views/questions/show.html.erb
ArtOfCode-: Ah, gotcha. Yeah, that makes sense. Can we make the `has-clear-clear` class work on the same element so that we don't have to add another element? Small thing, but I believe there's a way to do it.

ArtOfCode-:
Thanks for all the work you've done here. I've just set this up on my own machine for testing, and I've found a few things that need poking. Some of these may be here, others may be in Co-Design.

**1. Logo alignment**

Screenshot from 2020-03-04 10-49-30

Add a `display: flex;` to `a.header--site-name` to fix this.

**2. Header link contrast**
Contrast on the main navigation links is insufficient (contrast ratio 2.32; required AA=4.5, AAA=7.0). AA requires at most `#666`; AAA requires at most `#555`.

**3. Link boxes**
I'm talking about this effect:

Screenshot from 2020-03-04 10-56-50

This effect looks strange: the links present as simple text, but then morph into boxes on hover. General advice is to stick to one style (text or box) rather than morphing between both. I'd tend to go with simple text style there and either:

 * increase the color change to `#111` on hover, or
 * use a single-side box shadow for a "shadow-underline" type effect: reduce the padding to `padding: 0 8px` and use `box-shadow: 0 5px 5px -5px #777`.

**4. Vote buttons**
Vote buttons don't stay "active" (blue) when they're clicked: they turn blue on hover, but on successfully voting for a post they go back to being grey. The `.voted` class gets added to the parent `button` by the existing JS, if that helps you style them, or you may need to write some change logic into the voting JS.

**5. Development mode banner**
I like this. Admittedly it's not exactly a high-priority feature, but you could also add `.has-color-red-800` to the `<p>` inside it to add a little more coherence.

ArtOfCode-:
I've pushed a commit to this branch that fixes the voting buttons and JS logic - sorry, I know that's not the normal way of working, but in this case I figured it might be easier because of how nuanced the voting logic is, given that I already understand it.

luap42:
No problem. I had some other stuff to do the last days and will have for the next few days, but will probably be able to do something the next weekend. 2 and 3 will need to be fixed in Co-Design; I'll look into them.

ArtOfCode-:
A few more things:

**Questions header/sort buttons**
This thing:

Screenshot from 2020-03-04 11-27-43

Also not high priority, but I'd merge these into one row. That can be done with a wrapping `<div>` set as `display: flex; flex-direction: row;` which changes to `flex-direction: column;` responsively when there's not enough space for them both on one line.

**Responsiveness in general needs a review**
Some things in particular that I've noticed:

 * The left-hand side needs some padding when it's being responsive:

   Screenshot from 2020-03-04 11-30-51

 * Question pages aren't very responsive:

   Screenshot from 2020-03-04 11-32-33

  * Same screenshot: the menu disappears completely when the page hits responsive mode... which is not great.

I should also add that I'm happy to work on any of this if you don't have time or don't want to &mdash; and that, overall, I really like this design over the previous design.

luap42:
One more thing: Regarding the effect (3): I think I styled it like this, on order to be viable both on white background and on background images. I think the box shadow works better on background images than text color changes and it's less dependent on the actual image. (And I like the effect, but that's not what should be the deciding factor here)

ArtOfCode-:
> I think the box shadow works better on background images than text color changes and it's less dependent on the actual image.

Agree with you there, for sure. How does the shadow-underline effect I proposed work on background images? I don't have anything suitable to test with, but if you do then see what you think.

luap42:
Most of the above issues should be fixed now. The branch is now completed, except for the following, which I was not able to test:

- reports

ArtOfCode-:
@luap42 Biggest thing I can see that still needs fixing is responsiveness. Look at a question page in dev tools with a mobile screen size emulation, and it overflows the screen. That's definitely a thing that needs to be fixed before this is ready for production.

luap42:
Should be fixed now. @ArtOfCode-

ArtOfCode-:
Almost there. This should be the last round of changes.

 * Pagination breaks responsiveness. Fix by removing `display: flex` from `.pagination`.
 * Voting buttons are too big on mobile and result in posts having little space on the right-hand side. Reducing the size of the voting buttons should solve this. Font size 20px on the text and removing `.is-extremely-large` on the vote icons seems to be enough.
 * Main logo is too small on mobile. Removing the `style` attribute from the element (which should be done anyway) solves this.
 * Tags run into each other when they wrap over multiple lines. Adding a small `margin-top` and `margin-bottom` should solve this.
 * The error on /users is caused by moving user display on posts to using the `_user.html.erb` partial. Changing line 10 to `user.is_moderator` instead of `user.is_mod` fixes it.
 * This:
   
   Screenshot from 2020-03-16 11-36-01

   I'd drop the sidebar below the avatar and profile text when on mobile to fix this.
 * The "agree to relicense" checkbox should not be checked by default.

That seems to be it - once those are done we should be able to merge.

luap42:
Thanks. I'll look into them later.

However, the last point is something I'd like to reconsider:

> Remove word-break: break-all from .notice a - it's causing some weird word splitting on my screen size.

I added it, because otherwise the imported post link is sometimes overflowing all the boundaries, resulting in a vertical scrollbar. Maybe there's a better idea (maybe a specialised class for that link?), but that's a problem that needs to be resolved.

luap42:
Also, I don't know how I introduced the automatic on-value on the relicense checkbox. At least, I can't find anything in the source code for it. :)

luap42:
All the other issues are fixed.

ArtOfCode-:
Nice, looks good. The checkbox issue I can't find either and it doesn't look like it changed in this PR, so we'll figure that out later. As for word breaks, `break-word` should work, but I'll merge this in and we can change it on master.

PR #19, opened 2020-03-08
Track a db sample config file in favor of config/database.yml

luap42:
What really annoys me is, that the file config/database.yml is tracked. This requires strange behavior, whenever I want to start the server, because I need to change it to my environment, start the server and reset the file to prevent my config from being accidentally pushed.

Hence this PR replaces the file with a config/database.sample.yml file (same content as before) and gitignores it.


ArtOfCode-:
Changes look good, but seems you've also managed to catch Gemfile.lock and the executables under `bin/` in here (modes have changed from 755 to 644). Can you drop those files out of this PR so it's just the changes to `config/*` and `.gitignore`?

luap42:
@ArtOfCode- should be fixed.

PR #22, opened 2020-03-22
Different delete notice if by post author

luap42:
This fixes https://github.com/ArtOfCode-/qpixel/issues/60.

There'll be the following banner:

>The author of this post decided to delete it. It can only be seen by users with the delete privilege.

instead of the following:

> This post was deleted because it violates the rules of this site. It can only be seen by users with the delete privilege.

when the author deleted their post.


cellio:
Looks good to me.

PR #31, opened 2020-04-22
Added seed file for communities table

### LINKED ISSUE #28: Running rails db:seed during dev environment does not seed a community
When setting up an initial development environment following the instructions given on the front page, no community is created when running rails db:seed. This causes a 500 error when starting the server as no community can be loaded. 

image

image

CharlieBrumbaugh:
Should fix #28

### DIFF
--- a/db/seeds/communities.yml
+++ b/db/seeds/communities.yml
@@ -1,5 +1,2 @@
-- id: 1
-  name: Test
-  host: localhost
-  created_at: NOW()
-  updated_at: NOW()
+- name: Test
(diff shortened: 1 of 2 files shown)


db/seeds/communities.yml
```diff
@@ -0,0 +1,5 @@
+- id: 1
+  name: Test
+  host: localhost
```
ArtOfCode-, line 3: These two lines are all you need here. ID, `created_at`, and `updated_at` are all handled for you by Rails.

PR #38, opened 2020-04-26
Role grant (and revokation) UI

luap42:
This PR adds a role grant and revokation UI:

The described UI

Roles can be given and taken according to the following rules:

- community admins can grant and revoke community moderator roles.
- global admins can grant and revoke community moderators and admins and global moderator roles.
- global admins can grant, but not revoke, global admin roles.

Global admin roles can only be revoked by direct database access; this is by-design.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -80,8 +80,8 @@ class UsersController < ApplicationController
                                .to_h.select { |_, cs| cs.include?('user_id') }
                                .map do |k, _|
       k.singularize.classify.constantize
-                     rescue
-                       nil
+    rescue
+      nil
     end .compact
     needs_transfer.each do |model|
       model.where(user_id: @user.id).update_all(user_id: SiteSetting['SoftDeleteTransferUser'])
@@ -122,37 +122,35 @@ class UsersController < ApplicationController
     redirect_to edit_user_profile_path
   end
 
-
   def role_toggle
-    if params[:role] == "mod"
+    if params[:role] == 'mod'
       @user.community_user.update(is_moderator: !@user.is_moderator)
-      render json: { status: "success" }
+      render json: { status: 'success' }
       return
     end
 
     if current_user.is_global_admin
-      if params[:role] == "admin"
+      if params[:role] == 'admin'
         @user.community_user.update(is_admin: !@user.is_admin)
-        render json: { status: "success" }
+        render json: { status: 'success' }
         return
       end
 
-      if params[:role] == "mod-global"
+      if params[:role] == 'mod-global'
         @user.update(is_global_moderator: !@user.is_global_moderator)
-        render json: { status: "success" }
+        render json: { status: 'success' }
         return
       end
 
-      if params[:role] == "admin-global"
+      if params[:role] == 'admin-global'
         @user.update(is_global_admin: true)
-        render json: { status: "success" }
+        render json: { status: 'success' }
         return
       end
     end
 
-    render json: { status: "error", message: "Role not found: #{params[:role]}" }
+    render json: { status: 'error', message: "Role not found: #{params[:role]}" }
   end
-
 
   def stack_redirect
     response = Net::HTTP.post_form(URI('https://stackoverflow.com/oauth/access_token/json'),
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,8 +2,8 @@ require 'net/http'
 
 class UsersController < ApplicationController
   before_action :authenticate_user!, only: [:edit_profile, :update_profile, :stack_redirect, :transfer_se_content]
-  before_action :verify_moderator, only: [:mod, :destroy, :soft_delete]
-  before_action :set_user, only: [:show, :mod, :destroy, :soft_delete, :posts]
+  before_action :verify_moderator, only: [:mod, :destroy, :soft_delete, :role_toggle]
+  before_action :set_user, only: [:show, :mod, :destroy, :soft_delete, :posts, :role_toggle]
 
   def index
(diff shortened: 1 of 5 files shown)


ArtOfCode-:
> Global admin roles can only be revoked by direct database access; this is by-design.

Why's that? Seems to me it would make more sense to be the other way round - make granting privileges hard, but taking them away easy for emergencies. Or they could both be possible from the UI.

cellio:
Sounds good in general.  One question: what is a global moderator as distinct from a global admin?  Is this about separating direct DB/server access (admin) from community management (global mod)?

I assume that "community admin" means TL6, which we've said is a temporary trust level that enables instance admins to do things that mods can't do or aren't doing.  (So the model is: show up, get TL6, do whatever, give up TL6 -- instance admins are regular users most of the time and only take the priv when they need it.)

luap42:
> 
> 
> > Global admin roles can only be revoked by direct database access; this is by-design.
> 
> Why's that? Seems to me it would make more sense to be the other way round - make granting privileges hard, but taking them away easy for emergencies. Or they could both be possible from the UI.

My general idea was, that one can only grant and revoke privileges lower than one's own. However, this would require db access for everything related to global admins. Hence I extended my rule, so that it would be possible for global admins to grant the same set of privileges as one has, too, but this shouldn't be able to be revoked IMHO. However, this can be changed easily if you want to have it differently.



> 
> 
> Sounds good in general. One question: what is a global moderator as distinct from a global admin? Is this about separating direct DB/server access (admin) from community management (global mod)?
> 
> I assume that "community admin" means TL6, which we've said is a temporary trust level that enables instance admins to do things that mods can't do or aren't doing. (So the model is: show up, get TL6, do whatever, give up TL6 -- instance admins are regular users most of the time and only take the priv when they need it.)

I didn't change anything with the permission model, but as I can see it, there are admins and mods. Mods have moderation privileges (flags, user deletion, TBD, ...), whereas admins have these privileges and additionally some options to configure the site (site settings, closure reasons). Then there are -- as I see it - local ("community-") and global mods/admins. The former have the role only on one site (ex. Meta or Writing), whereas the latter have it on every site. Global settings (global site settings, global close reasons) can only be edited by the latter, which is one real privilege difference.

manassehkatz:
Jumping in to this a little late, so if I've totally missed something, let me know.
Comparing this to a customer's system that I am working on, there we ended up with:
* Superuser - can do pretty much anything. Very few people, but can add/change Admin user status, Customers (equivalent to Communities/Sites here). Automatically have access to all Customers.
* Admin - Similar to the Admin user as discussed here. Limited to specific Customers (in this case, Communities/Sites).
* Regular Users - anyone who isn't an Admin or Superuser. Could be an Admin for one Customer (Community) and Regular for another.
Most important part - all privileges (except Superuser) are based on Groups. So Trust Levels, Moderator types, etc. would each be a Group which would included the detailed actual privileges. Each user (except Superuser) set of privileges is the sum of all their Group/Customer combinations.

cellio:
> Mods have moderation privileges (flags, user deletion, TBD, ...), whereas admins have these privileges and additionally some options to configure the site (site settings, closure reasons).

I'd been thinking of things like closure reasons and category configuration as belonging to site moderators, but maybe that's SE-thinking and we should separate "things that can break the site" from the more routine moderation tasks?  In TL parlance, this would mean assigning site configuration to TL6, I believe.

That reminds me: all changes of trust level and changes of global privileges should be logged for auditing purposes.

At the global level, I don't see a strong reason to separate "mod" from "admin" but I'm also willing to go with it for now, with the prediction that some people will be assigned both.

luap42:
I'd say that this is all more or less temporary, unless we have real trust levels implemented.

cellio:
> I'd say that this is all more or less temporary, unless we have real trust levels implemented.

Fair enough.  If we revisit this when we do trust levels, it's fine for the code as it exists today.

app/controllers/users_controller.rb
```diff
@@ -122,6 +122,38 @@ def update_profile
...
+      end
+
+      if params[:role] == "admin-global"
+        @user.update(is_global_admin: true)
+        render json: { "status": "success" }
+        return
+      end
+    end
+
+    render json: { "status": "error", "message": "Role not found: " + params[:role] }
```
ArtOfCode-, line 153: Likewise no need to quote hash keys, and you can also make use of string interpolation here:

```suggestion
    render json: { status: "error", message: "Role not found: #{params[:role]}" }
```

app/views/users/mod.html.erb
```diff
@@ -1,12 +1,57 @@
...
+  
+      <% if @user.is_global_moderator %>
+        <button class="button is-filled js-role-grant-btn" data-toggle="mod-global" data-user="<%= @user.id %>">Revoke Global Moderator</button>
+      <% else %>
+        <button class="button is-outlined js-role-grant-btn" data-toggle="mod-global" data-user="<%= @user.id %>">Grant Gobal Moderator</button>
+      <% end %>
+  
+      <p>
+        You can also grant global admin permissions, but they can only be revoked by direct database
+        access, so be careful, to whom you give them.
```
ArtOfCode-, line 39: Too many commas!

```suggestion
        access, so be careful to whom you grant them.
```

ArtOfCode-:
CI is also failing, which'll need fixing before merge. It may just be Rubocop (i.e. code style) - run `rubocop -Da` in the qpixel root directory, and it'll fix many of the style errors for you.

luap42:
Okay. I have fixed the stuff from review. <s>CI is still failing, but it appears to be only complaining about style issues I didn't introduce.</s> Strange: the two CI issues left have been introduced by the command you told me to run (I had a clean workspace, run them, committed the changes). What should be done here?

ArtOfCode-:
I'm guessing version mismatch between your local Rubocop and the Circle version. Not a big issue, we'll merge and then I can fix it.

PR #50, opened 2020-04-28
Make communities customizable

luap42:
This PR adds various ways, which allow communities to be customized:

- Community Theme (CSS file included per-community automatically)
- Light/Dark Header switch
- Different category header colors per category (with default from site settings)

### DIFF
--- a/app/views/categories/_form.html.erb
+++ b/app/views/categories/_form.html.erb
@@ -55,5 +55,12 @@
     <span class="form-caption">Tick this box to set this category as the site homepage.</span>
   </div>
 
+  <div class="form-group">
+    <%= f.label :color_code, 'Color (in the header)', class: 'form-element' %>
+    <span class="form-caption">Select a color, which shall be used for the header of the views within this category. Leave empty for site-wide default.</span>
+    <%= f.select :color_code, options_for_select(['turqoise', 'green', 'blue', 'purple', 'gray', 'yellow', 'orange', 'pink', 'red'], selected: @category.color_code),
+                 { include_blank: true }, class: 'form-element' %>
+  </div>
+
   <%= f.submit 'Save', class: 'button is-filled' %>
 <% end %>
--- a/app/views/categories/_form.html.erb
+++ b/app/views/categories/_form.html.erb
@@ -56,7 +56,7 @@
   </div>
 
   <div class="form-group">
-    <%= f.label :color_code, 'Color (in the header)', class: 'form-element' %>
+    <%= f.label :color_code, 'Header color', class: 'form-element' %>
     <span class="form-caption">Select a color to be used for the header within this category. Leave empty for site-wide default.</span>
     <%= f.select :color_code, options_for_select(['turqoise', 'green', 'blue', 'purple', 'gray', 'yellow', 'orange', 'pink', 'red'], selected: @category.color_code),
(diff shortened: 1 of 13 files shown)


app/views/categories/_form.html.erb
```diff
@@ -55,5 +55,12 @@
     <span class="form-caption">Tick this box to set this category as the site homepage.</span>
   </div>
 
+  <div class="form-group">
+    <%= f.label :color_code, 'Color (in the header)', class: 'form-element' %>
```
ArtOfCode-, line 59: You can simplify this to just "header color"

app/views/categories/_form.html.erb
```diff
@@ -55,5 +55,12 @@
...
+    <span class="form-caption">Select a color, which shall be used for the header of the views within this category. Leave empty for site-wide default.</span>
```
ArtOfCode-, line 60: Good ole extra commas again :)

```suggestion
    <span class="form-caption">Select a color to be used for the header within this category. Leave empty for site-wide default.</span>
```

app/views/categories/_form.html.erb
```diff
@@ -55,5 +55,12 @@
...
+    <%= f.select :color_code, options_for_select(['turqoise', 'green', 'blue', 'purple', 'gray', 'yellow', 'orange', 'pink', 'red'], selected: @category.color_code),
```
ArtOfCode-, line 61: Typo in `turquoise`

db/seeds/site_settings.yml
```diff
@@ -12,6 +12,20 @@
   description: >
     The path, relative to the root of the domain, to the site's logo file. Can also be an absolute URL.
 
+- name: SiteCategoryHeaderDefaultColor
+  value: blue
+  value_type: string
+  category: SiteDetails
+  description: >
+    Select a color, which shall be used for categories, that don't define a header color and for category-neutral views. Valid are: turqoise, green, blue, purple, gray, yellow, orange, pink and red
```
ArtOfCode-, line 20: Grammar edits:

```suggestion
    Select a color to be used for categories that don't define a header color, and for category-neutral views. Valid are: turquoise, green, blue, purple, gray, yellow, orange, pink, red.
```

app/views/categories/_form.html.erb
luap42: Oh. haha. Problem: It's wrong in Co-Design. I'll fix it later there and then here.

luap42:
> Looking good so far. Might also be worth having some way to edit the community.css file from within the application, but I don't know whether that's one you intend to do in this PR or one to leave for another PR.

Uh. I'm not so sure. I'd rather defer it, because I don't know Ruby that well and don't want to accidentally open up a security hole.

ArtOfCode-:
Aye, fair enough - I can take a stab at that one another time. In theory it's just like any other controller, just writing to a file rather than the database, but there's a couple pitfalls around the file access.

app/views/categories/_form.html.erb
luap42: Uhm. I seem to have forgotten to build that change. This will be fixed in the next version of Co-Design, but I'll leave this for now (in this PR, to be fixed later)

ArtOfCode-:
Code is good, but you've got a failing CI build that needs fixing before we merge.

PR #57, opened 2020-04-30
Better error reporting tools

ArtOfCode-:
This PR introduces:

 * Nicer error pages (errors use the same design as the rest of the application now, rather than looking... weird)
 * Error logging to database for 500 ISE errors
 * Admin view for error reports.


app/controllers/errors_controller.rb
```diff
@@ -1,26 +1,25 @@
 # Provides web actions that represent errors. Rails' standard error pages are static HTML with inline CSS; by using
 # a custom error controller we get all the layouts and CSS.
 class ErrorsController < ApplicationController
-  def not_found
-    render status: 404
-  end
+  def error
+    @exception = request.env['action_dispatch.exception']
+    @status = ActionDispatch::ExceptionWrapper.new(request.env, @exception)&.status_code
+    views = {
```
cellio, line 7: Do other error codes not get reported or just not get nice names?  (I don't know what should happen here; this partial enumeration just made me wonder about others.)

app/views/errors/conflict.html.erb
```diff
@@ -1,5 +1,14 @@
...
 
-<% content_for :error_message do %>
-  <p>The information supplied by your browser created a conflicting operation on the server, and can't be processed.</p>
-<% end %>
+<h1 class="error-code has-color-red-500">409</h1>
+<h2 class="error-name has-color-red-400">conflict</h2>
+
+<p class="is-lead">Your request conflicted with another request.</p>
+
+<p>Any changes you were trying to make are unlikely to have saved - go back and try again.</p>
```
cellio, line 8: "have saved" -> "have been saved" for extra clarity

app/controllers/errors_controller.rb
ArtOfCode-: Only the 500s get reported; 403, 404, 409, 422, and 500 get nice error pages because those are the errors that come up in regular use. Any other errors _shouldn't_ come up so I haven't given them pages (and if they do, they're more likely to have been generated by nginx, where I can't control what they look like in any case).

app/views/errors/internal_server_error.html.erb
```diff
@@ -1,6 +1,37 @@
...
-  <p>The server encountered a configuration or processing error and was unable to complete your request. Please try
-    again, and if the problem persists contact the administrator.</p>
+<h1 class="error-code has-color-red-500">500</h1>
+<h2 class="error-name has-color-red-400">server error</h2>
+
+<p class="is-lead">
+  <% if Rails.env.production? %>
+    It's not your fault.
+  <% else %>
+    It's totally your fault. Fix it.
```
cellio, line 10: *laugh*  Uh, let's make sure this can never leak out to a user... with luck that is-production check does that. :-)

ArtOfCode-: If it's in development or test mode in a production setting... we already have much bigger problems than this :)

cellio:
Looks good to me just looking at the source.  I haven't seen it live.  And I see you fixed the test failures while I was writing other comments, cool.

ArtOfCode-:
It does, yes. I don't often find myself using that, which is why I don't have it in there, but I can add the console to the new error page if you find it useful.

luap42:
If that's more or less easily possible, than I'd prefer that, because I think it's helpful during development

ArtOfCode-:
Done, @luap42

PR #61, opened 2020-05-02
Licensing

ArtOfCode-:
In this PR:

 * Every post now has a license
 * Admins can define the available licenses, though a default Creative Commons set is provided
 * Categories can set a default license
 * Sites can also set a default license
 * When writing a post, the category default will be selected by default, followed by the site default if the category has no license set.
 * Licenses can be deactivated to prevent them being used on more posts, but can't be deleted to avoid license information being dropped from posts.
 * Fancy new admin dashboard

Screenshot from 2020-05-02 21-30-44

-----

Screenshot from 2020-05-02 21-34-59

-----

Screenshot from 2020-05-02 21-36-08


cellio:
Description and screenshots look good.  We should make sure that a post's license is immutable once set.  Could we set this up on the dev server to kick the tires before deploying?  (Since it touches a lot of places.)

app/models/post.rb
```diff
@@ -22,6 +23,7 @@ class Post < ApplicationRecord
   validates :body, presence: true, length: { minimum: 30, maximum: 30_000 }
   validates :doc_slug, uniqueness: { scope: [:community_id] }, if: -> { doc_slug.present? }
   validate :category_allows_post_type
+  validate :license_active
```
luap42, line 26: What happens here, when a user edits a post with an inactive license?

app/views/licenses/_license.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="widget">
+  <div class="widget--body">
+    <h3>
+      <%= license.name %>
+      <% if license.default? %>
+        <span class="badge is-tag is-filled">default</span>
+      <% end %>
+      <% unless license.enabled? %>
+        <span class="badge is-tag is-yellow is-filled">
```
luap42, line 9: I don't think this tag (disabled) should be filled. I think it looks better with "unfilled" styles:

show how this page would look with unfilled badge

PR #74, opened 2020-05-08
Category tags

ArtOfCode-:
In this PR:

 * Categories now have 'required tags' and 'topic tags'
 * Posts in a category with required tags must have at least one of those tags
 * No such requirement for topic tags
 * Both required and topic tags get special display: topic tags get an outline, required tags are filled.

### DIFF
--- a/app/views/categories/_form.html.erb
+++ b/app/views/categories/_form.html.erb
@@ -31,7 +31,7 @@
     <%= f.label :tag_set_id, 'Tag set', class: 'form-element' %>
     <span class="form-caption">Which tag set may posts in this category draw from?</span>
     <%= f.select :tag_set_id, options_for_select(TagSet.all.map { |ts|  [ts.name, ts.id] }, selected: @category.tag_set_id),
-                 { include_blank: true }, class: 'form-element' %>
+                 { include_blank: true }, class: 'form-element js-category-tag-set-select' %>
   </div>
 
   <div class="form-group">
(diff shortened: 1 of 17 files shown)


app/views/categories/_form.html.erb
```diff
@@ -90,5 +90,39 @@
...
+                                                       selected: @category.required_tags.pluck(:name)),
+                 { include_blank: true }, multiple: true, class: 'form-element js-tag-select',
+                 data: { tag_set: @category.tag_set&.id, create: 'false', use_ids: 'true' }, disabled: disabled %>
+  </div>
+
+  <div class="form-group js-category-tags-group">
+    <%= f.label :required_tag_ids, 'Topic tags', class: 'form-element' %>
+    <span class="form-caption js-tags-group-caption">
+      <span data-state="present" style="<%= @category.tag_set.nil? ? 'display: none' : '' %>">
+        Tags that will be highlighted as the most important tag on a question.
```
cellio, line 114: Could be more than one, right?  If a question uses two "topic" tags we highlight both?  (This question is about behavior more than it's about this message.)

cellio:
Your description sounds good to me.  I've been trying to think of a better way to talk about required tags (to users), because it's really that you must include one member of this set.  Any individual tag in that set isn't required; it's that you have to choose one from the group.  If I could come up with better language I'd want to use that in the UI (defining them, instructions to include one, error if you don't).  If anyone has any ideas please share.  If not we can go with this; I just have this nagging feeling...

Can we set this up on the dev server to poke at?

app/views/categories/_form.html.erb
ArtOfCode-: Correct. If there's more than one, both get highlighted.

ArtOfCode-:
Done. Dev server should be ready to poke.

ArtOfCode-:
That's rubocop - I'm going to ignore it for now and fix it later on develop.

Issue #84, opened 2020-05-18
Suggested edits


cellio:
At a certain rep/privilege level you can edit directly.  People below that level can suggest edits.

The MVP spec is pretty sparing: a post indicates that there's an edit pending, and the author gets notified, but there's nothing like review queues.  (We might consider a single page of links to all posts with pending edits, linked in the right column for people who can approve them, but even that is more than we specified.)  The backend is probably more complicated because we have to store that pending edit, prevent other pending edits while one's in flight, and do the notifications.  I don't have a sense of how hard this is but I suspect there's some significant work here.

Wireframe for suggested edit, which is for the full MVP but could devolve, for now, to just the current edit interface plus the notice at the top about it being a suggestion if it will be: https://www.figma.com/file/4oiLQnt5Au9p54fUyQWsHX/edit?node-id=0%3A1

Wireframe for reviewing a suggested edit: https://www.figma.com/file/QdUzSRBb6thmjpynDTvJUS/review-suggested-edit?node-id=0%3A1

luap42:
Yeah. I’m working on it, but I don’t have that much time right now.

cellio:
Sorry @luap42 -- lost track of that, so when I didn't see an issue tracking it I created one.  Not meant as a nag, just getting info linked together (like wireframes).

luap42:
No problem. That wasn't how I meant it. It's more a notice for others not to start working on it.

PR #85, opened 2020-05-24
Suggested Edits

luap42:
This PR adds **suggested edits** to QPixel.

Inline notification view (cannot edit while suggested edit is pending):

inline view

Pending suggested edit:

pending suggested edit

View for "decided" suggestion:

decided suggestion - rejected

**This PR does not yet contain tests; has not been checked in CI and contains one more TODO (filling suggested_edit_tags on edit creation)**

Closes: #84


cellio:
Based on the screen shots this looks great!  Somebody else should review the actual code (and yeah, what you said about tests and that TODO :-) ).

app/controllers/answers_controller.rb
```diff
@@ -30,19 +30,46 @@ def create
   end
 
   def edit
-    check_your_privilege('Edit', @answer)
   end
 
   def update
-    return unless check_your_privilege('Edit', @answer)
+    return if current_user.nil?
```
ArtOfCode-, line 36: This line is unnecessary - this action is behind a `before_action :authenticate_user!`, so it's not possible to access this without being signed in.

app/controllers/questions_controller.rb
```diff
@@ -63,21 +63,48 @@ def lottery
   end
 
   def edit
-    check_your_privilege('Edit', @question)
   end
 
   def update
-    return unless check_your_privilege('Edit', @question)
+    return if current_user.nil?
```
ArtOfCode-, line 69: Likewise unnecessary

app/controllers/suggested_edit_controller.rb
```diff
@@ -0,0 +1,79 @@
+class SuggestedEditController < ApplicationController
+    #require 'redcarpet/render_strip'
```
ArtOfCode-, line 2: You can get rid of the commented lines in here, I think?

app/models/post.rb
```diff
@@ -97,6 +97,14 @@ def answer?
     post_type_id == Answer.post_type_id
   end
 
+  def has_pending_suggested_edit?
+    !pending_suggested_edit.nil?
+  end
+
+  def pending_suggested_edit
+    SuggestedEdit.where(post_id: self.id, active: true).take
```
ArtOfCode-, line 105: Use `.any?` instead of `.take` here - it'll use a more efficient ActiveRecord query.

luap42: I think I can use it only for the [has_]pending_suggested_edit? method, though, right? (name change recommended by rubocop)

luap42:
Okay. Rubocop failures should have been fixed (there is one issue not from me and the Post controller is apparently too long, could be ignored by a `# rubocop:disable Metrics/ClassLength`, but I'm not sure, whether that's a good idea and hence leaving it for now) (pending one more commit to be deployed)

app/models/post.rb
ArtOfCode-: Yeah, that's right. Misplaced this comment, by the looks. `.any?` to check for existence. For _this_ method, you want `.last` instead of `.take`, again because it's more efficient SQL.

luap42:
Okay. This should now be ready for review. I'm still going to merge develop back into it and there is one open CI issue (overtly long Post class, but this shouldn't be something to block this PR IMO).

app/controllers/suggested_edit_controller.rb
```diff
@@ -0,0 +1,98 @@
+class SuggestedEditController < ApplicationController
+  before_action :set_suggested_edit, only: [:show, :approve, :reject]
+
+  def show
+    render layout: 'without_sidebar'
+  end
+
+  def approve
+    return unless @edit.active?
```
ArtOfCode-, line 9: `return` is a bad idea here: it'll leave the request unterminated. Use `not_found` instead to render a 404:

```suggestion
    not_found unless @edit.active?
```

Again, likewise for anywhere else you've done a `return` without rendering or 404ing.

app/controllers/suggested_edit_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+                                                  decided_by: current_user, updated_at: DateTime.now)
+      flash[:success] = 'Edit approved successfully.'
+      if @post.question?
+        render(json: { status: 'success', redirect_url: url_for(controller: :posts, action: :share_q,
+                                                                id: @post.id) }, status: 200)
+
+        return
+      elsif @post.answer?
+        render(json: { status: 'success', redirect_url: url_for(controller: :posts, action: :share_a,
+                                                        qid: @post.parent.id, id: @post.id) }, status: 200)
```
ArtOfCode-, line 32: No harm in it, but you can omit the `status: 200` here if you want since that's the default status.

app/controllers/suggested_edit_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+    else
+      render(json: { status: 'error', redirect_url: 'Cannot reject this suggested edit... Strange.' }, status: 400)
+    end
+  end
+
+  def set_suggested_edit
+    @edit = SuggestedEdit.find(params[:id])
+  end
```
ArtOfCode-, line 75: This should probably come _after_ the `private` modifier, otherwise it's technically an action and could be routed to.

ArtOfCode-:
Looks like we still have some CI failures, but once that and the review comments are done, this should be good to merge.

PR #89, opened 2020-06-03
Markdown parsers

ArtOfCode-:
Switches Markdown parser to using commonmarker on the backend, and markdown-it on the frontend. They should be much closer in terms of configuration, so the preview should be accurate more of the time. We also get superscript, subscript, proper footnotes, proper auto-linkify, and some other things.

This shouldn't cause any visible difference; this PR is here for code review and sanity-check.

### DIFF
--- a/.github/workflows/dev-deploy.yml
+++ b/.github/workflows/dev-deploy.yml
@@ -9,6 +9,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - uses: nelonoel/branch-name@v1
     - name: Set up SSH key
       run: |
         echo "${{ secrets.DEV_DEPLOY_KEY }}" > ~/deploy_key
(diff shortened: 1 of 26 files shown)


cellio:
How do we do proper superscript/subscript/footnotes?  (Is there documentation we can look at?)

ArtOfCode-:
We should write up some proper Markdown documentation at some point. Superscript and subscript should work via the HTML tags, `sup` and `sub` (and _possibly_ via `^(words)` and `_(words)`, but I wouldn't swear to that). Footnotes are `[^1]` inline, with `[^1]: text` at the bottom, much like URL references.

cellio:
Thanks.  I've been using the HTML tags and realized I have no idea what other notations might be available.

Where would somebody who wants to write up some doc on the markdown we support look for information?  Does the library we're using come with a language spec, for example?

ArtOfCode-:
Generally we should now support CommonMark; there's also good documentation for markdown-it and markdownit-footnote for specifics.

This is also up on dev for poking at now.

cellio:
Superscript and subscript on https://dev.codidact.org/a/3092/3097 looked fine in the preview but don't render right in the final post.

ArtOfCode-:
Should be fixed now.

cellio:
Confirmed; that looks right now.

PR #93, opened 2020-06-09
Articles!

ArtOfCode-:
Finally!

### Post lists
Screenshot from 2020-06-09 01-55-53

### Create a post
Screenshot from 2020-06-09 01-56-17

### Post page
Screenshot from 2020-06-09 01-56-39

Categories that only display one post type will generally not show any of the post type labels. This also means that a category's displayed post types are used as the source for the post-type selection box on the create-post page: if a category displays questions and articles but allows questions, articles, and answers, then only questions and articles will be available to select (because answers can only be created as children of questions).

Up on the dev server for poking - I haven't changed any category config, so whoever goes poking will need to change category settings as they want to.


cellio:
Nice!  I modified the Reviews category on the dev server to accept articles along with questions, and can confirm that in *that* category you get an option to choose a post type while in Q&A you don't.  I successfully posted an article and a question.

Bugs:

While logged in as https://dev.codidact.org/users/3, I am unable to vote on either the question or the answer here: https://dev.codidact.org/questions/3091 .  No error messages; it just doesn't take.  We've seen this on live sites occasionally and I think there it had something to do with imported users, maybe?

Should be fixed, but as a reminder to check when the fix is deployed: visiting https://dev.codidact.org/articles/3099 currently produces a 500.  Visiting the other (newly-created) article in the same category works fine.

Should also be fixed (check): post type is shown on post pages (but not lists) in Q&A where there is only one top-level post type.

cellio:
While compiling my previous comment I'd done some clicking around the site (nothing heavyweight) and then left it sitting in a tab.  It's been six minutes and the server has apparently crashed (or maybe hit OOM) again, returning a 504 from nginx.  The last URL I attempted to visit was https://dev.codidact.org/categories/6.  I then tried another (fresh) browser -- also fails.

cellio:
I don't know if this is a regression or was already present, but I tried to create a test user on the dev server and got an error that the user name couldn't contain an 'A' (just plain 'A', no unicode stuff):

Screen_Shot_2020-06-09_at_12 41 29_PM

When I renamed the user Bob it worked.

The problem isn't mailinator; it worked post-rename, and I also used a different mailinator address for my first test account today.

PR #96, opened 2020-06-11
Globalization

luap42:
This PR adds: 

- latest Co-Design (0.10.0)
- a better (less obnoxious-red) badge for new posts based on CD
- a /dashboard:

    screenshot of the dashboard

- global notifications

    screenshot of global notifications

### DIFF
--- a/app/views/application/dashboard.html.erb
+++ b/app/views/application/dashboard.html.erb
@@ -8,13 +8,13 @@
         <div class="widget">
             <div class="widget--header is-complex">
                 <div class="has-text-align-center has-font-weight-bold has-font-size-display">
-                    <a href="//<%= c.host %>" style="display: flex; justify-content: center; align-items: center;">
+                    <%= link_to c.host, class: "dashboard-full-center" do %>
                         <% if !logo_setting.nil? %>
-                            <img src="<%= logo_setting&.typed %>" alt="<%= c.name %>" style="height: 1.25em;">
+                            <img src="<%= logo_setting&.typed %>" alt="<%= c.name %>">
                         <% else %>
                             <%= c.name %>
                         <% end%>
-                    </a>
+                    <% end %>
                 </div>
             </div>
             <div class="widget--body h-fw-bold h-bg-tertiary-050"><i class="fa fa-file-alt"></i> Posts</div>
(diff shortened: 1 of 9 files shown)


cellio:
Nice!

Are inbox items individually marked read, or does visiting the inbox clear 'em all?

It looks like you did something in the vicinity of some mobile stuff, so could you verify that the inbox properly displays on a phone?  (I think you fixed that bug, right?  If so we don't want a regression.)

Thank you for using both styling and text (unread/read) in the inbox.  That helps with accessibility.

luap42:
> Are inbox items individually marked read, or does visiting the inbox clear 'em all?

The behavior for that should not have changed...



> It looks like you did something in the vicinity of some mobile stuff, so could you verify that the inbox properly displays on a phone? (I think you fixed that bug, right? If so we don't want a regression.)

Yes and no. I didn't fix *that* bug, because that would require changing some behaviors from the old JavaScript, what I didn't yet (would be a bit more complex). That I will do in a later PR, but it's basically on the top of my TODO list (unfortunately not of my WANT TO DO list :D).

cellio:
Ah, got it.  If you haven't fixed it yet you haven't fixed it; my memory was faulty and my only concern there was possible regressions.  Sounds good to me.  (As always, I leave actual code review to someone more capable than I.)

app/views/application/dashboard.html.erb
```diff
@@ -0,0 +1,49 @@
...
+        <div class="widget">
+            <div class="widget--header is-complex">
+                <div class="has-text-align-center has-font-weight-bold has-font-size-display">
+                    <a href="//<%= c.host %>" style="display: flex; justify-content: center; align-items: center;">
+                        <% if !logo_setting.nil? %>
+                            <img src="<%= logo_setting&.typed %>" alt="<%= c.name %>" style="height: 1.25em;">
+                        <% else %>
+                            <%= c.name %>
+                        <% end%>
+                    </a>
```
ArtOfCode-, line 17:
```suggestion
                    <%= link_to c.host do %>
                        <% if !logo_setting.nil? %>
                            <img src="<%= logo_setting&.typed %>" alt="<%= c.name %>" style="height: 1.25em;">
                        <% else %>
                            <%= c.name %>
                        <% end%>
                    <% end %>

```

and move that style attribute to a CSS file! :P

app/views/application/dashboard.html.erb
```diff
@@ -0,0 +1,49 @@
...
+                <i class="fa fa-flag"></i>
+                <a href="<%= flag_queue_url(host: c.host) %>">Flags</a>
+                (global mods only)
+                </div>
+            <div class="widget--body">
+            <% if open_flags != 0 %>
+                <div class="h-c-red-700 h-fw-bold"><span class="badge is-tag is-red is-small"><%= open_flags %></span> open flags</div>
+            <% end %>
+                <div><span class="badge is-tag is-muted is-small"><%= resolved_flags %></span> resolved flags</div>
+            </div>
```
ArtOfCode-, line 45: All of this needs one more level of indent to match up properly.

PR #113, opened 2020-07-15
Prevent accidental suspension notice

luap42:
This *should* prevent accidental display of the suspension message, when it is not selected. However, I wasn't able to test it right now, so please **verify, whether it actually works**.

### DIFF
--- a/app/controllers/mod_warning_controller.rb
+++ b/app/controllers/mod_warning_controller.rb
@@ -41,11 +41,11 @@ class ModWarningController < ApplicationController
     suspension_end = DateTime.now + suspension_duration.days
 
     @warning = ModWarning.new(author: current_user, community_user: @user.community_user,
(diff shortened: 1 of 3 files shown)


app/controllers/mod_warning_controller.rb
```diff
@@ -44,8 +44,8 @@ def create
                               body: params[:mod_warning][:body], is_suspension: params[:mod_warning][:is_suspension],
                               suspension_end: suspension_end, active: true)
     if @warning.save
-      if params[:mod_warning][:is_suspension]
-        @user.community_user.update(is_suspended: true, suspension_end: suspension_end,
+      if !!params[:mod_warning][:is_suspension]
+        @user.community_user.update(is_suspended: params[:mod_warning][:is_suspension], suspension_end: suspension_end,
```
ArtOfCode-, line 48: Should have an `!!` prepended to it too; if that's the condition, it should be the value too, so that we know it's the right type.

Issue #105, opened 2020-07-01
Sorting by age or score doesn't work in a search


sarahmccuan:
1. Go to (https://outdoors.codidact.com/posts/search)
2. Search for a common term and view results 
3. Click one of the sorting buttons (age or score)

**Expected behavior**
Posts and answers are sorted by their age or score 

**Actual behavior**
No sorting is applied 

**Screenshots**
![image](https://user-images.githubusercontent.com/35535105/86197992-9fa79c00-bb24-11ea-9d2f-9273bfd64156.png)

galegosimpatico:
norepro - looks fixed - consider closing

PR #117, opened 2020-07-19
Issue #105: sort search results by :sort instead of :search

kritixilithos:
Note: untested

Should fix issue #105 by editing what seems to be a typo

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -11,7 +11,7 @@ class SearchController < ApplicationController
                  posts.search(search_data[:search]).user_sort({ term: params[:sort], default: :search_score },
                                                               relevance: :search_score, score: :score, age: :created_at)
                else
-                 posts.user_sort({ term: params[:search], default: :score },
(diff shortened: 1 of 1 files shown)


app/controllers/search_controller.rb
```diff
@@ -8,7 +8,7 @@ def search
                        .paginate(page: params[:page], per_page: 25)
 
                if search_data[:search].present?
-                 posts.search(search_data[:search]).user_sort({ term: params[:search], default: :search_score },
+                 posts.search(search_data[:search]).user_sort({ term: params[:sort], default: :search_score },
                                                               relevance: :search_score, score: :score, age: :created_at)
                else
                  posts.user_sort({ term: params[:search], default: :score },
```
ArtOfCode-, line 14: This line needs to be changed to `params[:sort]` too, then this is good to go

kritixilithos: Done

ArtOfCode-:
Thanks!

PR #124, opened 2020-07-28
New user tour

luap42:
This PR adds a new user tour. Best to be experienced on a dev server (/cc @ArtOfCode-).

Includes cute and friendly neighborhood-Codidactyl and clippy. Texts need to be proofread.

Includes furthermore:

- Tour
- Fix of notification count bug
- Invitation to use tour on first sign-up.


app/assets/javascripts/tags.js
```diff
@@ -24,17 +24,34 @@ $(() => {
 
   $('.js-tag-select').each((i, el) => {
     const $tgt = $(el);
+    var $this;
```
ArtOfCode-, line 27: `let`

app/assets/javascripts/tour.js
```diff
@@ -0,0 +1,53 @@
+function tour_firstAnswer() {
```
ArtOfCode-, line 1: Don't mix snake_case and camelCase - use camelCase for JS, i.e. tourFirstAnswer. If you want to namespace them, wrap everything in a `tour` object, i.e.

```js
const tour = {
  firstAnswer: function () { ... }
};
```

app/controllers/tour_controller.rb
```diff
@@ -0,0 +1,26 @@
+class TourController < ApplicationController
```
ArtOfCode-, line 1: Add a `layout` call to the start of this controller so that you don't have to specify the layout for every action:

```ruby
class TourController < ApplicationController
  layout 'without_sidebar'

  def index; end
```

app/models/user.rb
```diff
@@ -185,5 +187,10 @@ def extract_ip_from(request)
     request.headers['CF-Connecting-IP'] || request.ip
   end
 
+  def send_welcome_tour_message
+    create_notification('👋 Welcome to ' + SiteSetting['SiteName'] + '! Take our tour to find out, ' \
```
ArtOfCode-, line 191: hello commas my old friend...

No comma needed after "out" :)

app/views/tour/more.html.erb
```diff
@@ -0,0 +1,147 @@
...
+                    <tr>
+                        <th>Category</th>
+                        <th>Description</th>
+                    </tr>
+                <% @header_categories.each do |cat| %>
+                    <% next unless (current_user&.trust_level || 0) >= (cat.min_view_trust_level || -1) %>
+                        <tr>
+                            <td class="h-fw-bold"><%= cat.name %></td>
+                            <td><%= raw(sanitize(render_markdown(cat.short_wiki), scrubber: scrubber)) %></td>
+                        </tr>
```
ArtOfCode-, line 37: No extra indent needed on these lines because the `next` above isn't a block.

app/views/tour/more.html.erb
```diff
@@ -0,0 +1,147 @@
...
+<div class="grid step-6 hide" id="6">
+    <div class="grid--cell codidactyl">
+        <img src="/codidactyl.png" alt="Codidactyl">
+    </div>
+    <div class="grid--cell is-flexible">
+        <div class="widget">
+            <div class="widget--body">
+                <p>Great. You are almost finished.</p>
+                <p>I'd like to point you into the direction you need to look at if you need more help.</p>
+                <p>There are three possible places, where you might want to look:</p>
```
ArtOfCode-, line 135: Might be worth adding the chat link in here too, if it's set.

app/models/user.rb
luap42: Just a few more? Please...

cellio:
Text here is cut/paste from what's running on the dev server.

> This is how the question form looks like. You can find it, by going to a category (the labels below the header) and clicking the button on the right, which should say "Ask Question".

This is the "ask question" form. You can find it by looking at the category bar (the labels below the header that say things like "Q&A" and "Meta") and clicking on the "Ask Question" button.  (In some categories this button might say something different, like "Create Challenge".)

> I have put some hints into the sidebar.

have -> will (since they're not there yet), per chat discussion.

> Let's imagine, you have a question about dragons.

Drop the comma.

> If you'd really had a question about dragons, you'd need to check first, whether...

That one too (the one after "first").

> My dragon Martha has constantly been chasing my sheep around, and I'd like to get her to stop this behavior.

Is the dragon's name going to be important?  It's folksy and all, but on some sites people might edit that out, so maybe we shouldn't do that.

> This is how the explanation could look like for your dragon Martha. 

Drop "like".  (How...could look or what...look like, but now how...look like.)

> Add a blank line to start a new paragraph.

Consider making this a paragraph of its own with "Formatting tip:" as a lead-in.  The rest is about content and this isn't; let's make that clearer.

> Add a short summary of your prior efforts and how they failed, to save the time of users answering your question if they don't need to try everything again.

Add a short summary of your prior efforts and how they failed.  Help the community help you by sharing what you already know, so they know where to start and they don't waste time suggesting things you've already tried.

(Another "how it could look like".)

> You can see how it is going to look like on the site later below the text input.

You can see a preview of how it will look below the text input.

> You might want to improve the formatting of the post. We use Markdown for that. You can also use the buttons at the top of the post editor.

Add: Each button has a tooltip.

> After writing the question body, you should give your post a title. This will be shown to other users looking for posts to answer and should summarize your question.

After writing the question body, summarize your question in a title.  This summary will be shown to other users looking for posts to answer, so make it clear and specific.

(Now that I've advanced to the next section I see you're going to cover that, but it doesn't hurt to set expectations early.  The reason you're doing it is so people will be interested in your question from amongst all the others.)

> Good question titles include neither too much nor too less information

nor too little  (not less)

> The title can, but doesn't have to be a question.

Either add a comma after "be" or remove the one after "can".  "but doesn't have to be" is a subordinate clause and gets either two commas or zero.

> For example, the title "Dragon vs Sheep" might be not good in our current question

might not be good in -> is not good for

> Tags help to categorizes posts and help to make answerers find your post.

Tags categorize posts. They're shown alongside the title, which helps people browsing the list of questions, and they can also be searched for, helping people who are experts in particular topics find questions to answer (like yours).

> Your first question is almost complete. Just check through everything once again to make sure, that there are no spelling errors and that everything looks okay.

Remove comma after "sure".

Add: Try to imagine that you don't already know about Martha and your sheep as you review your question.  Is the question clear and complete? 

I got a server error (500) when I clicked the final "post" button.

luap42:
> Is the dragon's name going to be important? It's folksy and all, but on some sites people might edit that out, so maybe we shouldn't do that.

idk. I took the question from Mithical. I can change this if it is better, but I personally don't care. :)

> Either add a comma after "be" or remove the one after "can". "but doesn't have to be" is a subordinate clause and gets either two commas or zero.

Yay! I got an extra comma!! :D

---

All but the one notice with the inclusion of the dragon name have been addressed. @ArtOfCode- said, that they'll put it on the dev server later.

cellio:
Picking up where I left off:

> Read through it and consider, whether it seems helpful or not. In most cases, you could also try, whether the solution there works.

Comma overflow again. :-)  Read through it and consider whether it helps you solve your problem. Try out the solution if it's practical.

Clicking on "add comment" under the answer reloaded the page.  We should probably intercept clicks we don't want to handle and emit some sort of "disabled during the tour" message.

> Flagging and voting (both up and down) are important to this site, because they help the community moderating posts. Good posts are promoted to the top, while bad ones are send to the bottom and possibly deleted.

I would split this up -- explain voting back when you have the person cast a vote, and explain flagging here.

I think there was some language about voting letting the author know the answer was helpful.  Add there:  Voting also helps good content be more visible. Good answers are promoted to the top, while bad ones sink to the bottom. Please help curate the community by voting responsibly.

Then, for flagging:> Flagging is important to help maintain the site.  If you see something rude, flag it.  Inappropriate content will get deleted, but somebody has to know about it first -- so flag to let moderators know.  If you see anything else that doesn't look quite right, flag it and let us know about it.

> First, there is more than just pure questions and answers. We have multiple categories (which you can navigate through in the top bar), which structure content. The categories are for different purposes. You can find those out by looking at the description shown after clicking it. 

Last sentence: Click each category to see its description, or click the "Categories" link at the top to see all of them.

> An other thing is, that this site is community-run. 

Another thing: this site is community-run.

> There is no for-profit company behind, and the organization providing the platform won't interfere with community business.

...behind it  (add "it").

> This means, that all community-governance issues need to be decided in the local meta category. 

(Comma again. :-) )

This means that all community-governance issues are decided... 

> his is a special category, which isn't about the topic of the current site, but about the community. 

Remove comma after "site".

> ...suggest/report it in the local meta category or on Codidact Meta, too.

Drop "too"; we don't want people to think they have to report things in two places.

> Did you know, that there are other communities, too?

Remove comma after "know".

> You can check them out later on codidact.com 

I would drop "later".

> I'd like to point you into the direction you need to look at if you need more help.
> 
> There are three possible places, where you might want to look:

I would collapse this down to: Looking for more information?  Here are some more places to find help:

> The help center contains all kind of useful help and policy documents...

I would drop "all kind of", but if you keep it, change "kind" to "kinds" (because "all" goes with plurals, at least for counting nouns).

> Codidact Meta is where network-wide governance questions are considered and where support questions, bug reports and feature requests can be posted.

You'll love this: it would be quite fint o add a comma after "reports".  You don't have to; this is the debate about the Oxford comma.  But you can. :-)

Nice job on this!  I know I left lots of comments, some of which probably sound nit-picky, but I don't want that to obscure the good work here.

PR #128, opened 2020-08-02
Fixing meta/277029

luap42:
This PR *should* fix https://meta.codidact.com/questions/277029, but it's untested :P


app/controllers/votes_controller.rb
```diff
@@ -13,16 +13,18 @@ def create
     recent_votes = Vote.where(created_at: 24.hours.ago..Time.now, user: current_user).count
     max_votes_per_day = SiteSetting['FreeVotes'] + (@current_user.reputation - SiteSetting['NewUserInitialRep'])
 
-    if recent_votes >= max_votes_per_day
-      vote_limit_msg = 'You have used your daily vote limit of ' + recent_votes.to_s + 'votes. Gain more reputation' \
-                       ' or come back tomorrow to continue voting.'
+    unless post&.parent&.user&.id == current_user.id
```
ArtOfCode-, line 16: `post&.parent&.user_id` instead, to avoid the user load

PR #130, opened 2020-08-07
Only Today: 50% off for One Click to New Site!

luap42:
50 % off if you buy this PR today. Will go back up tomorrow!!!

j/k. This PR includes a wizard that will ease and guide the new site process. If @ArtOfCode- has snippets for the Step II (changes to nginx config, new cert, CF, restart server), then I'll happily include them to make the whole process even more easy. Otherwise, this needs some friendly tests but is otherwise good to go.

**Step 1:**

Step 1 - Site Config

**Step 2:**

Step 2 - Proxy Config

**Step 3:**

Step 3 - Customization

**Step 4:**

Step 4 - Other open stuff


ArtOfCode-:
Nice. I'd probably leave the rest of it as-is, manual, for now - the config changes are simple but could break a lot of things if there's  a bug in any automation, and the TLS cert generation is an interactive command-line thing, which isn't easy to turn into a web page.

Other settings that I always have to set up for a new site:
 * `AdminBadgeCharacter`
 * `ModBadgeCharacter`
 * `SyntaxHighlightingEnabled`
 * `SEApiClientId`
 * `SEApiClientSecret`
 * `SEApiKey`
 * `AllowContentTransfer`
 * `AdministratorContactEmail`

It'd also be useful if we could add creating initial tags in here (for meta, that's discussion, support, feature-request, bug, status-completed, status-declined, status-review, status-planned, status-deferred: the first four are required tags, the rest are mod-only tags).

luap42:
@ArtOfCode- re the settings:
- SyntaxHighlightingEnabled will be added (I missed it, because I hadn't the latest update)
- AdminBadgeCharacter (and ModBadgeCharacter) sure. Do you want them to be auto-set or manually set?
- SEApi* - could work, but I don't know, if every site has it. If yes, we might want to copy it from any other site? (or can they be different per site?)
- AllowContentTransfer needs to be manual, right?
- AdministratorContactEmail can be auto/copied from prior sites too, right?

> It'd also be useful if we could add creating initial tags in here (for meta, that's discussion, support, feature-request, bug, status-completed, status-declined, status-review, status-planned, status-deferred: the first four are required tags, the rest are mod-only tags).

This sounds like a good idea and hopefully not too hard. I'll look into it.

ArtOfCode-:
@luap42 

* Badge characters can probably be copied from another site, likewise SE API settings and admin email - they all _can_ be different per site, but for our purposes we probably want them the same everywhere.
* Content transfer and syntax highlighting need to be manual, yes.

PR #132, opened 2020-08-12
User Abilities

luap42:
This PR is goin' to become quite huge, probably. (**Edit:** Yes.)

Adding this screenshot, so that I can add an explanation later :D

grafik

Ready for review.


app/views/users/show.html.erb
```diff
@@ -1,161 +1,237 @@
...
+      Account
+    <% end %>
+    <%= link_to edit_user_profile_path, class: "tabs--item" do %>
+      Edit
+    <% end %>
+    <%= link_to subscriptions_path, class: "tabs--item" do %>
+      Subscriptions
+    <% end %>
+  <% end %>
 </div>
```
ArtOfCode-, line 33: This... seems to have been changed back from `render 'tabs'` to having the links hardcoded in the file? You might need to merge develop into your branch, or if not - let's change this back to `render 'tabs'`.

app/views/users/show.html.erb
```diff
@@ -1,161 +1,237 @@
...
-  votes_by_type = @user.votes.group(:vote_type).count(:vote_type)
-  votes_by_post_type = @user.votes.joins(:post).group(Arel.sql('posts.post_type_id')).count(Arel.sql('posts.post_type_id'))
-%>
+    <h2>Statistics</h2>
+    <%
+      posts_by_post_type = @user.posts.undeleted.group(Arel.sql('posts.post_type_id')).count(Arel.sql('posts.post_type_id'))
+    %>
+    <table class="table is-full-width">
+      <tr>
+        <th colspan=3>
```
ArtOfCode-, line 139: Always quote HTML tag attributes.

Also applies to other occurrences in this file.

```suggestion
        <th colspan="3">
```

db/migrate/20160428151421_create_privileges.rb
```diff
@@ -1,8 +0,0 @@
-class CreatePrivileges < ActiveRecord::Migration
```
ArtOfCode-, line 1: Never delete migration files - just create a new migration removing the unwanted fields/tables. This one is _probably_ done with, but deleting migration files has the potential to make things go _badly_ wrong.

db/seeds/trust_levels.yml
```diff
@@ -0,0 +1,53 @@
+- name: Participate Generally
+  internal_id: unrestricted
+  icon: fas fa-heartbeat
```
ArtOfCode-, line 3: Can we make this `heartbeat` and insert the `fas fa-` in views automatically? That way the user/admin doesn't need to know about the internals of how icons work.

scripts/recalc_abilities_for_all_users.rb
```diff
@@ -0,0 +1,18 @@
+puts "Recalculating all the Abilities"
+puts
+
+User.unscoped.all.map do |u|
+    puts "Scope: User.Id=#{u.id}"
```
ArtOfCode-, line 5: Rubocop's not going to like your 4-space indents...

db/migrate/20160428151421_create_privileges.rb
luap42: Ah. That one seems to have been done automatically by rails when I ordered it to remove the Privilege model.

ArtOfCode-: Ah, that'd do it. I tend to delete stuff manually rather than getting Rails to do it, for reasons just like this.

app/models/community_user.rb
```diff
@@ -17,4 +17,90 @@ def suspended?
...
+    Rails.cache.fetch("privileges/#{id}/post_score", expires_in: 2.hours) do
+      good_posts = Post.where(user: user).where('score > 0.5').count
+      bad_posts = Post.where(user: user).where('score < 0.5').count
+
+      (good_posts + 1.0) / (good_posts + bad_posts + 2.0)
+    end
+  end
+
+  def edit_score
+    Rails.cache.fetch("privileges/#{id}/edit_score", expires_in: 2.hours) do
```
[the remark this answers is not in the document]
luap42: Not doing this for now, maybe later ;)

app/views/abilities/index.html.erb
```diff
@@ -0,0 +1,32 @@
+<h1>Your Abilities</h1>
+
+<p class="is-lead">Abilities show, what the community trusts you to do. You earn them by writing good posts, suggesting good edits or raising good flags.</p>
```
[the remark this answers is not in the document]
luap42: Commants?

app/views/users/show.html.erb
luap42: Yes. When I’m online (WLAN) again I need to pull and merge.

scripts/recalc_abilities_for_all_users.rb
luap42: Luckily RC doesn’t scan that file. :)

app/controllers/questions_controller.rb
```diff
@@ -115,7 +115,7 @@ def update_as_suggested_edit
   end
 
   def destroy
-    unless check_your_privilege('Delete', @question, false)
+    unless check_your_privilege('flag_curate', @question, false)
       flash[:danger] = 'You must have the Delete privilege to delete questions.'
```
cellio, line 119: Can you put the names of the privs in consts and use them in messages like these?  (a) there isn't a delete priv (though there's a priv that grants delete votes), so that might confuse users, but (b) instead of just hard-wiring it here, maybe put them all in one place so if we change the (UI) name of any of them we don't have to search the code for all uses?

app/models/user.rb
```diff
@@ -88,7 +74,7 @@ def website_domain
   end
 
   def is_moderator
-    is_global_moderator || community_user&.is_moderator || is_admin || false
+    is_global_moderator || community_user&.is_moderator || is_admin || community_user.privilege?('mod') || false
```
cellio, line 77: What's the difference between is_moderator and privilege?('mod') ?

app/views/abilities/show.html.erb
```diff
@@ -0,0 +1,110 @@
...
+<% else %>
+<div class="widget">
+    <div class="widget--body">
+        <p class="h-m-0"><i class="fas fa-fw fa-check h-c-green-600"></i> You have already earned this ability.</p>
+    </div>
+    <% if @your_ability&.is_suspended %>
+    <div class="widget--body">
+        <div class="notice is-danger">
+            <% if @your_ability.suspension_end.nil? %>
+                <p><strong>Your use of the <%= @ability.name %> ability has been suspended.</strong></p>
```
cellio, line 96: We're only showing these to the user, right?

cellio:
Question: we have the goal of letting moderators suspend users from specific activities.  Some align with abilities, but one we definitely identified is comments -- let the user participate otherwise, but block commenting 'cause that user just can't pass up a chance to argue and there's a current hot-button argument (for example).  Will mods still have the ability to prevent specific activities, or does this mean we should spin out a "comment anywhere" ability as a separate thing, even though it will have the same criteria as "participate generally"?  (Similarly, might need to block voting for cases of targeted voting but otherwise-productive participation.)

app/controllers/questions_controller.rb
ArtOfCode-: I think (correct me if I'm wrong, @luap42) these are internal names, i.e. constant - the public UI names are distinct from these.

luap42: Yes this is true. However, the error messages for privilege errors will need to be improved/updated, too. It's on my TODO list, though. :)

app/models/user.rb
luap42: is_moderator is the moderator status in the db (column on Community User), privilege?('mod') is the Moderator Ability. I try to connect those, because for compatibility with older code, it'd be quite an effort to remove the column.

app/views/abilities/show.html.erb
luap42: The ability pages can be seen for any user by any user, but the suspension information is only shown to moderators and the user.

luap42:
> Question: we have the goal of letting moderators suspend users from specific activities. Some align with abilities, but one we definitely identified is comments -- let the user participate otherwise, but block commenting 'cause that user just can't pass up a chance to argue and there's a current hot-button argument (for example). Will mods still have the ability to prevent specific activities, or does this mean we should spin out a "comment anywhere" ability as a separate thing, even though it will have the same criteria as "participate generally"? (Similarly, might need to block voting for cases of targeted voting but otherwise-productive participation.)

This PR includes the option to suspend privileges. More specific blocks could be done additionally if wanted. Splitting up abilities might work, too. It's as easy as adding a seed and changing the name in the privilege check.

cellio:
We should probably go ahead and make everything we want to be individually controllable by mods into abilities, to avoid confusion later.   (This is where the idea of a *privilege* giving you multiple *abilities* would have come in, but we didn't go that route.)  We can't easily do that for *posting* with our privs structure, but maybe that's ok -- if you want to block posts entirely then you probably also want to block edits, comments, etc, so just suspend the user.  We should discuss this more to make sure we're not coding ourselves into a corner.

app/views/categories/_form.html.erb
```diff
@@ -40,7 +40,8 @@
       Users must have at least this trust level to post in this category. Leave blank to allow
       anyone to post.
     </span>
-    <%= f.number_field :min_trust_level, class: 'form-element', min: 0, max: 10, step: 1 %>
+    <%= f.select :min_trust_level, options_for_select([['everyone', 0], ['anyone with an user account', 1], ['all but new users', 2], ['veteran users', 3], ['moderators only', 4], ['staff only', 5]], selected: @category.min_trust_level),
```
cellio, line 43: "with a user account" (not "an")

app/views/categories/_form.html.erb
```diff
@@ -49,7 +50,8 @@
       Users under this trust level cannot see that this category exists. Leave blank to leave the category public.
       This can be different (higher or lower) than the minimum posting trust level.
     </span>
-    <%= f.number_field :min_view_trust_level, class: 'form-element', min: 0, max: 10, step: 1 %>
+    <%= f.select :min_view_trust_level, options_for_select([['everyone', 0], ['anyone with an user account', 1], ['all but new users', 2], ['veteran users', 3], ['moderators only', 4], ['staff only', 5]], selected: @category.min_view_trust_level),
```
cellio, line 53: "a user account" (not "an")

db/seeds/abilities.yml
```diff
@@ -1,52 +1,52 @@
...
   edit_score_threshold: 0.95   # min. +18/-0
   flag_score_threshold: null
 
 - name: Edit Tags
   internal_id: edit_tags
-  icon: fas fa-tags
-  description: >
+  icon: tags
+  summary: >
     Edit tag metadata (excerpt and wiki) and review suggested edits to tags
```
cellio, line 23: This reminds me: I don't know if we currently have UI for suggesting edits to tag wikis and seeing that there are suggestions pending.  If we don't, could you add an issue for that so we don't lose track of it?  Thx.

db/seeds/abilities.yml
```diff
@@ -1,52 +1,52 @@
...
   edit_score_threshold: 0.975   # min. +38/-0
   flag_score_threshold: null
 
 - name: Vote on Holds
   internal_id: flag_close
-  icon: fas fa-lock
-  description: >
+  icon: lock
+  summary: >
     Vote to put questions on hold, vote to reopen on holded questions and review hold votes
```
cellio, line 32: Vote to put questions on hold, see pending hold votes, and vote to keep open or reopen.

(I know we don't have our new hold UI yet with the running tallies, but let's plan for that.  Also, we'd need to change "on holded" anyway.)

luap42: You don't like "on holded"? :P

app/assets/javascripts/privileges.js
```diff
@@ -25,9 +26,12 @@ $(() => {
     const $td = $tgt.parent();
     const $input = $td.find('.js-privilege-edit');
     const name = $input.data('name');
+    const type = $input.data('type');
     const value = $input.val();
 
-    const resp = await fetch(`/admin/privileges/${name}`, {
+    if(!value && value !== 0 && value !== "0") { value = null }
```
ArtOfCode-, line 32: This line doesn't do what you think it does. `0` is already falsy so would already be caught by `!value`. You're also assigning to `value`, which is going to cause a runtime error because `value` is a `const` - you want `let value` on the original definition if you want to reassign to it.

Aside from that, not quite sure what the purpose of this line is - what are you trying to do here? I suspect there's almost certainly an easier way to express it.

app/views/users/show.html.erb
```diff
@@ -1,161 +1,240 @@
...
     </div>
   </div>
-</div>
+  <div class="grid--cell is-3-lg is-12">
+    <div class="widget is-tertiary">
+      <div class="widget--body user-profile--image">
+        <img alt="user avatar" src="<%= avatar_url(@user, 512) %>" />
+      </div>
+      <% if @user.staff? %>
+      <div class="widget--body h-ta-center" title="This user is an official Codidact Staff member.">
```
ArtOfCode-, line 111: I'd leave this (and the next line) as simply "staff member", rather than "Codidact staff member", to make it easier for anyone who wants to use the software for themselves.

app/views/users/show.html.erb
```diff
@@ -1,161 +1,240 @@
...
+          See all &raquo;
+        <% end %>
       <% end %>
     </div>
   </div>
-</div>
+  <div class="grid--cell is-3-lg is-12">
+    <div class="widget is-tertiary">
+      <div class="widget--body user-profile--image">
+        <img alt="user avatar" src="<%= avatar_url(@user, 512) %>" />
```
ArtOfCode-, line 108: Is the avatar actually going to be displayed as 512px? If not, we should use a smaller size here to cut down on download.

app/assets/javascripts/users.js
```diff
@@ -20,4 +20,62 @@ $(() => {
       location.reload();
     }
   });
+
+  $('.js-ability-grant-btn').on('click', async ev => {
+    const $tgt = $(ev.target);
+    const resp = await fetch(`/users/${$tgt.attr('data-user')}/mod/privileges`, {
```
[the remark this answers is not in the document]
luap42: Yay for copying. ✂️

app/controllers/abilities_controller.rb
```diff
@@ -1,17 +1,30 @@
...
+    redirect_to abilities_url(for: @user.id)
+  end
+
+  private
+
+  def set_user
+    return not_found if current_user.nil?
+
+    @user = current_user
+    if params[:for]
```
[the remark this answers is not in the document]
luap42: Deliberatly open for all. Nothing sensitive will be shown and it might be helpful in some cases (ex. support on Meta). For example, SE allows similar stuff, too.

app/models/flag.rb
```diff
@@ -3,9 +3,11 @@ class Flag < ApplicationRecord
...
   belongs_to :user
   belongs_to :handled_by, class_name: 'User', required: false
-
-  validates :reason, length: { minimum: 10, maximum: 1000 }
+  belongs_to :post_flag_type
 
   scope :handled, -> { where.not(status: nil) }
   scope :unhandled, -> { where(status: nil) }
+
+  scope :confidential, -> { where(post_flag_type: PostFlagType.confidential).or(where(post_flag_type: nil)) }
```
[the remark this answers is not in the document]
luap42: Hm.... This doesn't seem true. I was just now able to create a nil flag (other flag)

app/views/categories/_form.html.erb
luap42: TIL that there is used 'a' before 'user' and not 'an' (as I thought because of the 'u'; it appears to be, because it is spoken 'yu')

db/seeds/abilities.yml
luap42: Added to Trello

app/views/users/show.html.erb
luap42: Yes.

app/assets/javascripts/privileges.js
luap42: Fixed.

app/assets/javascripts/privileges.js
```diff
@@ -25,15 +26,22 @@ $(() => {
...
     const resp = await fetch(`/admin/privileges/${name}`, {
       method: 'POST',
       headers: { 'Content-Type': 'application/json', 'X-CSRF-Token': QPixel.csrfToken() },
-      body: JSON.stringify({threshold: value})
+      body: JSON.stringify({type, threshold: value})
     });
     const data = await resp.json();
 
-    $td.removeClass('editing').html('').text(data.privilege.threshold.toString());
+    $td.removeClass('editing').html('').text((data.privilege[type+'_score_threshold'] || '-').toString());
```
ArtOfCode-, line 45: Likewise, prefer interpolation.

app/controllers/admin_controller.rb
```diff
@@ -16,25 +16,30 @@ def error_reports
...
+      @ability.update(post_score_threshold: params[:threshold])
+    elsif params[:type] == 'edit'
+      pre = @ability.edit_score_threshold
+      @ability.update(edit_score_threshold: params[:threshold])
+    elsif params[:type] == 'flag'
+      pre = @ability.flag_score_threshold
+      @ability.update(flag_score_threshold: params[:threshold])
+    end && AuditLog.admin_audit(event_type: 'ability_threshold_update', related: @ability, user: current_user,
+                           comment: "#{params[:type]} score\nfrom <<#{pre}>>\nto <<#{params[:threshold]}>>")
+    render json: { status: 'OK', privilege: @ability }, status: 202
```
ArtOfCode-, line 42: Has Rubocop not complained about this yet? You've got some repeated code in the conditionals that could factor out, and some misalignment in the audit log. I'd suggest simplifying this method to something like:

```suggestion
    @ability = Ability.find_by internal_id: params[:name]
    type = ['post', 'edit', 'flag'].include?(params[:type]) ? params[:type] : nil
    return not_found if type.nil?
    
    pre = @ability.send("#{type}_score_threshold".to_sym)
    @ability.update("#{type}_score_threshold" => params[:threshold])
    AuditLog.admin_audit(event_type: 'ability_threshold_update', related: @ability, user: current_user,
                         comment: "#{params[:type]} score\nfrom <<#{pre}>>\nto <<#{params[:threshold]}>>")
    render json: { status: 'OK', privilege: @ability }, status: 202
```

app/controllers/comments_controller.rb
```diff
@@ -3,14 +3,27 @@ class CommentsController < ApplicationController
...
       render json: { status: 'failed', message: 'Comments have been disabled on this post.' }, status: 403
       return
     end
+
     @comment = Comment.new comment_params.merge(user: current_user)
+
+    unless @post.user_id == current_user.id || @post&.parent&.user_id == current_user.id || \
+           current_user.privilege?('unrestricted')
+      AuditLog.rate_limit_log(event_type: 'comment', related: @post, user: current_user,
+                            comment: "limit: only comment on own posts\n\ncomment:\n#{@comment.attributes_print}")
```
ArtOfCode-, line 20: Misalignment: line `comment` up with `event_type`

app/controllers/comments_controller.rb
```diff
@@ -3,14 +3,27 @@ class CommentsController < ApplicationController
...
+
+      new_user_comment_block_msg = 'New users can only comment on their own posts and on answers to those.'
```
ArtOfCode-, line 22: "New users can only comment on their own posts and on answers to them."

app/views/categories/_form.html.erb
```diff
@@ -40,7 +40,8 @@
       Users must have at least this trust level to post in this category. Leave blank to allow
       anyone to post.
     </span>
-    <%= f.number_field :min_trust_level, class: 'form-element', min: 0, max: 10, step: 1 %>
+    <%= f.select :min_trust_level, options_for_select([['everyone', 0], ['anyone with a user account', 1], ['all but new users', 2], ['veteran users', 3], ['moderators only', 4], ['staff only', 5]], selected: @category.min_trust_level),
```
ArtOfCode-, line 43: Probably worth extracting these options out to a constant on the Ability model, so that we only ever have to change them in one place (L53 likewise).

ArtOfCode- (changes requested):

app/assets/javascripts/privileges.js
```diff
@@ -25,15 +26,18 @@ $(() => {
     const $td = $tgt.parent();
     const $input = $td.find('.js-privilege-edit');
     const name = $input.data('name');
-    const value = $input.val();
+    const type = $input.data('type');
+    let value = $input.val();
+
+    const value = parseFloat($input.val() || '') || null;
```
ArtOfCode-, line 32: Redeclaration of `value`, which prevents compilation. Removing line 30 should solve this.

ArtOfCode- (changes requested):

luap42:
Remind future me to add:

1. An "everyone" ability, which lists all these things everyone has (before unrestricted)

app/models/community_user.rb
```diff
@@ -3,11 +3,14 @@ class CommunityUser < ApplicationRecord
   belongs_to :user
 
   has_many :mod_warnings, dependent: :nullify
+  has_many :user_ability, dependent: :destroy
```
ArtOfCode-, line 6: This should be :user_abilities - Rails usually handles the plurals for you

db/seeds/abilities.yml
```diff
@@ -0,0 +1,69 @@
...
+  flag_score_threshold: 0   # Don't change this, or I'll bite!
+
+- name: Participate Generally
+  internal_id: unrestricted
+  icon: heartbeat
+  summary: >
+    New user restrictions are removed: No limit on posts, vote and comment anywhere
+  description: $FILE abilities/unrestricted.html
+  post_score_threshold: 0.75  # min. +4/-0
+  edit_score_threshold: null
```
ArtOfCode-, line 18: `null` in YAML ends up being a literal string, `"null"` - you want `~`

ArtOfCode-: likewise all the other instances of this

db/seeds/abilities/edit_posts.html
```diff
@@ -0,0 +1,9 @@
+<h2>What does this ability allow me to do?</h2>
+<p>This ability allows you to unilaterally edit posts, as well as review suggested edits from other users.</p>
+
+<p>Edits should be used to improve posts, including edits improving spelling, grammar, and formatting, as well as fixes for missing <a href="/help/alt-text">alt text</a>, broken images and links, and other improvements. Edits should, however, respect the goal of the original poster.</p>
+
+<img src="https://meta.codidact.com/uploads/PW9TNLAZwQWfaX1RMCBEbHDt" alt="Edit post button">
```
ArtOfCode-, line 6: Any images in these should be included in this PR and linked as other included assets (usually an `/assets/xxx` path)

ArtOfCode- (changes requested):

app/views/posts/_post_tools.html.erb
```diff
@@ -53,11 +53,11 @@
           <%= form_tag post_lock_path(post), remote: true, class: 'js-lock' do %>
             <%= label_tag :length, 'Locked until', class: 'form-element', for: "purge_comments_id_#{post.id}" %>
             <% if current_user.is_moderator %>
-              <div class="form-caption">You can set an end date, at which the post will automatically unlock.</div>
+              <div class="form-caption">You can set an amount of days, after which the post will automatically unlock.</div>
```
ArtOfCode-, line 56: "number of days" reads better

app/assets/javascripts/application.js
```diff
@@ -55,10 +55,14 @@ $(document).on('ready', function() {
...
     })
-    .fail((jqXHR, textStatus, errorThrown) => {
-      QPixel.createNotification('danger', '<strong>Failed:</strong> ' + jqXHR.status);
-      console.log(jqXHR.responseText);
-      self.parents(".js-flag-box").removeClass("is-active");
+      .fail((jqXHR, textStatus, errorThrown) => {
+        let message = jqXHR.status;
+        try {
+          message = JSON.parse(jqXHR.responseText)['message'];
+        } finally {
```
ArtOfCode-, line 62: Bracing style :)

```suggestion
        }
        finally {
```

app/controllers/answers_controller.rb
```diff
@@ -13,10 +13,36 @@ def new
...
+
     @answer = Answer.new(answer_params.merge(parent: @question, user: current_user, score: 0,
-                                             body: helpers.render_markdown(params[:answer][:body_markdown]),
-                                             last_activity: DateTime.now, last_activity_by: current_user,
-                                             category: @question.category))
+      body: helpers.render_markdown(params[:answer][:body_markdown]),
+      last_activity: DateTime.now, last_activity_by: current_user,
+      category: @question.category))
+
+    recent_second_level_posts = Post.where(created_at: 24.hours.ago..Time.now, user: current_user) \
```
ArtOfCode-, line 22: No need for the `\` at the end of this line

app/controllers/answers_controller.rb
```diff
@@ -13,10 +13,36 @@ def new
...
   def create
     @question = Question.find params[:id]
+
     @answer = Answer.new(answer_params.merge(parent: @question, user: current_user, score: 0,
-                                             body: helpers.render_markdown(params[:answer][:body_markdown]),
-                                             last_activity: DateTime.now, last_activity_by: current_user,
-                                             category: @question.category))
+      body: helpers.render_markdown(params[:answer][:body_markdown]),
+      last_activity: DateTime.now, last_activity_by: current_user,
+      category: @question.category))
```
ArtOfCode-, line 20: These should line up with the line above:

```suggestion
                                             body: helpers.render_markdown(params[:answer][:body_markdown]),
                                             last_activity: DateTime.now, last_activity_by: current_user,
                                             category: @question.category))
```

app/controllers/flags_controller.rb
```diff
@@ -9,6 +9,21 @@ def new
              PostFlagType.find params[:flag_type]
            end
 
+    recent_flags = Flag.where(created_at: 24.hours.ago..Time.now, user: current_user).count
+    max_flags_per_day = SiteSetting[current_user.privilege?('unrestricted') ? 'RL_Flags' : 'RL_NewUserFlags']
+
+    if recent_flags >= max_flags_per_day
+      flag_limit_msg = 'Thank you. Flags from people like you help us keep this site clean.' \
+                  ' However, you have reached your daily flag limit of ' + max_flags_per_day.to_s + \
```
ArtOfCode-, line 17: Should line up with above

ArtOfCode- (changes requested):
A few comments - there's also a few more places where things should line up, but Rubocop should correct those if you run `rubocop -Da`

app/assets/javascripts/application.js
luap42: wait.... this rule is in our styleguide?

app/views/users/show.html.erb
```diff
@@ -1,163 +1,239 @@
...
+      Account
+    <% end %>
+    <%= link_to edit_user_profile_path, class: "tabs--item" do %>
+      Edit
+    <% end %>
+    <%= link_to subscriptions_path, class: "tabs--item" do %>
+      Subscriptions
+    <% end %>
+  <% end %>
+</div>
```
ArtOfCode-, line 35: **question**: why is this expanded from the `render 'tabs'`? If there are changes needed, we should put them in the tabs view; we get out of sync quickly otherwise.

ArtOfCode- (changes requested):

cellio:
@luap42 you requested a review from me, which I'm happy to do but I'm not sure what to focus on.  I'm not going to be useful for general code review; is there UI text you want me to look at or a dev instance you want me to poke?

public/assets/ability-help/edit-link.png:Zone.Identifier
```diff
@@ -0,0 +1,4 @@
+[ZoneTransfer]
```
ArtOfCode-, line 1: Um. Whoops?

ArtOfCode- (approved):

Issue #138, opened 2020-08-23
README.md outdated


galegosimpatico:
**Describe the bug**
Can't bootstrap a local install _completely_ as per http://github.com/codidact/qpixel/blame/b9e0b1a/README.md#L87, will fail in https://github.com/codidact/qpixel/blob/b9e0b1aa1e720ab26583de3a9dc3c3e5acbd4d51/app/models/user.rb#L195 on impossibility of fetching `SiteSetting['SiteName']`.

**To Reproduce**
1. Follow README.md in order to create a brand new local install (don't forget that you need an empty database to reproduce!).
2. See error

**Expected behavior**
~~A user was created~~ something else preventing user creation.

**Screenshots**
Would add if requested.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: don't care for server error.
 - Version: don't care for server error.

cheers

galegosimpatico:
I can do this, but not until picturing the model or being hinted in which way to fix it (enlarge README.md?, add a populating script?, add fail-safe around `app/models/user.rb:195`?).

galegosimpatico:
database was not properly initialized https://discord.com/channels/634104110131445811/676843271762214932/747325688318656522

I claim the database script can fail silently, would report if find a repro guide.

PR #140, opened 2020-08-24
Proposal for self checkpointing.

galegosimpatico:


galegosimpatico:
related to https://github.com/codidact/qpixel/issues/138#issuecomment-678939060

README.md
```diff
@@ -72,7 +72,13 @@ Set up the database:
...
     rails r db/scripts/create_tags_path_view.rb
     rails db:migrate
-    rails db:seed
+
+&nbsp;
+
+    $ rails db:seed
+    Category: Created 2, skipped 0
+    [...]
+    $ _
```
ArtOfCode-, line 81: What's this line for?

galegosimpatico: I represent a blinking prompt with the underscore character (a caret would be better...), but it doesn't make sense if the last lines aren't shown. I should remove it.

PR #142, opened 2020-08-24
review perms

galegosimpatico:


luap42:
Huh? What does this change?

galegosimpatico:
It is just a reminder that several docs rest with the executable flag set, where that doesn't mean anything; while at the same time offering a click opportunity to unset the flag on the docs, if you choose to do so, or if fits your development process somehow.

Issue #141, opened 2020-08-24
tags page will fail on no tags


galegosimpatico:
**Describe the bug**
Tags page will fail on no tags.

**To Reproduce**
1. Init an empty and populated database with `rails db:seed`.
2. Check the database is populated (e.g. 2 tag sets). `SELECT * FROM tag_sets;`.
3. Check the database has no tags. `SELECT count(*) FROM tags;`.
3. Create user following `README.md`.
4. Go to tags page.
5. See error. Blame no field `tags` on lvalue `@tag_set` at: https://github.com/codidact/qpixel/blob/b9e0b1aa1e720ab26583de3a9dc3c3e5acbd4d51/app/controllers/tags_controller.rb#L30

**Expected behavior**
An empty tags page was shown.

**Screenshots**
Will add on request.

**Desktop (please complete the following information):**
 - OS: N/A.
 - Browser: N/A.
 - Version: N/A.

galegosimpatico:
Hopefully I can fix myself from reading how there can be, however, posts page when there are no posts.

galegosimpatico:
I was wrong

PR #143, opened 2020-08-24
Do not break on empty tags set.

galegosimpatico:
this attempts a fix at https://github.com/codidact/qpixel/issues/141

works for me

cheers


ArtOfCode-:
I suspect this is a misleading error. When categories are seeded, they're seeded without tag sets attached to them, because the tag sets don't yet exist at that point. They need to be edited manually to attach a tag set - once that's done, I suspect you'll find this error goes away.

galegosimpatico:
It did. Thank you!

PR #145, opened 2020-08-24
Add a debugger to the bundle.

galegosimpatico:
I know this is probably very opinionated, inconvenient, even obscene. If so, please forgive me.

But just in order to ask you, how do you debug this software?


luap42:
The usual stuff: A lot of `puts` statements including some of `*************************************` to clearly see it in the log. :D

ArtOfCode-:
There's already a debugger in the bundle: `byebug` - it's only installed on development (and test?) environments, so you won't see it if you're running in production.

To use it, call `byebug` anywhere, and when the app hits that point it'll stop the request in progress and launch the debugger in the terminal.

galegosimpatico:
Pry looks like an awesome lisp-like repl environment for schemers but yeah byebug seems the simpler just right tool for web development.

PR #147, opened 2020-08-25
Stamp date on subsequent error modals.

galegosimpatico:
Before this change modals stack on the same exact place, so the user
does not receive visual feedback until all of them are closed, in case
the errored action is requested over and over again.

This change adds a stamp to subsequent error modals, so the user would
receive visual feedback every time one of the error modals is closed.

**At the time of writing this code hasn't been run yet.**

### DIFF
--- a/app/assets/javascripts/qpixel_api.js
+++ b/app/assets/javascripts/qpixel_api.js
@@ -13,7 +13,7 @@ window.QPixel = {
   },
 
   /** Counts notifications popped up at any time. */
-  var popped_modals_ct = 0;
+  let popped_modals_ct = 0;
 
   /**
    * Create a notification popup - not an inbox notification.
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/qpixel_api.js
```diff
@@ -12,15 +12,42 @@ window.QPixel = {
     return token;
   },
 
+  /** Counts notifications popped up at any time. */
+  var popped_modals_ct = 0;
```
ArtOfCode-, line 16: We're using ES6, so `const` and `let` - `const` if the value won't change, `let` if you need to re-assign to the variable later.

app/assets/javascripts/qpixel_api.js
```diff
@@ -12,15 +12,42 @@ window.QPixel = {
...
+
   /**
    * Create a notification popup - not an inbox notification.
    * @param type the type to apply to the popup - warning, danger, etc.
    * @param message the message to show
    */
   createNotification: function(type, message) {
+    /** Some messages include a date stamp. */
+    var append_date = false;
```
ArtOfCode-, line 25: Likewise `const` and `let` - just assume I've repeated this for every instance of `var` :)

app/assets/javascripts/qpixel_api.js
```diff
@@ -12,15 +12,42 @@ window.QPixel = {
...
+  /** Counts notifications popped up at any time. */
+  var popped_modals_ct = 0;
+
   /**
    * Create a notification popup - not an inbox notification.
    * @param type the type to apply to the popup - warning, danger, etc.
    * @param message the message to show
    */
   createNotification: function(type, message) {
+    /** Some messages include a date stamp. */
```
ArtOfCode-, line 24: Single-line comments are `//`

app/assets/javascripts/qpixel_api.js
```diff
@@ -12,15 +12,42 @@ window.QPixel = {
...
+    var message_with_date = message;
+    var span, button;
+    if (type === 'danger') {
+      if (popped_modals_ct > 0 ) {
+        /* Modals stack each on top, so repeating an errored action over
+         * and over again is gonna create multiple error modals in the
+         * same exact place. While this happens this way, an user
+         * closing the error modal will not have an immediate visual
+         * action feedback if two or more error modals are printed.
+         * A date is stamped in order to cope with that. */
```
ArtOfCode-, line 35: Max characters per line is 120, including comments, so these lines can be longer and the comment won't span so many lines.

ArtOfCode- (changes requested):

galegosimpatico:
remains unrun from my side for now, but if everything goes well with my local setup, now I should be able to check

galegosimpatico:
I've seen the tags cache (what I think it is) being populated before, but for some reason not today, not even when creating a brand new category and tags set, clearing cache, restarting server... 🤔

galegosimpatico:
Now I have seen it working.

app/assets/javascripts/qpixel_api.js
```diff
@@ -12,15 +12,40 @@ window.QPixel = {
...
+    // Some messages include a date stamp, `append_date` governs that.
+    let append_date = false;
+    let message_with_date = message;
+    if (type === 'danger') {
+      if (popped_modals_ct > 0 ) {
+        /* At the time of writing, modals stack each one exactly on top of previous one, so repeating an errored action
+         * over and over again is going to create multiple error modals in the same exact place. While this happens this
+         * way, an user closing the error modal will not have an immediate visual action feedback if two or more error
+         * modals have been printed. A date is stamped in order to cope with that. Could be anything. Probablye a cycle
+         * of different emoji characters would be cuter while having the purpose met. But if so make sure character in
```
galegosimpatico, line 33: obviously insperd by discord's this thing

Screenshot 2020-08-26 at 02 12 24

Screenshot 2020-08-26 at 02 12 30

app/assets/javascripts/qpixel_api.js
```diff
@@ -12,9 +15,6 @@ window.QPixel = {
     return token;
   },
 
-  /** Counts notifications popped up at any time. */
-  let popped_modals_ct = 0;
```
galegosimpatico: https://thecodinglove.com/when-i-run-my-code-for-the-first-time

app/assets/javascripts/qpixel_api.js
galegosimpatico: `var` in https://github.com/codidact/landing-page/pull/53/files too?

ArtOfCode-: I have yet to review that one properly, but probably :)

PR #148, opened 2020-08-25
Add docker-compose for development containers

### LINKED ISSUE #70: Docker (Compose) deployment options?
Not much detail needed, I believe `:)`

Wrapping this software up in a Docker container will surely make development, testing and deployment easier. At first glance, three containers are required: a Redis, a MySQL, and the last one for this app itself.

The Dockerfile would be pretty straightforward: Install necessary software from APT repository and RubyGems, then `ENTRYPOINT /usr/src/app/bin/rails`. A custom script to initialize the application by reading environment variables and generating the configuration files at the first start is also in need.

With all those stuff ready, a `docker-compose.yml` should be easy to work out.

The [`Dockerfile` file](https://github.com/Charcoal-SE/metasmoke/blob/master/Dockerfile) from Charcoal-SE/metasmoke would be a good point to start at, although it's missing some best-practices for Docker.

Ideas?

vsoch:
This pull request will add a "docker" folder with contents to allow for a dockerized development deployment of QPixel. Notably, this initial version is done for development, and we have two ways to proceed:

 1. discuss how the user can set an environment variable for the docker-compose that would signal production, and also configure email notifications to work, make the changes in this PR so the containers can be deployed for development or production.
 2. instead, have this pull request be to provide a development version of the application, and address the first point when someone explicitly asks for it.

One thing that I was thinking - it would be really useful if there was code to allow for any (or some subset) of config options to be grabbed in the environment. That way, the user could customize anything from the environment in Docker instead of having the check for `.dockerenv`. Let me know your thoughts!

This will close #70


vsoch:
I added an environment variable to check if the PR is a fork, because it will fail in that case (and I don't think you'd want to try doing some kind of deploying from some rando person submitting a PR?)

db/schema.rb
```diff
@@ -506,11 +506,6 @@
     t.index ["community_user_id"], name: "index_warnings_on_community_user_id"
   end
 
-  add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
-  add_foreign_key "audit_logs", "communities"
-  add_foreign_key "audit_logs", "users"
-  add_foreign_key "categories", "licenses"
-  add_foreign_key "categories", "tag_sets"
```
ArtOfCode-, line 513: Why the removal of these lines?

docker/create_admin_and_community.rb
```diff
@@ -0,0 +1,12 @@
+# 1. Create the community
+community_name = ENV['COMMUNITY_NAME'] || 'Dinosaur Community'
+Community.create(name: community_name, host: '0.0.0.0:3000')
+
+# 2. Create the admin user, ensure doesn't require confirmation
+username = ENV['COMMUNITY_ADMIN_USERNAME'] || 'admin'
+password = ENV['COMMUNITY_ADMIN_PASSWORD'] || 'password'
+email = ENV['COMMUNITY_ADMIN_EMAIL'] || 'codadict@noreply.com'
+
+User.create(username: username, password: password, email: email, is_global_admin: true, is_global_moderator: true, staff: true)
```
ArtOfCode-, line 10: Is there a reason that we can't just set `confirmed_at` here too?

db/schema.rb
vsoch: Huh, that's strange, I didn't intentionally remove them, I don't even remember opening this file. I can add them back in.

docker/create_admin_and_community.rb
vsoch: I tried this approach - it didn't work. I'm not sure why, to be honest. But re-thinking it, in the same way that it's required by design, we should also require the user to intentionally issue the command to match that. I wrote the instruction clearly in the docs.

db/schema.rb
vsoch: okay added back! I must have opened the file and hit backspace over a selected section without realizing it. /bonk.

vsoch:
Awesome! Please ping me for any issues or additional needs that might come up. Thanks for all the help! And a question for your team - I wanted to see if I could feature qpixel for the weekly software showcase that I run for the research software encyclopedia? E.g see https://rseng.github.io/rseng/blog/. It’s of particular interest to the research community because I know of several groups trying to create a community like this.

luap42:
That sounds great. Thanks for your work on this PR.

vsoch:
Awesome! Are there any docs about background / history of the project, pictures, or anything in particular you'd like me to share? If not, I'll come up with my best effort for an introduction, rationale, and then sharing various links/ideas (akin to the others).

luap42:
@vsoch are you on our dev chat server? If yes, ping me. If not, go to [here](https://discord.gg/7jf8UzS), join and then ping me. :D

cellio:
Sharing with the research science encyclopedia sounds great, thanks!  If you haven't seen our high-level vision yet, please take a look at https://meta.codidact.com/articles/276296.  It's about where we're headed, not so much where we came from; I don't know if the latter is stated clearly anywhere.  

(Basically: unhappy users and moderators on Stack Exchange set out to build an alternative, then realized that communities are more than Q&A and we needed to think bigger.  The focus is not on duplicating or competing with SE; SE was just the catalyst.  By building a new thing we have the opportunity to apply lessons learned over the years on other platforms.)

Or, I now see the comment from luap42 -- do that. :-)

Issue #87, opened 2020-05-25
Indicate that, and when, a post was edited


cellio:
Feature request from Meta: https://meta.codidact.com/questions/74966

If a post has been edited, showing the timestamp of the most recent edit will help people figure out what changed when a post gets bumped.  Right now you have to check the history for the question and each answer until you find it.

galegosimpatico:
related: schema inconsistency 
* never edited post woud have creation date and update date in UTC while last activity date in local time .
* edited post would have creation date and update and activity all in UTC

cellio:
PR was merged; closing.

PR #151, opened 2020-08-25
attempt for issue #87

galegosimpatico:
early draft for https://github.com/codidact/qpixel/issues/87

"works for me"

checklist:
* [x] place / position (easy? / not my thing).
* [x] style (should be easy? / not my thing).
* [x] date fuzzyfication.
* [x] find fuzzied dates compressor-minifier.
  [ ] ~~write fuzzied dates compressor-minifier with tests (if not existed before.~~ methodologies exist.
* [x] fuzzied dates compressor-minifier wired in.
* [x] last editor name (view hack).
  [ ] ~~last editor name (passed from controller).~~ there seem to be many more db queries in the view before and after here, so I guess we won't be pursuing this _right now_.
* [x] last editor name is link to last editor (hack).
  [ ] ~~last editor name is link to last editor (tidy)~~ there seem to be many more db queries in the view before and after here, so I guess we won't be pursuing this _right now_.
  [ ] ~~times edited (hard for me atm / not familiar with schema / not familiar with framework)~~ - not requested.
* [x] tidy up revs, rebase on top of tip, etc


galegosimpatico:
status waiting for label place/position

edited:
Screenshot 2020-08-26 at 01 15 43

nonedited:
Screenshot 2020-08-26 at 01 15 29

galegosimpatico:
example of edited post: Screenshot 2020-08-26 at 15 19 18

galegosimpatico:
I guess it requires review from product development and from front end.

cellio:
On placement, I think your "last edited about X hours ago" line below the post looks good.  If the edit was by somebody other than the author, I'd add "by $username" (with that being a link to the user profile).  There's no need to show the "usercards" (avatar and stats); I think that would be too much emphasis.

I can't tell if it's just your test data or a proposal, but let's not add "edited" to titles.  Over time many posts will be edited; this isn't information that will affect people's actions, the way something being closed is.

I'm ambivalent on "Never edited" versus just leaving the line out until there's an edit.  Either's fine with me; I'll defer to stronger opinions.

luap42:
I have no idea if that will look good or is otherwise reasonable, but what about moving the whole thing to the line with the license behind it? I think the current version takes too much space, but that's just my opinion. Everything could be shortened a little bit, too.

Something like this:

    [scales-icon] CC-BY-SA [pencil-icon] 2d ago by luap43

luap43 would be a link and would only be shown, when not by the author, as @cellio suggested.

I don't know, whether that's the right order, you might put edits before the license.

cellio:
Something like this:
> 
> ```
> [scales-icon] CC-BY-SA [pencil-icon] 2d ago by luap43
> ```

What happens with long user names on a phone? Sure we'll have that problem anyway, but it'll be worse if it's not on a line of its own.

galegosimpatico:
> [...] If the edit was by somebody other than the author, I'd add "by $username" (with that being a link to the user profile). There's no need to show the "usercards" (avatar and stats); I think that would be too much emphasis.

Ok, will explore.

> I can't tell if it's just your test data or a proposal, but let's not add "edited" to titles. [...]

It's because my test data, but I'll treat them with more care from now on.

> I'm ambivalent on "Never edited" versus just leaving the line out until there's an edit. Either's fine with me; I'll defer to stronger opinions.

Me too. I guess it can go away without loss of information, with a win in preventing capturing user focus unnecessarily. It's added only because it was a cheap addition, and for reusing the div/span (label, and label place), and for an enhanced contrast with the edited post. But it can be removed any moment.

galegosimpatico:
> [...] but what about moving the whole thing to the line with the license behind it?

Awesome. Will explore.

> [...] I think the current version takes too much space, [...]

Absolutely. It had just popped out right there.

> [...] Everything could be shortened a little bit, too.
> 
> Something like this:
> 
> ```
> [scales-icon] CC-BY-SA [pencil-icon] 2d ago by luap43
> ```

👌🏿👍🏿🙇🏿

galegosimpatico:
> Something like this:
> 
> > ```
> > [scales-icon] CC-BY-SA [pencil-icon] 2d ago by luap43
> > ```
> 
> What happens with long user names on a phone? Sure we'll have that problem anyway, but it'll be worse if it's not on a line of its own.

IDK; if I make it, would check that before throwing the line back to review.

galegosimpatico:
There should be chances for costlessly replacing

`[pencil-icon] 2d ago by jdoe213`

for `[pencil-icon] 2d ago by me` when the post is being presented to jdoe213. If that's desirable.

galegosimpatico:
new pics - click links for larger

---

[Screenshot 2020-08-26 at 19 53 46][0]

---

[Screenshot 2020-08-26 at 19 54 11][1]

---

[Screenshot 2020-08-26 at 19 54 30][2]

---

label is black and not some greys pictured there - it is missning any style

label missing minification for the fuzzy date - is that used already in qpixe somewhere elsel? if not: needs finding a library? or prefers custom code (if so requires testing)?

[0]: https://user-images.githubusercontent.com/53513496/91338978-16d07800-e7d6-11ea-9d71-6805f5a7ff0e.png
[1]: https://user-images.githubusercontent.com/53513496/91338984-1932d200-e7d6-11ea-8ed7-7a3cc6387c13.png
[2]: https://user-images.githubusercontent.com/53513496/91338986-19cb6880-e7d6-11ea-9842-505e5c9fb9d8.png

cellio:
Images look good to me.  We allow usernames up to 50 characters currently (and if some of them are unicode things those can be wider), so if you could rename the editing user in your last example to something long to test what happens there, that'd be great.

galegosimpatico:
> Images look good to me. We allow usernames up to 50 characters currently (and if some of them are unicode things those can be wider), so if you could rename the editing user in your last example to something long to test what happens there, that'd be great.

Here you are, good catch:

Screenshot 2020-08-26 at 20 16 44

---

It's going to be too large; even if the pencil, the "time", and the 'ago' are made to stick together with the name and the whole "edit phrase" is put under the "license phrase". However, with fuzzy date minification, maybe it could fit, in an independent "edit phrase" under the "license phrase". Then there would be a choice whether to have the two "phrases" always one under the other, or have them fit a line whenever possible.

I don't really know HTML that much... just to live for another day. And know almost no CSS.

cellio:
My main concern was whether it was going to overlap other stuff or run off the right beyond the bounds.  This looks fine to me.  I'm not an expert either, particularly with CSS.  Perhaps @luap42 has other suggestions but this works for me.

Is there a tooltip on the pencil?

galegosimpatico:
> [...]
> 
> Is there a tooltip on the pencil?

There was none. Now it will pop with "Edited". Code is taken from the License icon next to it. Will appear in the PR next time I push updates.

galegosimpatico:
"about" is gone:
Screenshot 2020-08-27 at 16 09 45

english locale gets hardwired:
https://github.com/galegosimpatico/qpixel/blob/ed4e04fd9a213afe255c190b5fa4320824e51474/app/views/posts/_expanded.html.erb#L135
unhardwiring en locale is a job for another day right?

galegosimpatico:
seen the preexisting abbrev en locale https://github.com/codidact/qpixel/blob/12f53bb30d59559a8bd8996cf13cdd7d92961c13/config/locales/en-ABBREV.yml haven't yet figured out how to reuse that one

app/views/posts/_expanded.html.erb
```diff
@@ -118,15 +121,27 @@
               <% end %>
             </div>
           <% end %>
-          <% if post.license.present? %>
-            <div class="post--licensing has-margin-left-2">
-              <i class="fas fa-balance-scale" title="License"></i>
-              <%= link_to post.license.name, post.license.url %>
+          <% if post.license.present? or been_edited then %>
```
ArtOfCode-, line 124: `or` in Ruby is a control flow operator not a logical operator - you want `||`. Also no `then` after an if statement if it's multiple lines (including HTML blocks)

app/views/posts/_expanded.html.erb
```diff
@@ -118,15 +121,27 @@
...
+            <div class="post--licensing--edited has-margin-left-2">
+              <% if post.license.present? then %>
+                <i class="fas fa-balance-scale" title="License"></i>
+                <%= link_to post.license.name, post.license.url %>
+              <% end %>
+              <% if post.license.present? and been_edited then %>
```
ArtOfCode-, line 130: Likewise `and` vs `&&` and no `then` (and likewise elsewhere as well)

app/views/posts/_expanded.html.erb
```diff
@@ -118,15 +121,27 @@
...
+                &nbsp;
+              <% end %>
+              <% if been_edited then %>
+                <i class="fa fa-pencil-alt" title="Edited"></i>
+                <%= time_ago_in_words(post.last_activity, locale: :en_abbrev) %> ago
+                <% if not last_edited_by_self then %>
```
ArtOfCode-, line 136: Same with `not` - you want `!`

config/locales/time.en_abbrev.yml
```diff
@@ -0,0 +1,8 @@
+en_abbrev:
```
ArtOfCode-, line 1: This file can be merged with the rest of en-abbrev

ArtOfCode- (changes requested):

ArtOfCode-:
Can we also make this the same colour as the posted timestamp above the author card (i.e. slightly grayed out)?

config/locales/time.en_abbrev.yml
galegosimpatico: I proposed wiping out the previous one, in https://github.com/codidact/qpixel/pull/151/commits/f5e7360af5abc02d20cc2cbbc007487eb03af2ea

galegosimpatico:
> Can we also make this the same colour as the posted timestamp above the author card (i.e. slightly grayed out)?

Done in https://github.com/codidact/qpixel/pull/151/commits/734fd4db0bcddef75f9cdb1e25b47120c63bda:

Screenshot 2020-08-28 at 12 53 30

galegosimpatico:
sorry for rebasing before being told to

app/views/posts/_expanded.html.erb
```diff
@@ -100,7 +100,10 @@
 
       <div class="post--body">
         <%= raw(sanitize(post.body, scrubber: scrubber)) %>
-  
+
+        <% been_edited = post.last_activity_by_id != nil %>
+        <% if (been_edited) then last_edited_by_self = post.user_id == post.last_activity_by_id end %>
```
ArtOfCode-, line 105: Not blocking on this, but you don't need the parens around the `if` condition - Ruby doesn't require them. It won't break, because it just makes it a subclause of the condition (i.e. same effect), but it's not necessary.

ArtOfCode- (approved):

Issue #152, opened 2020-08-26
Prevent authors from deleting answered questions


cellio:
Somebody asked a question and received multiple answers.  The answers were all upvoted and the question was downvoted.  The author then deleted the question.

I don't think we should allow this.  Deleting a question can deprive the community of good answers.  In this case the question was closed, so maybe in that case it's no big loss, but it reminded me of the problem that Some Other site has with people trying to quickly delete the evidence of cheating on homework.  That's pretty abusive.

I think we should block deletion if the question has at least one answer with Wilson score 0.5.  If there are extenuating circumstances (including a need to comply with CC before we have disassociation), the person can always raise a flag for moderators to review.

This won't be visible to most people, but here's the question where this happened: https://judaism.codidact.com/questions/277460.

cellio:
What do we need to do to finish this other than making the error message meaningful?  @luap42 have you reviewed?

PR #156, opened 2020-08-27
attempt at issue #152.

galegosimpatico:


app/controllers/questions_controller.rb
```diff
@@ -120,6 +120,11 @@ def destroy
       redirect_to(question_path(@question)) && return
     end
 
+    if @question.answer_count > 0 then
+      flash[:danger] = 'Error 1729'
```
ArtOfCode-, line 124: Error messages should be useful - let's make this say something like "This question cannot be deleted because it has answers."

(There is an exception for errors for trolls, but those are defined elsewhere)

ArtOfCode- (changes requested):

galegosimpatico:
I missed checking the Wilson score. 😱

app/controllers/questions_controller.rb
galegosimpatico: Would claim that us both are to co-blame for the message choice.

ArtOfCode-: Eh, there's no blame necessary - we just need this one to say something useful :)

app/controllers/questions_controller.rb
```diff
@@ -120,6 +120,15 @@ def destroy
...
 
+    if @question.answer_count > 0
+      prevent = false  # Prevent based on being tied to well scored answers.
+      @question.answers.each{ |x| if x.score >= 0.5 then prevent = true; break; end }
+      if prevent
+        flash[:danger] = 'This question cannot be deleted because it has answers.'
+        redirect_to(question_path(@question)) && return
+      end
+    end
+
```
ArtOfCode-, line 131: Simpler way of doing it:

```suggestion
    if @question.answer_count > 0 && @question.answers.any? { |a| a.score >= 0.5 }
      flash[:danger] = 'This question cannot be deleted because it has answers.'
      redirect_to(question_path(@question)) && return
    end
```

ArtOfCode- (changes requested):

Issue #136, opened 2020-08-23
RTL directives in a username overflowed into the UI


cellio:
![Screen Shot 2020-08-22 at 9 44 05 PM](https://user-images.githubusercontent.com/5557942/90969106-a4eceb80-e4c2-11ea-841c-557d5812b243.png)

In this screenshot, the UI elements ("delete", "edit", and posting time) are reversed.  That is from a comment here: https://meta.codidact.com/questions/277420.  The user was playing with RTL directives in usernames:

> I simply added U+202E (right-to-left override) and ihsoM (the ltr override flips my nick, but i didn't expect it to flip everything else)

Since some languages require RTL directives we probably should allow these in usernames, *but* we ought to fence them so they can't leak out and cause shenanigans. :-)

galegosimpatico:
Assignees read https://github.com/codidact/qpixel/pull/157.

cellio:
What still needs to happen here?

galegosimpatico:
[Coding the fix][0], unless it's somehow wired in already (if so, shouldn't be able to reproduce anymore).

Problem is also sighted in the `A quiet enigma. We don't know anything about USER yet.` placeholder of the profile page, where it can turn into `A quiet enigma. We don't know anything about .tey USER/RESU/ETC`.

[0]: https://github.com/codidact/qpixel/pull/157#issuecomment-683179647

ArtOfCode-:
a624079f8cafce185a42bcf3e4f1a1ea8c9e80d8

galegosimpatico:
Nope. https://github.com/codidact/qpixel/commit/a624079f8cafce185a42bcf3e4f1a1ea8c9e80d8 results in `A quiet enigma. We don't know anything about  tey.USER` instead of `A quiet enigma. We don't know anything about .tey RESU`. But you want `A quiet enigma. We don't know anything about USER yet.`. Similarly, comments don't get fixed.

ArtOfCode-:
Not deployed yet

galegosimpatico:
just witnessed, only in Safari Version 13.1.2 (15609.3.5.1.3)

[discovered by MathPhysics in older Safari on iOS](https://discord.com/channels/634104110131445811/634104110131445815/756084425401303050)

consider reopening?

PR #157, opened 2020-08-28
attempt a fix for issue #136

galegosimpatico:
Tests are missing, I know. Would add upon request.

Some lines are too large.

I'll try fixing all of that if the PR is reviewed "reasonable" overall.

I mean IDK if wontfix is reasonable at this point. :shrug:

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -138,4 +138,34 @@ module ApplicationHelper
       false
     end
   end
+
+  def is_rtl_override(char)
+    # rtl_override_hex = '\u202E'
+    char_bytes = char.bytes
+    if char_bytes.length() != 3 then
+      return false
+    end
+    if char_bytes[0] != 226 then  # 0xE2
+      return false
+    end
+    if char_bytes[1] != 128 then  # 0x80
+      return false
+    end
+    if char_bytes[2] != 174 then  # 0xAE
+      return false
+    end
+    return true
+  end
+
+  def sanitize_username(username)
+    return username
+  end
+
+  def directional_formatting_pops_after_username_ct(username)
+    if is_rtl_override(username[0]) then
+      # pop_directional_formatting_hex = '\u202C'
+      return 1
+    end
+    return 0
+  end
 end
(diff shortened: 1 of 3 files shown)


ArtOfCode-:
I suspect there's a much easier way to do this, rather than messing with Unicode (again...): just stick an LTR-override character before and after every username, regardless of whether it's actually got RTL in it or not.

Could be done by implementing a `ltr_safe_username` method on User which simply returns `"\u202D#{username}\u202D"`, then searching for all `username` calls in views and replacing them with `ltr_safe_username`.

galegosimpatico:
> I suspect there's a much easier way to do this, rather than messing with Unicode (again...): just stick an LTR-override character before and after every username, regardless of whether it's actually got RTL in it or not.
> 
> Could be done by implementing a `ltr_safe_username` method on User which simply returns `"\u202D#{username}\u202D"`, then searching for all `username` calls in views and replacing them with `ltr_safe_username`.

I hadn't tried that exactly... before arriving to this contrived idea, I tried to set up LTR override jails like that, just done in HTML because I suck so much at Ruby, but that didn't fix my stuff. Now I just tried your idea and didn't fix it either.

At least now I picture how to create a chain of directional formatting pops within a Ruby sanitizing function without resorting to a template loop on adding the HTML escape sequence of the directional formatting pop over and over again.

😕

ArtOfCode-:
Okay, got it. Add this method to User:

```ruby
def rtl_safe_username
  "\u202D#{username}\u202D"
end
```

Then, for any usage of `u.username` in a view, where `u` is a User instance, replace it with `u.rtl_safe_username`, _and_ add a `dir="ltr"` attribute to the surrounding HTML element (add a `<span>` for containment if necessary).

galegosimpatico:
If you are sure, then I should yield it so you can have parent issue becoming [meta: good first issue][0].

[0]: https://github.com/codidact/qpixel/labels/meta%3A%20good%20first%20issue

ArtOfCode-:
I don't mind at all - if you want it, you're welcome to it; no point letting stuff sit around for ages just because a new contributor _might_ want to do it.

galegosimpatico:
I couldn't care less about what's cooking in no tag's land.

PR #172, opened 2020-09-04
Review suggested edit using Diffy.

galegosimpatico:
Use Diffy ("edits history" view) when reviewing the suggested edit. It just works.

### DIFF
--- a/app/views/suggested_edit/show.html.erb
+++ b/app/views/suggested_edit/show.html.erb
@@ -59,11 +59,16 @@
                 <% if @edit.body_markdown.nil? %>
                     <div class="diff-inline raw-markdown"><%= post.body_markdown %></div>
                 <% else %>
+                    <%if @edit.active %>
+                    <% diff = Diffy::SplitDiff.new(post.body_markdown, @edit.body_markdown, format: :html) %>
+                    <div class="diff-old is-changed raw-markdown"><%= raw(sanitize(diff.left, scrubber: scrubber)) %></div>
+                    <div class="diff-new is-changed raw-markdown"><%= raw(sanitize(diff.right, scrubber: scrubber)) %></div>
+                    <% else %>
                     <div class="diff-old is-changed raw-markdown"><%= post.body_markdown %></div>
                     <div class="diff-new is-changed raw-markdown"><%= @edit.body_markdown %></div>
+                    <% end %>
                 <% end %>
             </div>
-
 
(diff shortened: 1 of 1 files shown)


luap42:
Cool.

However, I see, that this is only for active edits. Likely, because the code can currently only compare towards the *status quo* of the post, but the edit should (once no more active) compare to the version that was current, when it was decided. Otherwise, any accepted edit will just be marked as "no changes" or if anything else was added later, as reverting that.

So, we'd all be very happy, if you save the current version (either the revision or likely more error-tolerant by adding new columns to SuggestedEdit) upon edit submittal and add a comparison to that. ;)

(Or if you don't want to, we can also approve this and do it later of course)

galegosimpatico:
I want to. It's almost done in fact! But I won't ruin this PR with the pile of stuff remaining. I don't buy that inspecting a past suggestion is the same feature as assessing a current active suggestion, because by chance and technical detail it reuses the view.

Best case, expect a new PR around Sunday 13th. I don't feel like stacking more changes here.

Hopefully someone makes it before me.

galegosimpatico:
I think I have a good lead - revs 0fb56f79d5c7d036bddfd322475627bbb557cd12 and d6bf3a4362d807e6f46106ff8143386d77766949. It seems good to me, but I'll torture that next week to be sure (different post types maybe...). After that, I would open PR while advising that I'd open yet a third PR getting it right - like referencing the previous markdown instead of duping it.

I hope I'm in the right direction, else you are welcome to destroy that!

galegosimpatico:
https://github.com/codidact/qpixel/commit/d6bf3a4362d807e6f46106ff8143386d77766949 Missing change for questions and articles - noticed 👌🏿👩🏿‍🌾👍🏿

ArtOfCode-:
No problem if you're planning to do historic suggestions in a different PR; we'll hold off merging this one until they're both ready, then merge & deploy both together so we deploy the whole feature set in one go.

galegosimpatico:
### Master Foo and the changes batch.

A visiting programmer came with a batch of two changes, saying "Here's this tiny change, it's small but is a huge life improvement change for reviewing edits; and also there's that huge, complex, messy change for inspecting past edit reviews. Luckily, they are loosely coupled, so you can push the small one early and have the big one pushed once it's ready. Release early, they say!".

Master Foo spoke: "That's right. Once the simple change is pushed, reviewing suggested edits won't be a pain, so people will stop refraining from it.".

Upon hearing this, the visiting programmer was enlightened.

Issue #175, opened 2020-09-11
Attempting to merge a tag into itself spits out a server fault


donielf:
In attempting to merge a tag into itself, I instead received an error 500 Server Fault, with the accompanying error ID `16e9af53-6e7e-45ea-bf67-ddf171f033c5`. The tag was left unchanged through this process, and I didn't crash the site in testing this.

In contrast, when trying to set a tag as its own parent, I received an error message just saying that "you can't do that."

galegosimpatico:
Hopefully resolved in #176.

PR #176, opened 2020-09-11
Better visualization for tag merge (with itself).

galegosimpatico:
Flash 400 instead of 500 for the invalid request.

Attempts to tackle issue #175.

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -94,6 +94,26 @@ class TagsController < ApplicationController
 
   def merge
     @primary = @tag
+
+    # No merge to self.
+    if params[:merge_with_id].eql? @primary.id.to_s
+      flash[:danger] = 'Cannot merge a tag with itself.'
+
+      # redirect_to tag_path(id: @category.id, tag_id: @primary.id) && return
+
+      # render :edit, status: 400
+      # return
+
+      # render json: { success: false, status: 400, tag: @tag }
+      # return
+
+      # render 'tag', category: @category, tag: @tag
+      # return
+
+      redirect_back fallback_location: categories_path
+      return
+    end
+
     @subordinate = Tag.find params[:merge_with_id]
 
     # Take the tag off posts
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -96,20 +96,8 @@ class TagsController < ApplicationController
     @primary = @tag
 
     # No merge to self.
-    if params[:merge_with_id].eql? @primary.id.to_s
+    if params[:merge_with_id] == @primary.id.to_s
       flash[:danger] = 'Cannot merge a tag with itself.'
-
-      # redirect_to tag_path(id: @category.id, tag_id: @primary.id) && return
-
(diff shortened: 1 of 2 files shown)


app/controllers/tags_controller.rb
```diff
@@ -94,6 +94,26 @@ def select_merge; end
 
   def merge
     @primary = @tag
+
+    # No merge to self.
+    if params[:merge_with_id].eql? @primary.id.to_s
+      flash[:danger] = 'Cannot merge a tag with itself.'
+
+      # redirect_to tag_path(id: @category.id, tag_id: @primary.id) && return
```
galegosimpatico, line 102: This didn't do what I expected, at all.

app/controllers/tags_controller.rb
```diff
@@ -94,6 +94,26 @@ def select_merge; end
...
+
+      # render :edit, status: 400
```
galegosimpatico, line 104: This worked better than the previous one, but didn't do what anyone would expect it the page to do, TBH I was seeing it coming.

app/controllers/tags_controller.rb
```diff
@@ -94,6 +94,26 @@ def select_merge; end
...
+      # return
+
+      # render json: { success: false, status: 400, tag: @tag }
```
galegosimpatico, line 107: This was a very bad idea that seemed not so bad until executed.

app/controllers/tags_controller.rb
```diff
@@ -94,6 +94,26 @@ def select_merge; end
...
+      # return
+
+      # render 'tag', category: @category, tag: @tag
```
galegosimpatico, line 110: This wasn't tested very much, I'm no longer even sure it worked at all.

app/controllers/tags_controller.rb
```diff
@@ -94,6 +94,26 @@ def select_merge; end
...
+      # return
+
+      redirect_back fallback_location: categories_path
```
galegosimpatico, line 113: This didn't what I was expecting it to do either, at all, but I think the results were unexpectedly great.

app/controllers/tags_controller.rb
galegosimpatico: applying technique https://github.com/codidact/qpixel/pull/178/commits/dfae8b9f0a486aca417804fbf14e2d44308841da uses fixes it, but https://github.com/codidact/qpixel/pull/176#discussion_r487302984 is actually better

app/controllers/tags_controller.rb
```diff
@@ -94,6 +94,26 @@ def select_merge; end
 
   def merge
     @primary = @tag
+
+    # No merge to self.
+    if params[:merge_with_id].eql? @primary.id.to_s
```
ArtOfCode-, line 99:
```ruby
if params[:merge_with_id] == @primary.id.to_s
```

ArtOfCode- (changes requested):

app/controllers/tags_controller.rb
galegosimpatico: I tested it and then committed

PR #205, opened 2020-10-11
Update error codes from 404 to 403 in various places

### LINKED ISSUE #201: "Forbidden" ("403") page actually uses HTTP status 401 Unauthorized instead of 403 Forbidden for deleted posts without permission
https://meta.codidact.com/questions/277552

> There's a variety of actions which results in a display page saying "403 FORBIDDEN"; for example, accessing a deleted question without sufficient reputation to view deleted questions, including when not signed in.
>
> However, the actual HTTP response code for that, as used in the HTTP response, is not 403, but 401.

See more explanation there.

sau226:
This commit replaces 401 status codes with 403s in some places to conform with HTTP specs

At the time of writing, no tests were run by the PR author.

Please feel free to mark this PR as draft if changes are required - this way its status is clear to everyone.

GitHub issue linking tag: closes #201


ArtOfCode-:
Thanks!

PR #209, opened 2020-10-16
Add a Birthday Challenge Ranking page

luap42:
Shhhhhh.


app/controllers/birthday_controller.rb
```diff
@@ -0,0 +1,14 @@
+class BirthdayController < ApplicationController
+  def index
+    render layout: 'without_sidebar'
+  end
+
+  def ranking
+    @ranking = Rails.cache.fetch('birthday-ranking', expires_in: 1.hour) do
+      User.unscoped.joins(:posts).where('posts.score > 0.6') \
```
ArtOfCode-, line 8: `>` or `>=`?

public/assets/undraw_happy_birthday_s72n.svg:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
+ZoneId=3
+HostUrl=about:internet
```
ArtOfCode-, line 3: Um?

app/controllers/birthday_controller.rb
luap42: > is correct per an internal discussion. (we don't want exactly 1 upvote to suffice)

public/assets/undraw_happy_birthday_s72n.svg:Zone.Identifier
luap42: good question...

Issue #211, opened 2020-10-22
Comment character count doesn't update for insert from context menu


cellio:
https://meta.codidact.com/questions/277566

Inserting text into a comment from the clipboard works; the character count is updated.  Instead inserting from the browser context menu ("Paste") doesn't work.  See linked post for a video.  I just reproduced in Chrome on a Mac.

PR #219, opened 2020-10-23
tackling comment character count not updating when pasting and cutting from context menu

galegosimpatico:
tries to bind character count to mouse events too

tackles [Comment character count doesn't update for insert from
context menu](https://github.com/codidact/qpixel/issues/211)

works for me

early wip

will reformat

cheers

### DIFF
--- a/app/assets/javascripts/character_count.js
+++ b/app/assets/javascripts/character_count.js
@@ -36,12 +36,14 @@ $(() => {
       $el.text(text);
     }
 
-    $target.bind('paste', function(e) {
-      setTimeout(function() { character_count(e); }, 100);
-    });
-
-    $target.bind('cut', function(e) {
-      setTimeout(function() { character_count(e); }, 100);
+    $target.on('cut paste', function(e) {
+      //   This `millis` constant is not optimal, I just picked a number which is small enough for me to almost not
+      // notice the timed wait (without which this change stops working!), while getting the work done. I guess this
+      // hardwired timed wait might potentially not be slow enough for a browser running in a very CPU-busy client!!!
+      // Here would use a media query on "system load" and choose a `millis` value safe enough so that the code in
+      // `character_count` never gets run ahead of when it's supposed to.
+      const millis = 100;
+      setTimeout(function() { character_count(e); }, millis);
     });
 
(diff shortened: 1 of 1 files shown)


MoshiKoi:
I believe you can just do `$target.on('keyup cut paste cutcc-reset', ` all in one instead of repeatedly using .bind

I also suggest binding 'input' which by specification will catch any and every change, the rest is basically support for browsers that don't fully support 'input' (which tbh shouldn't be many).

https://stackoverflow.com/questions/5494648/catching-all-changes-to-the-contents-of-an-input-box-using-javascript-jquery/5494697#5494697

ArtOfCode-:
:+1: as Moshi says for merging all the triggers into one `on` call. Also worth noting that `bind` is not the preferred method of doing it - better to use `on`.

ArtOfCode- (changes requested):
:+1: as Moshi says for merging all the triggers into one on call. Also worth noting that `bind` is not the preferred method of doing it - better to use `on`.

galegosimpatico:
aye

galegosimpatico:
but well the timing sh is necessary (just remembered), so the only thing we can do here is to merge paste and cut - which will do.

OR

apply the timeout to everything. Which is more condensed but would use the timing sh even when it's not necessary

thx 4 ur input!

ArtOfCode-:
Okay, if the timing's necessary let's do as you suggest and just merge paste and cut - we should also change the `bind` to `on`, then we're good to go.

galegosimpatico:
I have checked (hence [here][0] and [here][1]) how to use `on` instead of `bind` unfortunately doesn't remove the need for the timed wait. :(

[0]: https://github.com/codidact/qpixel/pull/219#event-3916805324
[1]: https://github.com/codidact/qpixel/pull/219#event-3916807110

galegosimpatico:
> [...]
> 
> I also suggest binding 'input' which by specification will catch any and every change, the rest is basically support for browsers that don't fully support 'input' (which tbh shouldn't be many).
> 
> [...]

I'm not investigating this myself, but only because I've read, elsewhere, the comments system will be completely revamped.

PR #220, opened 2020-10-24
tweaking comments to provide comment cancellation

galegosimpatico:
https://meta.codidact.com/questions/278770

https://meta.codidact.com/questions/277948

cheers

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -50,12 +50,11 @@ $(() => {
     // no need to do anything else at this point in this block, until proven otherwise.
   });
 
-  $('.comment-form').find('input[value="Nope"]').on('click', function (data) {
-      const $input_button = $(data.target);
-      const $div_actions = $input_button.parent();
-      const $div_form_group_horizontal = $div_actions.parent();
-      const $form_js_comment_form = $div_form_group_horizontal.parent();
-      $form_js_comment_form.hide();
+  $('.comment-form').find('input[value="Discard"]').on('click', function(data) {
+      const $tgt = $(data.target);
+      const $form = $tgt.parents('form');
+      $form[0].reset();  // Clear the comment field before hiding the form for a fresh start.
+      $form.hide();
   });
 
   $(document).on('click', '.js-comment-edit', async evt => {
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -50,6 +50,14 @@ $(() => {
     // no need to do anything else at this point in this block, until proven otherwise.
   });
 
+  $('.comment-form').find('input[value="Nope"]').on('click', function (data) {
+      const $input_button = $(data.target);
+      const $div_actions = $input_button.parent();
+      const $div_form_group_horizontal = $div_actions.parent();
+      const $form_js_comment_form = $div_form_group_horizontal.parent();
+      $form_js_comment_form.hide();
+  });
+
   $(document).on('click', '.js-comment-edit', async evt => {
     evt.preventDefault();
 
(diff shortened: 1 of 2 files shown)


app/views/posts/_expanded.html.erb
```diff
@@ -349,6 +349,7 @@
                     </div>
                     <div class="actions">
                       <%= f.submit 'Post', class: 'button is-outlined' %>
+                      <input type="button" name="button" value="Nope" class="button is-outlined" />
```
ArtOfCode-, line 352: To make this more consistent with elsewhere:

```html
<input type="button" name="cancel-comment" value="Cancel" class="button is-danger is-outlined" />
```

app/assets/javascripts/comments.js
```diff
@@ -50,6 +50,14 @@ $(() => {
     // no need to do anything else at this point in this block, until proven otherwise.
   });
 
+  $('.comment-form').find('input[value="Nope"]').on('click', function (data) {
+      const $input_button = $(data.target);
+      const $div_actions = $input_button.parent();
+      const $div_form_group_horizontal = $div_actions.parent();
+      const $form_js_comment_form = $div_form_group_horizontal.parent();
+      $form_js_comment_form.hide();
+  });
```
ArtOfCode-, line 59: * Variable names should be camelCase in JavaScript
* Also, changes for updated button HTML and for simplicity:

```suggestion
  $('.comment-form').find('input[name="cancel-comment"]').on('click', function(data) {
      const $tgt = $(data.target);
      const $form = $tgt.parents('form');
      $form[0].reset(); // clear the comment field before hiding the form for a fresh start
      $form.hide();
  });
```

ArtOfCode- (changes requested):

app/views/posts/_expanded.html.erb
galegosimpatico: Hmm.
* "Cancel" would be hard wired English in code. "Nope" is leet universal immune to l10n.
* Doesn't require `is-danger` because it is no danger.
* Agree on a different `name`.

app/assets/javascripts/comments.js
galegosimpatico: Ahmhá, yeah, easier access to the parent form, thank you. But `.reset()` is not needed, it's needlessly destructive, and it doesn't ask confirm before destruction, being the only warning is the passive protection `.is-danger`.

app/views/posts/_expanded.html.erb
ArtOfCode-: The whole of the application is hard-wired English - i18n is something we'll come to eventually, but not something we're worrying about right now. Whether it's "Cancel" or "Discard" or something, "Nope" doesn't make clear exactly what the button does, which is why I'm suggesting changing it.

app/assets/javascripts/comments.js
ArtOfCode-: However, it's what users expect a Cancel button to do - if you click Cancel, you _want_ to discard whatever you had there. Not doing so subverts those expectations.

app/views/posts/_expanded.html.erb
galegosimpatico: "Cancel" is no better than "Nope" besides being historic, maybe an historic error. "Discard" is very good, I will apply it.

app/assets/javascripts/comments.js
galegosimpatico: "Cancel" doesn't allow newcomer to expect (or not to expect) a confirmation modal, nor does convey any definitive information about the content being discarded. I'm keen to apply the reset suggestion with the "Discard" suggestion.

galegosimpatico:
I'm gonna keep forgetting to camelCase JavaScript variables. That won't be on purpose, snake case just feels more natural for JavaScript to me.

I also don't prepend the dollar, I just did it in this case for consistency with the lines around.

ArtOfCode- (approved):

PR #223, opened 2020-10-24
Fix first visit notice

### LINKED ISSUE #217: Can't open hamburger menu on Meta when signed out
https://meta.codidact.com/questions/278713

On Meta, but not other communities, when signed out the hamburger menu does not respond.  I reproduced this problem in Firefox on Android.  I note that Meta is the only community with the new-visitor popup enabled, and I can't dismiss that either.  These might be related but I'm not sure.

sau226:
This PR fixes an issue with the first visit notice, which also resolves an issue with opening the hamburger menu on mobile for sites with such a notice enabled.

This patch was tested on a local dev instance and on first inspection appears to work.

Please mark this as draft or submit a request changes review if changes are required.

GitHub issue linking tag: resolves #217.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -135,8 +135,9 @@ $(document).on('ready', function() {
     });
   });
 
-
-  document.getElementById("fvn-dismiss").onclick = function() {
-    document.cookie = 'dismiss_fvn=true; path=/; expires=Fri, 31 Dec 9999 23:59:59 GMT';
+  if (document.cookie.indexOf ('dismiss_fvn') == -1 ) {
+    document.getElementById("fvn-dismiss").onclick = function() {
+      document.cookie = 'dismiss_fvn=true; path=/; expires=Fri, 31 Dec 9999 23:59:59 GMT';
+    }
   };
 });
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -135,7 +135,7 @@ $(document).on('ready', function() {
     });
   });
 
-  if (document.cookie.indexOf ('dismiss_fvn') == -1 ) {
+  if (document.cookie.indexOf('dismiss_fvn') === -1 ) {
     document.getElementById("fvn-dismiss").onclick = function() {
       document.cookie = 'dismiss_fvn=true; path=/; expires=Fri, 31 Dec 9999 23:59:59 GMT';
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/application.js
```diff
@@ -135,8 +135,9 @@ $(document).on('ready', function() {
     });
   });
 
-
-  $('.js-first-visit-notice').on('close.bs.alert', async () => {
-    document.cookie = 'dismiss_fvn=true; path=/; expires=Fri, 31 Dec 9999 23:59:59 GMT';
-  });
+  if (document.cookie.indexOf ('dismiss_fvn') == -1 ) {
```
ArtOfCode-, line 138: **suggestion** (code-issues, blocking): `===` for equality checks in JS other than `== undefined`

**nitpick** (code-style, non-blocking): no space before opening paren in function call, i.e. `indexOf('dismiss_fvn')`

app/assets/javascripts/application.js
```diff
@@ -135,8 +135,9 @@ $(document).on('ready', function() {
...
+    document.getElementById("fvn-dismiss").onclick = function() {
+      document.cookie = 'dismiss_fvn=true; path=/; expires=Fri, 31 Dec 9999 23:59:59 GMT';
+    }
```
ArtOfCode-, line 141: **suggestion** (es6, libraries, blocking): we have both ES6 and jQuery available, which we should use to maintain consistency here:

```suggestion
    $('#fvn_dismiss').on('click', () => {
      document.cookie = 'dismiss_fvn=true; path=/; expires=Fri, 31 Dec 9999 23:59:59 GMT';
    });
```

ArtOfCode- (changes requested):
See https://conventionalcomments.org/ for comment format.

app/assets/javascripts/application.js
sau226: All changes in this review comment were implemented

app/assets/javascripts/application.js
sau226: All changes in this comment implemented via committing of the suggested change

luap42:
I see @ArtOfCode- already commented on some stuff. I think that right now using classes and not IDs is more generally consistent, but I’ll have to look, which I’ll be able to do tomorrow evening. Besides that, looks good.

Issue #133, opened 2020-08-17
Show recent voting activity on user profile


cellio:
With reputation, you see the number change and want to know what happened -- but nothing tells you.  Even without reputation, though, it's useful to be able to see what posts are getting activity.  On SE, when I got rep notifications for something old, that told me it was being seen again (linked, bounty, whatever), I would take a look to see if it was still good or could be improved.  I always found it frustrating that on per-site metas, where there's no rep, we also had no indication of voting activity, meaning that voting could have changed wildly on some old meta discussion about scope or something and we'd never notice.

Regardless of what ends up in the usercard or user stats, could we have a tab on the user profile that shows voting events, reverse-chronological, for posts anywhere on the site?  I'm imagining a compact list; we don't need to show tags and post types, I don't think.  Category + title on one line, along with the voting change, should be fine.  Grouping all activity on a post on one day together into one entry is nice but not essential.  I don't think we need timestamps of individual votes, just dates.

I think this is primarily useful to the affected user (and maybe mods), though I don't know of a reason to keep it private.

cellio:
I bumped the priority up because this keeps coming up and I think it's a pretty significant gap in engaging users.

cellio:
Here's a mockup of what I have in mind:

![votes-tab](https://user-images.githubusercontent.com/5557942/128651061-aaa97db5-2f86-4070-984a-e2b7b02cba7a.png)

Istiakshovon:
Just a reminder. Maybe, a related issue was created recently.

PR #231, opened 2020-10-26
User profile: votes by date

galegosimpatico:
Playfully approach https://github.com/codidact/qpixel/issues/133.


cellio:
What does it do?  It looks like it's doing something with votes *cast*, though I might be misreading.  The request was for votes *received* -- which of your stuff is getting voted on?

galegosimpatico:
oops

PR #233, opened 2020-10-26
User profile: votes by date

galegosimpatico:
Playfully approach https://github.com/codidact/qpixel/issues/133.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -123,7 +123,7 @@
   votes_by_post_type = @user.votes.joins(:post).group(Arel.sql('posts.post_type_id')).count(Arel.sql('posts.post_type_id'))
 %>
 
-<h2 class="user-profile-heading">Votes</h2>
+<h2 class="user-profile-heading">Casted votes</h2>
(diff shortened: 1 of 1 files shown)


cellio:
Could you add a screen shot, for those of us who can't just mentally compile and execute the code? :-)  Thanks.

galegosimpatico:
Screenshot 2020-10-26 at 15 33 34

galegosimpatico:
um missed some kind of header, which wasn't necessary for casted votes because it was fitting in the same section

galegosimpatico:
Screenshot 2020-10-26 at 15 39 44

galegosimpatico:
* [ ] Check freshly created user.

cellio:
Thanks for adding the header; was typing a comment about that when you fixed it. :-)

What happens when there are hundreds of votes?  Pagination?  (Ultimately we probably want a separate tab on the user profile for this, but if a long list doesn't overwhelm the page, we can do that later.)

galegosimpatico:
> What happens when there are hundreds of votes? Pagination? [...]

ATTOW only the most recent few comments are printed.

* [ ] "Received votes" auto transforming to "Recently received votes" when there are more 'received votes' than `how_many`.

galegosimpatico:
The new layout breaks visualization of this; let's get back here at some point later - won't remove the branch or commits.

PR #242, opened 2020-10-27
Luap42/new site btn

luap42:
This PR preserves a previously already merged PR state (#130), which was for some extremely weird reasons removed.

Likely not supposed to be merged, but maybe that fixes it.


app/controllers/admin_controller.rb
```diff
@@ -61,6 +61,85 @@ def audit_log
...
+    settings.find_by(name: 'ModBadgeCharacter')
+            .update(value: default_settings.find_by(name: 'ModBadgeCharacter').value)
+    settings.find_by(name: 'SEApiClientId')
+            .update(value: default_settings.find_by(name: 'SEApiClientId').value)
+    settings.find_by(name: 'SEApiClientSecret')
+            .update(value: default_settings.find_by(name: 'SEApiClientSecret').value)
+    settings.find_by(name: 'SEApiKey')
+            .update(value: default_settings.find_by(name: 'SEApiKey').value)
+    settings.find_by(name: 'AdministratorContactEmail')
+            .update(value: default_settings.find_by(name: 'AdministratorContactEmail').value)
```
ArtOfCode-, line 118: **issue** (code-clarity): this section is not ideal - think DRY. Can we create a hash instead mapping `primary_color: 'SiteCategoryHeaderDefaultColor'` etc instead and iterate through that to apply each setting - and likewise for the auto-load defaults?

app/controllers/admin_controller.rb
```diff
@@ -61,6 +61,85 @@ def audit_log
...
+      'discussion', 'support', 'feature-request', 'bug',
+      'status-completed', 'status-declined', 'status-review', 'status-planned', 'status-deferred'
+    ]
+    Tag.create(tags.map { |t| { name: t, community_id: @community.id, tag_set: TagSet.meta } })
+
+    # Set Meta tags as required/mod-only
+    meta_category = Category.where(name: 'Meta').last
+    meta_category.required_tags << Tag.where(name: ['discussion', 'support', 'feature-request', 'bug'])
+    meta_category.moderator_tags << Tag.where(name: ['status-completed', 'status-declined', 'status-review',
+                                                     'status-planned', 'status-deferred'])
```
ArtOfCode-, line 131: **issue** (bug): the `Tag` selectors here are missing a tag set scope - at present they're going to add every instance of each of these tags from every community to the new community's required/mod tags.

app/controllers/admin_controller.rb
```diff
@@ -61,6 +61,85 @@ def audit_log
...
+
+    # Clear cache
```
ArtOfCode-, line 133: **suggestion** (non-blocking): unnecessary comment - pretty clear what the following line does already

app/controllers/admin_controller.rb
```diff
@@ -61,6 +61,85 @@ def audit_log
...
+    Rails.cache.clear
+
+    # Audit Log
```
ArtOfCode-, line 136: **suggestion** (non-blocking): unnecessary comment - pretty clear what the following line does already

app/views/admin/setup.html.erb
```diff
@@ -0,0 +1,88 @@
...
+
+<p class="is-lead">Welcome to your new <%= @community.name %> community!</p>
+
+<p>We are almost ready. We just need some more information to complete creation of the site.</p>
+
+<p>Don't worry. You will be able to change all these things in the site settings later.</p>
+
+<%= form_tag setup_save_url do %>
+
+<div class="widget"><div class="widget--body">
```
ArtOfCode-, line 11: **issue** (code-clarity): missing an indent level here after the opening `form_tag`. Also, shouldn't have two opening `<div>`s on the same line - separate lines with a new indent level for the second. Likewise the closing tags, and likewise the rest of this file.

ArtOfCode- (changes requested):

app/views/admin/setup.html.erb
galegosimpatico: [https://thecodinglove.com/content/041/cfkRtbl.webm](https://thecodinglove.com/when-i-find-the-missing-div)

Issue #206, opened 2020-10-12
History for articles is missing the title/score/link block


cellio:
As reported here: https://meta.codidact.com/questions/278476

Post history for a question or answer shows the "block" with the title/link and score: 

![image](https://user-images.githubusercontent.com/5557942/95776647-b8caf780-0c92-11eb-9a06-079b73d5ac29.png)

But articles do not:

![image](https://user-images.githubusercontent.com/5557942/95776664-c1bbc900-0c92-11eb-8020-0e2433c87853.png)

I hope this is a simple matter of including the same element for articles as we do for questions.  (Not shown here, but it's also there for answers, with a smaller font.  Articles and questions are both top-level posts, so copy from questions not answers.)

rileyradoll:
Hello @cellio I'd like to try tackling this issue. Can you assign this to me?

cellio:
@rileyradoll all yours!  Thanks for tackling it.

ArtOfCode-:
Thanks! Let us know if you need any help.

rileyradoll:
Hi @cellio I think I've solved the issue. Can you check out my PR when you have a chance?
https://github.com/codidact/qpixel/pull/245 Thanks!

cellio:
@ArtOfCode- or @luap42 can you review this?  Thanks.

PR #245, opened 2020-10-28
Add missing title/score block for article history

rileyradoll:
Closes: #206

### DIFF
--- a/app/views/posts/_list.html.erb
+++ b/app/views/posts/_list.html.erb
@@ -1,4 +1,5 @@
 <% is_question = post.post_type_id == Question.post_type_id %>
+<% is_article = post.post_type_id == Article.post_type_id %>
 <% is_meta = (is_question && post.meta?) || (!is_question && post.parent&.meta?) %>
 <% active_user = post.last_activity_by || post.user %>
 <% @show_type_tag = !!defined?(show_type_tag) ? show_type_tag : false %>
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
Thanks for your work on this! There's an easier way to solve this one, though: the `posts/_type_agnostic` view is a "list" view that works regardless of post type. Instead of creating a new article-list view, can we instead use the type agnostic view and just remove any "if post type is question/answer" conditionals?

cellio:
A question list can have a mix of questions and articles; it's not necessarily just one type.  (See Judaism's Challenges for an example.)  I didn't expect lists to be involved here and maybe I'm misunderstanding Art's comment, but mentioning this just in case it's relevant.

ArtOfCode-:
"list" is just the name of the view used here, @cellio - it's not part of a list, just using the same format

rileyradoll:
@ArtOfCode- @cellio thanks for the feedback and context. I'll get this changed this weekend.

cellio:
Hi, just checking in.  Were you able to make those changes?  Thanks.

rileyradoll:
Still working on this. Will have something soon.

rileyradoll:
@ArtOfCode- Made some changes based on your feedback. Let me know if you'd like to see more changes. Thanks!

ArtOfCode-:
Thanks!

PR #273, opened 2020-11-08
Implement table of contents

### LINKED ISSUE #224: Support answer summaries or leaderboard (needed by code golf, wanted more generally)
Code Golf needs a leaderboard for its challenges.  (On SE they used a userscript that depends on the SE API.  Our API isn't a swap-in replacement.)  Specifically, the leaderboard appears at the bottom of the "question" (challenge), summarizing answers, ordered by byte count and also showing language.  We have previously talked about adding answer summaries (using scores, not byte count), so ideally we will do something for CG that advances our general need too.

Possibilities:  

0. Add the concept of a "TOC" on a top-level post.  The default TOC summarizes answers in score order as described in use cases and wireframes elsewhere.  (Primitive wireframe sketch: https://www.figma.com/file/AwmzxwL9KNFfDHdnTHcrv1/expand-answers?node-id=0%3A1).  The default TOC orders by score and displays upvotes, downvotes, first ~300 characters, author, last-modified date.  So long as golf challenge answers follow a fixed format, length and language will appear in the summaries, but the list will be sorted by score not length.  **Per discussion in comments**, this is the approach we'll take for now, and we'll come back to custom sorting later.  Difficulty assessment is based on this decision.

1. Introduce the idea of a custom TOC type, settable per post type per category, where admins can specify what to order by and what fields to display.  Might require a custom answer type (to supply those fields).   Vague idea for later.
 
2. Make a new post type for challenges.  A challenge is just like a question; a challenge answer has additional fields that need to be filled in (size and language), and the challenge uses that info to add a leaderboard (probably just size, language, link).  We might need a "challenge answer" post type anyway to support #1, since otherwise the TOC code would have to parse answers.  (I'm not fluent enough in the code to evaluate that trade-off.)  Vague idea for later.

#1 is appealing because, ooh, generality, but it's also hard and I don't yet know if we have a third use case.  (It's better to design general interfaces with more than one or two cases in mind.)  #2 is probably easier *except* that we encode post type in a bunch of places -- URLs, category configuration, abilities, probably more -- so it might bite us in the end.  On the third hand, maybe the post customization we need is only for *answers*, and a "golf answer" could be used with a (general) "question" post type, reducing the splash damage.  On the fourth hand, I'd bet that'll break something else...

Daksh14:
Closes #224 implement table of contents and other site params


app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
...
+                <div class="toc--full">
+                  <% sanitised = raw(sanitize(answer.body, scrubber: scrubber)) %>
+                  <% max_length = SiteSetting["TableOfContentsExcerptLength"] %>
+                  <p>
+                    <%=raw sanitised[ 0 ... max_length ] %>
+                    <% if sanitised.length > max_length %>
+                      ...
+                    <% end %>
+                  </p>
+                  <p class="has-font-size-small"><%= time_ago_in_words(post.last_edited_at, locale: :en_abbrev) %> ago</p>
```
luap42, line 316: [**issue:**](https://conventionalcomments.org/) This won't work on most setups, because the last_edited_at value is only set for some posts with recent edits, because it's only been added recently (for the "updated X days ago by" labels) and is hence limited to some specific actions.

Consider using `updated_at` here.

app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
           </div>
         <% end %>
 
+        <% if is_question && post.children.length >= SiteSetting["TableOfContentsThreshold"] %>
+          <div class="toc" id="toc-example-toggle">
+            <button class="toc--header" data-toggle="#toc-example-toggle" data-toggle-property="class" data-toggle-value="is-active">Table of Contents</button>
+            <% post.children.each do |answer| %>
+              <a href="#" class="toc--entry">
```
luap42, line 300: [**issue:**](https://conventionalcomments.org/) The ToC-entry should probably link to the answer. You can use the helper `generic_show_link(POST)`.

app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
           </div>
         <% end %>
 
+        <% if is_question && post.children.length >= SiteSetting["TableOfContentsThreshold"] %>
+          <div class="toc" id="toc-example-toggle">
+            <button class="toc--header" data-toggle="#toc-example-toggle" data-toggle-property="class" data-toggle-value="is-active">Table of Contents</button>
+            <% post.children.each do |answer| %>
```
luap42, line 299: [**praise:**](https://conventionalcomments.org/) Good idea to use post.children instead of generating a new database query in the controller (which was my original idea TBF). This is definitely cleaner and faster.

app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
           </div>
         <% end %>
 
+        <% if is_question && post.children.length >= SiteSetting["TableOfContentsThreshold"] %>
```
luap42, line 296: [**suggestion (non-blocking):**](https://conventionalcomments.org/) Add a third check here, which checks whether the Threshold is not equal to `-1` to allow disabling the ToC. Non-blocking because for now we could achieve the same effect with setting an unreasonably high answer count as threshold, but having an "official value" would be nicer of course.

app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
...
+          <div class="toc" id="toc-example-toggle">
+            <button class="toc--header" data-toggle="#toc-example-toggle" data-toggle-property="class" data-toggle-value="is-active">Table of Contents</button>
```
luap42, line 298: [**nitpick:**](https://conventionalcomments.org/) Please rename the ID to something more specific; this is not an example. You can assume that the TOC is shown only once on a page, so you don't have to add anything post-specific.

luap42, line 298: [**suggestion (if-minor):**](https://conventionalcomments.org/) It would be super-awesome if communities could customize the title of the TOC. For example Code Golf might want to name it "Leaderboard".

app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
           </div>
         <% end %>
 
+        <% if is_question && post.children.length >= SiteSetting["TableOfContentsThreshold"] %>
+          <div class="toc" id="toc-example-toggle">
```
luap42, line 297: [**nitpick (non-blocking):**](https://conventionalcomments.org/) I don't know if that's just me, but the TOC feels somehow off. Please consider adding some left margin, so that it aligns better with the post tools and the comments.

screenshot demonstrating the issue

app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
...
+                  <span class="badge is-tag is-green">+<%= answer.upvote_count %></span>
+                </div>
+                <div class="toc--badge">
+                  <span class="badge is-tag is-red">&minus;<%= answer.downvote_count %></span>
+                </div>
+                <div class="toc--full">
+                  <% sanitised = raw(sanitize(answer.body, scrubber: scrubber)) %>
+                  <% max_length = SiteSetting["TableOfContentsExcerptLength"] %>
+                  <p>
+                    <%=raw sanitised[ 0 ... max_length ] %>
```
luap42, line 311: [**info:**](https://conventionalcomments.org/) In latest Rails versions, `raw` makes an output being send as-is, so without stripping/escaping HTML. Not sure if that is what you want.

(PS: Also, the raw is doubled, there is already one in line 308. So this can probably go)

app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
...
+            <% post.children.each do |answer| %>
+              <a href="#" class="toc--entry">
+                <div class="toc--badge">
+                  <span class="badge is-tag is-green">+<%= answer.upvote_count %></span>
+                </div>
+                <div class="toc--badge">
+                  <span class="badge is-tag is-red">&minus;<%= answer.downvote_count %></span>
+                </div>
+                <div class="toc--full">
+                  <% sanitised = raw(sanitize(answer.body, scrubber: scrubber)) %>
```
luap42, line 308: [**issue (blocking):**](https://conventionalcomments.org/) This isn't enough. It does prevent malicious input, but we also want to block most Markdown here. Either just send the Markdown or use the `strip_markdown` helper to strip all the markdown.

This will also make formatting easier, because the output will no longer be put into `<p>` tags, but you can fully decide that yourself.

app/views/posts/_expanded.html.erb
```diff
@@ -293,6 +293,33 @@
...
+                  <% max_length = SiteSetting["TableOfContentsExcerptLength"] %>
+                  <p>
+                    <%=raw sanitised[ 0 ... max_length ] %>
+                    <% if sanitised.length > max_length %>
+                      ...
```
luap42, line 313: [**nitpick (non-blocking):**](https://conventionalcomments.org/) It may be better to use the HTML `&hellip;` entity for semantical/accessibility reasons. OTOH we aren't doing it elsewhere, so this is probably fine.

db/schema.rb
```diff
@@ -280,7 +280,7 @@
     t.integer "post_type_id", null: false
     t.text "body_markdown"
     t.integer "answer_count", default: 0, null: false
-    t.datetime "last_activity", default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.datetime "last_activity", default: -> { "current_timestamp()" }, null: false
```
luap42, line 283: [**chore:**](https://conventionalcomments.org/) Please revert this manually. This change happens automatically sometimes, but we don't want it.

db/seeds/site_settings.yml
```diff
@@ -313,3 +313,17 @@
   category: AdvancedSettings
   description: >
     The minimum score a question must have to qualify for selection for the Hot Posts sidebar and for being selected as random advertisement.
+
+- name: TableOfContentsThreshold
+  value: 5
+  value_type: integer
+  category: Display
+  description: >
+    The number of child posts before the TOC is shown. If the threashold is not met, do not show TOC at all.
```
luap42, line 322: I'll defer language to @cellio on this one, but please fix the spelling for "threashold" :)

db/seeds/site_settings.yml
```diff
@@ -313,3 +313,17 @@
...
+
+- name: TableOfContentsExcerptLength
+  value: 120
+  value_type: integer
+  category: Display
+  description: >
+    The length of the answer shown in the TOC, don't keep it too large, it may fill up the TOC and become unreadable.
```
luap42, line 329: Same here, but I'd start a new sentence after the first "TOC".

luap42 (changes requested):
Looks good so far, just some smaller changes needed. Don't worry about the many comments; most of them are nitpicky or easily solved.

app/views/posts/_expanded.html.erb
ArtOfCode-: Seconded. Not a security issue, but you'd be better off using `strip_markdown(answer.body_markdown)[0..SiteSetting['TableOfContentsExcerptLength']]`

db/schema.rb
ArtOfCode-: Eh, non-blocking. They're both equivalent as far as MySQL is concerned.

app/views/posts/_expanded.html.erb
cellio: @luap42 what would it take to eliminate that inconsistency?  Would a run through the DB to set a missing last_edited_at to the value of whatever that other field is fix this once and for all?

db/seeds/site_settings.yml
cellio: The minimum number of child posts before the table of contents is shown.  If the number of child posts is less than this threshold, the TOC is not shown.

(Expand TOC on first use, unless there's already context for that. Rest is just wordsmithing.)

db/seeds/site_settings.yml
cellio: The number of characters to show in the answer excerpt in the TOC.  The TOC strips most formatting, so beware of large values -- the TOC might become hard to read.

--

Nitpick: the earlier message refers to "child posts"; this one refers to "answers".  Please pick one and use it in both places.

app/views/posts/_expanded.html.erb
Daksh14: I'm sure that would be a good thing but I'll refrain from that right now since we are keeping things at a simpler side

app/views/posts/_expanded.html.erb
luap42: @cellio I think it may be possible by setting last_edited_at to updated_at,. Although it is likely hard to tell, whether that's what we want, because last_edited_at implies something different, namely a proper post edit, while updated_at and last_activity have different meanings (updated_at: the database entry changed, last_activity: something happened to the post or its children). And I'm not sure whether anything else bases on last_edited_at being set correctly.

app/views/posts/_expanded.html.erb
luap42: 👍 Makes sense. We can add a site setting for that later, too.

app/views/posts/_expanded.html.erb
cellio: @luap42 oh I see.  I misunderstood.  Yeah, if it's not completely clear and unambiguous, better to leave the DB alone and adjust the code.

cellio:
What is this PR waiting on?  Does @luap42 need to look at changes?  Are there changes we still need?

ArtOfCode- (approved):

Issue #255, opened 2020-11-04
Misalignment in edit history (overlap or smushed text and some vertical skew)


cellio:
https://meta.codidact.com/questions/278949

> In the post history, the "by..." and edit reason are misaligned from the "Post edited" (more obvious when there is an edit reason)

> E.g. https://meta.codidact.com/posts/278406/history

galegosimpatico:
https://github.com/codidact/qpixel/commit/9b9c49d2e6e21232717ef760c76c3612e111c28b#commitcomment-44049323

PR #282, opened 2020-11-11
Adds nbsp before 'by' in post history

mattjbrent:
This is a super small fix for a small annoyance i've had on the site where the spacing between by and the edit type was missing. This addresses that.
broken: 
Screen Shot 2020-11-11 at 4 39 44 PM

Fixed:
Screen Shot 2020-11-11 at 4 40 17 PM


cellio:
Yay!  Looks good to me.

There's an issue for this: https://github.com/codidact/qpixel/issues/255

PR #302, opened 2020-11-18
Add sorting to contents, move toc entry to a partial file

### LINKED ISSUE #295: Table of contents shows deleted answers
https://meta.codidact.com/questions/279157

I know this is already being worked on, but filing this for tracking back to the meta report when it gets fixed/deployed.

Daksh14:
Sort TOC contents according to pre-calculated score in the db and move TOC entries to a partial file.
Closes  #295

### DIFF
--- a/app/views/posts/_expanded.html.erb
+++ b/app/views/posts/_expanded.html.erb
@@ -293,7 +293,7 @@
           </div>
         <% end %>
 
-        <% if is_question %>
+        <% if is_question && post.children.length >= SiteSetting["TableOfContentsThreshold"] && SiteSetting["TableOfContentsThreshold"] != -1 %>
           <div class="toc has-margin-left-4" id="toc-toggle">
             <button class="toc--header" data-toggle="#toc-toggle" data-toggle-property="class" data-toggle-value="is-active">Show More</button>
             <% sorted_answers = post.children.sort_by { |answer| answer.score }.reverse! %>
--- a/app/views/posts/_expanded.html.erb
+++ b/app/views/posts/_expanded.html.erb
@@ -293,29 +293,13 @@
           </div>
         <% end %>
 
-        <% if is_question && post.children.length >= SiteSetting["TableOfContentsThreshold"] && SiteSetting["TableOfContentsThreshold"] != -1 %>
+        <% if is_question %>
           <div class="toc has-margin-left-4" id="toc-toggle">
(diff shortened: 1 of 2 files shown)


cellio:
Where do we stand on this -- need review?

app/views/posts/_expanded.html.erb
```diff
@@ -295,27 +295,11 @@
...
-                      &hellip;
-                    <% end %>
-                  </p>
-                  <p class="has-font-size-small"><%= time_ago_in_words(post.updated_at, locale: :en_abbrev) %> ago</p>
-                </div>
-              </a>
+            <button class="toc--header" data-toggle="#toc-toggle" data-toggle-property="class" data-toggle-value="is-active">Show More</button>
+            <% sorted_answers = post.children.sort_by { |answer| answer.score }.reverse! %>
+            <% sorted_answers.each do |answer| %>
+              <% next if answer.deleted? && !moderator? %>
```
ArtOfCode-, line 301: nitpick (non-blocking): this _only_ allows moderators to see deleted posts, not all users with the privileges to view deleted. Maybe that's what we want; don't know.

app/views/posts/_toc_entry.html.erb
```diff
@@ -0,0 +1,19 @@
+<a href="<%= generic_share_link(answer) %>" class="toc--entry <%= 'deleted-content' if answer.deleted? %>">
```
ArtOfCode-, line 1: suggestion: this should use a link_to helper:

```suggestion
<%= link_to generic_share_link(answer), class: "toc--entry #{answer.deleted? ? 'deleted-content' : ''}" do %>
  ... (rest of this HTML goes here) ...
<% end %>  
```

ArtOfCode- (changes requested):

cellio:
Where are we on this?  I saw some chat discussion; @Daksh14 do you have everything you need or are there open questions?  Thanks.

app/views/posts/_expanded.html.erb
cellio: For consistency, it'd be better if people who can see deleted posts farther down the page could also see them here.  Otherwise we're going to confuse people and probably get bug reports.  We'd want them to go to the bottom, like on the main page, though -- is that hard?

ArtOfCode- (changes requested):

PR #304, opened 2020-11-19
Fix user avatar overflow on non-standard image dimensions

### LINKED ISSUE #291: Avatar overflow: user avatars can exceed intended space and affect spacing
See the overly-tall avatar in the second row:

screenshot
 
I don't know what causes *this* one to be tall, where most others are (I think) resized.  I think I've seen some others that were either wider or taller than the intended space.  

I don't think we want avatars that are "too tall" or "too wide" for our intended design.  Ideally we would catch this at image-upload time and either resize or crop, but that's more complex.  When *displaying* avatars, should we enforce the bounding box we want somehow?

Technohacker:
Fixes #291 

image
image

### DIFF
--- a/app/assets/stylesheets/posts.scss
+++ b/app/assets/stylesheets/posts.scss
@@ -105,6 +105,12 @@ h1 .badge.is-tag.is-master-tag {
     display: flex;
     flex-direction: column;
     align-self: flex-end;
+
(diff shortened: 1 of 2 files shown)


ArtOfCode-:
Thanks!

PR #334, opened 2020-12-06
Post types

ArtOfCode-:
We're currently using separate controllers, routes, views - the lot - for each different type of post. It works, but it's not tenable for any future additions. This PR moves as much of the separated stuff as possible back under a single Post umbrella.

This changes a number of assumptions that we'd made about how things operate, so it may break some stuff, although I'm adding tests as I go. This also changes our URL schemes: `/questions/:id` and `/articles/:id` both become `/posts/:id`. Answers never had their own top-level URLs to get to them, except for the "share" URLs (`/a/123/456`). I'll redirect existing URLs to the new path so that things don't break.

Still to do:

1. [x] Remove tests for removed actions
2. [x] Tests for new actions
    * [x]    new: simple GET works, all variants
    * [x]    new: requires auth, all variants
    * [x] create: POST with correct parameters works, all variants
    * [x] create: requires auth, all variants
    * [x] create: category post type rejects without category
    * [x] create: category post type checks required trust level
    * [x] create: parented post type rejects without parent
    * [x] create: ensures community_user is created
    * [x] create: help/policy checks permissions
    * [x]   show: simple GET works, no deleted answers
    * [x]   show: parented post type redirects to parent post
    * [x]   show: deleted post checks permissions
    * [x]   show: user with permissions sees deleted answers
    * [x]   edit: simple GET works
    * [x]   edit: requires auth
    * [x] update: POST with correct parameters works, generates history event
    * [x] update: POST with correct parameters, unprivileged user, works, generates suggested edit
    * [x] update: requires auth
    * [x] update: rejects no-change edit
3. [x] More actions & relevant tests
    * [x] Close/Reopen
    * [x] Delete/Undelete
4. [x] Fix all the remaining broken things
5. [x] ???
6. [x] Profit
7. [ ] **After deploy**: create nginx redirects for `/q/:id`, `/a/:qid/:id`, `/ar/:id`, `/p/:id`, `/questions/:id`, `/articles/:id`


cellio:
I'm very glad to see us addressing this.  I didn't know how hard it would be so tried not to push hard, but a general "post" URL will make some things much easier.  In particular, it seems weird that you would need to know the type of a post in order to access it.  If we can redirect all the extant URLs and then use /post going forward, that'll be great!  I assume this will also simplify the addition of other post types later when we need them.

ArtOfCode-:
It's not a simple task, but either we do it now or we do it two or three post types down the line, and I already know which one of those is more of a PITA. Yes, it'll also simplify adding post types - add a record in the database and _possibly_ a model file, but it might just need the record.

ArtOfCode-:
Tested and ready for review. There are CI tests on every page that includes one of the new post forms, so any errors will be caught. I've also tested them all manually to make sure the UI is correct and that they work as intended.

app/controllers/advertisement_controller.rb
```diff
@@ -197,7 +197,7 @@ def select_random_post(category = nil)
     if category.nil?
       category = Category.where(use_for_advertisement: true)
     end
-    Post.undeleted.where(last_activity: (Rails.env.development? ? 365 : 7).days.ago..Time.now)
+    Post.undeleted.where(last_activity: (Rails.env.development? ? 365 : 7).days.ago..Time.zone.now)
```
luap42, line 200: Question (curiosity): What's the `.zone` doing here?

app/controllers/application_controller.rb
```diff
@@ -184,8 +206,8 @@ def setup_request_context
 
     Rails.logger.info "  Host #{host_name}, community ##{RequestContext.community_id} " \
                       "(#{RequestContext.community&.name})"
-    unless RequestContext.community.present?
-      render status: 422, plain: "No community record matching Host='#{host_name}'"
+    if RequestContext.community.blank?
```
luap42, line 209: Question (curiosity): Why this change?

app/controllers/questions_controller.rb
```diff
@@ -57,129 +11,6 @@ def lottery
...
-    if @question.update(deleted: false, deleted_at: nil, deleted_by: nil,
-                        last_activity: DateTime.now, last_activity_by: current_user)
-      PostHistory.post_undeleted(@question, current_user)
-    else
-      flash[:danger] = "Can't undelete this question right now. Try again later."
-    end
-    redirect_to url_for(controller: :questions, action: :show, id: @question.id)
-  end
-
   def feed
```
luap42, line 14: Question: Is there a specific reason to have (a) this method in Questions controller and (b) not for Articles (controller was deleted)

app/models/notification.rb
```diff
@@ -2,7 +2,5 @@ class Notification < ApplicationRecord
   include CommunityRelated
   belongs_to :user
 
-  def community_name
-    community.name
-  end
+  delegate :name, to: :community, prefix: true
```
luap42, line 5: Quick check: Do the notifications still work after that? (because it may be neccessary for the JSON-encoder to have it like in the old version)

luap42 (approved):
Looks good. A few curiosity/clarification questions, but besides that, ready to go.

app/controllers/advertisement_controller.rb
ArtOfCode-: ...good question, blaming Rubocop. Should be `DateTime.now` really. Will fix.

app/controllers/application_controller.rb
ArtOfCode-: Rubocop... don't ask me _why_ it prefers this way :P

app/controllers/questions_controller.rb
ArtOfCode-: This is where the RSS feed is sourced from. RSS feeds aren't something I've touched in this PR, so it's just left as-is. May need reviewing in the future to see if it still does what we want it to.

app/models/notification.rb
ArtOfCode-: Also a Rubocop change - but yes, this doesn't change the method name, it's still `community_name`

PR #338, opened 2020-12-07
Fix broken link

Mithrandir24601:
Fix link from (broken) `asking` to `how-to-ask`.


luap42 (approved):

cellio:
Does something in the docs repo need to get updated to (for future pushes/seeds), or was it only wrong here?

PR #378, opened 2020-12-27
Fix suggested edit notification URL

### LINKED ISSUE #373: Notification for suggested edit uses local community's URL slug, not where the edit is
https://meta.codidact.com/posts/280086

If you click on the notification for a suggested edit, and that suggested edit is on a different community than the one you're presently on, the link is wrong (usually 404) because it uses the local slug instead of the one for the correct community.

I'm pretty sure this is a recent regression; I'm usually parked on meta and approved an edit on Judaism a few days ago because I was notified.  I could be misremembering where I saw the notification, but we also haven't gotten reports of bad links before now.  I just found that notification and followed the link from meta, reproducing this bug.

sau226:
This PR fixes a suggested edit notification by using a URL instead of a site path.

I tested the PR by adding a second community on my local dev box and this patch works, at least for me.

Issue linking tag: resolves #373

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -184,7 +184,7 @@ class PostsController < ApplicationController
           if @post_type.has_parent
             message += " on '#{@post.parent.title}'"
           end
-          @post.user.create_notification message, suggested_edit_path(edit)
+          @post.user.create_notification message, suggested_edit_url(edit, host: @post.community.host)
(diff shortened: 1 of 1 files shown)


cellio:
@ArtOfCode- or @luap42 , could you please approve this one-line bug fix?  Thanks.

ArtOfCode- (approved):

PR #396, opened 2021-01-01
Fix incorrect grammar in close reasons

sau226:
I think one of the close reasons has slightly incorrect grammar - this PR fixes that instance of incorrect grammar.

Happy for this to be closed and change directly made by a dev with write access.


cellio:
I agree with this change.  @luap42 does anything else need to be done other than updating this seed file?

ArtOfCode-:
Merged and live reasons updated.

PR #398, opened 2021-01-02
Comments

luap42:
\o/

Comments are going to be *waaay* more fun.

Whoever finds an extra comma, may keep, sell, or return it..


ArtOfCode-:
> Whoever finds an extra comma, may keep, sell, or return it..

Found one!

app/assets/javascripts/comments.js
```diff
@@ -129,6 +79,30 @@ $(() => {
...
+
+    const $tgt = $(evt.target);
+    const threadID = $tgt.data("thread")
+    const action = $tgt.data("action")
+    const route = $tgt.hasClass("js--restrict-thread") ? 'restrict' : 'unrestrict';
+
+    const resp = await fetch(`/comments/thread/${threadID}/${route}`, {
+      method: 'POST',
+      credentials: 'include',
+      headers: { 'X-CSRF-Token': QPixel.csrfToken(), 'Content-Type': 'application/x-www-form-urlencoded; charset=utf-8' },
```
ArtOfCode-, line 93: **suggestion** (code-clarity): Usually more clear to use `application/json` - Rails will accept it either way, but means you can use `JSON.stringify({ your: 'data', goes: 'here' })` for the `body`.

app/controllers/comments_controller.rb
```diff
@@ -77,6 +101,8 @@ def update
 
   def destroy
     if @comment.update(deleted: true)
+      @comment_thread = @comment.comment_thread
+      @comment_thread.update(reply_count: @comment_thread.comments.undeleted.size)
```
ArtOfCode-, line 105: **suggestion** (rails, code-clarity): This (and any equivalents, like L119) would be better handled by a counter_cache in the model - see Post for an example:

```ruby
counter_culture :parent, column_name: proc { |model| model.deleted? ? nil : 'answer_count' }
```

I *think* what you'd need is a similar line added to Comment:

```ruby
counter_culture :comment_thread, column_name: proc { |model| model.deleted? ? nil : 'reply_count' }
```

app/controllers/comments_controller.rb
```diff
@@ -106,15 +134,84 @@ def show
...
+      @comment_thread.update(locked: true, locked_by: current_user, locked_until: lu)
+
+      redirect_to comment_thread_path(@comment_thread.id)
+      return
+    when 'archive'
+      @comment_thread.update(archived: true, archived_by: current_user)
+    when 'delete'
+      @comment_thread.update(deleted: true, deleted_by: current_user)
+    when 'follow'
+      ThreadFollower.create comment_thread: @comment_thread, user: current_user
```
ArtOfCode-, line 171: **question**: This combined with the first line of this method seems to mean that only users with the flag_curate privilege can follow a thread? Is that intentional?

app/controllers/comments_controller.rb
```diff
@@ -128,25 +225,80 @@ def set_comment
...
+    end
+  end
+
+  def comment_rate_limited
+    recent_comments = Comment.where(created_at: 24.hours.ago..DateTime.now, user: current_user).where \
+                             .not(post: Post.includes(:parent).where(parents_posts: { user_id: current_user.id })) \
+                             .where.not(post: Post.where(user_id: current_user.id)).count
+    max_comments_per_day = SiteSetting[current_user.privilege?('unrestricted') ? 'RL_Comments' : 'RL_NewUserComments']
+
+    # Provides mainly web actions for using and making comments.
```
ArtOfCode-, line 285: You've been Rubocop'd!

app/helpers/comments_helper.rb
```diff
@@ -1,10 +1,66 @@
...
+  def comment_link(comment)
+    if comment.deleted
+      comment_thread_path(comment.comment_thread_id) + "?show_deleted_comments=1#comment-#{comment.id}"
+    else
+      comment_thread_path(comment.comment_thread_id) + "#comment-#{comment.id}"
+    end
+  end
+
+  def render_pings(comment)
+    comment.gsub(/\[\[PING [0-9]+\]\]/) do |id|
```
ArtOfCode-, line 11: **suggestion** (non-blocking): `\d` instead of `[0-9]`

app/views/comments/_new_thread_modal.html.erb
```diff
@@ -0,0 +1,29 @@
+<div class="modal js--<%= post.id %>-new-thread-modal is-with-backdrop">
+  <div class="modal--container">
+    <div class="modal--header">
+      <button class="button is-close-button modal--header-button" data-modal=".js--<%= post.id %>-new-thread-modal">&times;</button>
+      Start new comment thread
+    </div>
+    <div class="modal--body">
+      <% if post.comment_threads.size > 0 %>
+        <p>You can create a new comment thread. Please check first, whether an existing comment thread covers already what you need.</p>
```
ArtOfCode-, line 9: comma bingo: 1 (no comma after "first")

app/views/comments/thread.html.erb
```diff
@@ -0,0 +1,182 @@
...
+          
+          <%= label_tag :content, 'Your message', class: 'form-element' %>
+          <%= text_area_tag :content, '', class: 'form-element' %>
+          
+          <%= submit_tag 'Add reply', class: 'button is-muted is-filled' %>
+        <% end %>
+      <% elsif @comment_thread.deleted %>
+        <p class="h-c-red-700"><i class="fas fa-trash fa-fw"></i> You cannot reply to this thread, because it has been deleted.</p>
+      <% elsif @comment_thread.locked %>
+        <p class="h-c-red-700"><i class="fas fa-lock fa-fw"></i> You cannot reply to this thread, because it has been locked.</p>
```
ArtOfCode-, line 101: comma bingo: 2 and 3

scripts/create_general_thread_for_old_comments.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  comments = comments.all
+
+  new_thread = CommentThread.new(title: 'General comments', post: post, reply_count: comments.size, locked: false,
+                                 archived: false, deleted: false, community: post.community)
+  new_thread.save
+
+  comments.each do |c|
+    c.update comment_thread: new_thread
+  end
```
ArtOfCode-, line 13: **suggestion** (performance):

```ruby
comments.update_all(comment_thread_id: new_thread.id)
```

scripts/create_general_thread_for_old_comments.rb
```diff
@@ -0,0 +1,16 @@
+Post.unscoped.each do |post|
```
ArtOfCode-, line 1: **suggestion** (performance): filter out the posts we don't need to check in SQL

```ruby
Post.unscoped.where(id: Comment.unscoped.where(comment_thread: nil)
                               .select(Arel.sql('distinct post_id'))).each do |post|
```

ArtOfCode- (changes requested):

PR #416, opened 2021-01-14
Stop same post being used in close reasons for itself

### LINKED ISSUE #218: A question can be closed as a duplicate of itself
https://meta.codidact.com/questions/278655

While detecting and handling duplicate loops is hard, can we detect this special case and not allow a question to be closed as a duplicate of itself?

sau226:
This PR blocks a post from being used to close itself as a duplicate, as well as for any other situation requiring a post as part of the close reason.

Currently, this just displays an HTTP 400 error, with details in JSON. Feel free to improve on the UX here, but it has the desired effect already.

GitHub issue linking tag: Fixes #218

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -229,7 +229,7 @@ class PostsController < ApplicationController
         return
       end
       
-      if other = @post
+      if other == @post
         render json: { status: 'failed', message: 'You can not close a post as a duplicate of itself' }, status: :bad_request
         return
       end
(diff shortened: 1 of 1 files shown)


cellio:
What does the user see when triggering this error?  One of those red banner pop-ups?  With what text?

app/controllers/posts_controller.rb
```diff
@@ -228,6 +228,11 @@ def close
         render json: { status: 'failed', message: 'Invalid input for other post.' }, status: :bad_request
         return
       end
+      
+      if other = @post
```
ArtOfCode-, line 232: **error**: you want `other == @post` here - i.e. equality check, rather than assignment

ArtOfCode- (changes requested):

sau226:
> What does the user see when triggering this error? One of those red banner pop-ups?

Correct - the user gets a 400 error pop up (maybe with a timestamp) - the actual error is visible in browser console preview and afaik we do something similar for most exceptions like this.

PR #422, opened 2021-01-17
Discard edits when commenting

### LINKED ISSUE #243: Add a way to cancel a comment edit
https://meta.codidact.com/questions/278770

You can now discard a *new* comment (there are buttons for "post" and "discard"), but you can't cancel *editing* a comment -- the only button during an edit is "post", so it's either that or refresh the page.

Could we add the other option when editing a comment?  For clarity, it should probably say "discard edit" rather than just "discard" (which could be interpreted as "delete").

I'm marking this "easy" because I *think* the fix is to copy this code from "post comment" and adjust it for "edit comment".  If there's something that makes it more complicated, feel free to change the label or comment about it.

BoxedFruits:
This PR resolves #243 to discard edits when commenting.

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -86,14 +86,14 @@ $(() => {
         </div>
         <div class="actions">
           <input type="submit" class="button is-outlined" value="Post" />
-          <input type="button" name="js-discard-edit" id="js-discard-edit-${commentId}" value="Discard Edit" class="button is-danger is-outlined" />
+          <input type="button" name="js-discard-edit" value="Discard Edit" class="button is-danger is-outlined js-discard-edit-${commentId}" />
         </div>
       </div>
     </form>`;
 
     $comment.html(formTemplate);
 
-    $(`#js-discard-edit-${commentId}`).click(() => {
+    $(`.js-discard-edit-${commentId}`).click(() => {
       $comment.html(originalComment);
     });
   });
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -69,6 +69,7 @@ $(() => {
     const $tgt = $(evt.target);
     const $comment = $tgt.parents('.comment');
     const commentId = $comment.attr('data-id');
+    const originalComment= $comment.find('p.comment--content').clone();
 
     const resp = await fetch(`/comments/${commentId}`, {
       credentials: 'include',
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/comments.js
```diff
@@ -85,11 +86,16 @@ $(() => {
         </div>
         <div class="actions">
           <input type="submit" class="button is-outlined" value="Post" />
+          <input type="button" name="js-discard-edit" id="js-discard-edit-${commentId}" value="Discard Edit" class="button is-danger is-outlined" />
```
ArtOfCode-, line 89: Can we make this `class="js-discard-edit" data-comment-id="${commentId}"` instead of packing it all into the ID attribute?

app/assets/javascripts/comments.js
```diff
@@ -85,11 +86,16 @@ $(() => {
...
         </div>
       </div>
     </form>`;
 
     $comment.html(formTemplate);
+
+    $(`#js-discard-edit-${commentId}`).click(() => {
```
ArtOfCode-, line 96: This would need updating to use the class attribute as well.

ArtOfCode- (changes requested):

ArtOfCode- (changes requested):
Should've made myself more clear, sorry. I'd like this to use the class `js-discard-edit`, with _no_ ID on the end of it, and then an _additional_ `data-comment-id` attribute where the comment ID can go. The ckick handler can then either select on all `.js-discard-edit` and deal with all in one handler, or it can select on `.js-discard-edit[data-comment-id="${commentId}"]` if it's easier to handle each one individually.

BoxedFruits:
> Should've made myself more clear, sorry. I'd like this to use the class `js-discard-edit`, with _no_ ID on the end of it, and then an _additional_ `data-comment-id` attribute where the comment ID can go. The ckick handler can then either select on all `.js-discard-edit` and deal with all in one handler, or it can select on `.js-discard-edit[data-comment-id="${commentId}"]` if it's easier to handle each one individually.

Ohhhh gotcha. I'll update in a bit!

ArtOfCode- (approved):

ArtOfCode-:
Thanks! There's a few more [good first issues](https://github.com/codidact/qpixel/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22meta%3A+good+first+issue%22) or [low-complexity tasks](https://github.com/codidact/qpixel/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22complexity%3A+easy%22) if you're looking for more to do.

PR #424, opened 2021-01-23
to make it easier to translate

dyaskur:
If no problem with this, I will do more like this.

Sorry if any wrong/bad practice, feel free to give criticism and suggestions

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,7 +13,7 @@ class ApplicationController < ActionController::Base
   before_action :enforce_2fa
   before_action :block_write_request, if: :read_only_mode?
 
-  helper_method :top_level_post_types, :second_level_post_types, :helpers
+  helper_method :top_level_post_types, :second_level_post_types
 
   def upload
     redirect_to helpers.upload_remote_url(params[:key])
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -13,7 +13,7 @@ class ApplicationController < ActionController::Base
   before_action :enforce_2fa
   before_action :block_write_request, if: :read_only_mode?
 
-  helper_method :top_level_post_types, :second_level_post_types
```
ArtOfCode-, line 16: You don't actually need this; helpers are available in views by default. You can remove this and just call `i18ns` directly in the views.

ArtOfCode- (changes requested):
Looks good, with one slight change needed. Making things easier to translate and people who can provide translations are definitely things we need, we just haven't had time to do it ourselves.

app/controllers/application_controller.rb
dyaskur: removed, I see I can use `t` method too, so I changed `i18ns` to `t` .
Thanks

ArtOfCode- (approved):

ArtOfCode-:
Thanks!

cellio:
Thanks @dyaskur!  What should folks know going forward, to avoid regressing?  For example, if we add new UI elements that have text, what's the right way to do it?  (The question on collab.codidact.org about this might be a good place to document best practices for new code.)

Issue #387, opened 2020-12-29
Exact search on a post title doesn't find the post


cellio:
https://meta.codidact.com/posts/279392

I reproduced the problem by searching for the linked post by its title in quotes.  No hits.  When I removed the quotes I got a bunch of hits, but this post wasn't near the top.

I wondered if stop words were interfering, so on Meta Blog I searched for "Newsletter #1 (May 2020)" and didn't get any hits either.  I also tried "Resource section" on Judaism Meta.  This rules out stop words as culprits.

BoxedFruits:
I'd like to work on this!

Also, I'm going to post my research here because I am not too familiar with Ruby on Rails yet and my SQL is rusty so I would appreciate any advice/help.

I looked in the logs and it looks like its only matching using **body_markdown** when doing a search:
```sql
 SELECT  `posts`.*, MATCH (`posts`.`body_markdown`) AGAINST ('searchTerm' IN BOOLEAN MODE) AS search_score FROM `posts` WHERE `posts`.`community_id` = 1 AND `posts`.`post_type_id` IN (1, 2, 5) AND (MATCH (`posts`.`body_markdown`) AGAINST ('searchTerm' IN BOOLEAN MODE)) ORDER BY `search_score` DESC LIMIT 25 OFFSET 0
```
We need to match using both the body_markdown and title so in MySQL I executed this command. **I don't know the implications of doing this to the prod DB.**
```sql
ALTER TABLE posts ADD FULLTEXT (title,body_markdown);
```

The function that parses the search term is in **posts.rb**
```Ruby
  def self.search(term)
    match_search term, posts: :body_markdown
  end
```
In order to match with both body_markdown and title we change the third line to: 
```Ruby
match_search term, posts: [:body_markdown, :title]
```
Exact matches for title and body_markdown now work... but not non-exact matches and I am not too sure why.

cellio:
Thanks for looking into this!

Why is the additional column needed in the table if you can modify match_search to look in both the body and the title?  I'm confused by the DB change.

manassehkatz:
> 
> 
> Thanks for looking into this!
> 
> Why is the additional column needed in the table if you can modify match_search to look in both the body and the title? I'm confused by the DB change.

The DB change is to make search speed reasonable by indexing the fields. However, it should be fine to index each field independently, and I think that would be better for a number of reasons.
**Do current indexes include title, body_markdown, neither one or both?**

cellio:
[...]

PR #428, opened 2021-02-07
[WIP] Exact search on a post title

BoxedFruits:
Working on addressing #387.

Previously the search function was only using body_markdown. I changed it so that it also uses the title for the search. Because of this, the sql statement got a little more complex. I am trying to have the title's search score weigh more than body_markdown and was having a bit of trouble figuring out how to do that without hard coding it. Wondering if there is any way I could improve this before opening up the PR for review.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -8,7 +8,7 @@ class SearchController < ApplicationController
                        .paginate(page: params[:page], per_page: 25)
 
                if search_data[:search].present?
-                 posts.search(search_data[:search]).user_sort({ term: params[:sort], default: :search_score },
+                 posts.search(search_data[:search]).user_sort({ term: params[:sort] },
                                                               relevance: :search_score, score: :score, age: :created_at)
                else
                  posts.user_sort({ term: params[:sort], default: :score },
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -8,12 +8,11 @@ class ApplicationRecord < ActiveRecord::Base
 
   def self.match_search(term, **cols)
     sanitized = sanitize_for_search term, **cols
-    relevancyMultiplier = 0
-    mappedCols = sanitized.map{ |val| "#{val} AS search_score_#{sanitized.find_index(val)}" }.join(", ")
-    whereStatement = sanitized.map{ |val| "#{val}" }.join(" OR")
-    Rails.logger.debug(cols.values[0])
-    byebug
-    select(Arel.sql("`#{table_name}`.*, #{mappedCols}")).where(whereStatement).order('search_score_0*5').order(' search_score_1 ')
+
+    mappedCols = sanitized.map { |val| "#{val} AS search_score_#{sanitized.find_index(val)}" }.join(', ')
+    whereClause = sanitized.map { |val| val.to_s }.join(' OR')
+
+    select(Arel.sql("`#{table_name}`.*, #{mappedCols}")).where(whereClause).order('search_score_0 * 2 + search_score_1 * 1 DESC')
   end
 
   def self.sanitize_name(name)
@@ -31,15 +30,16 @@ class ApplicationRecord < ActiveRecord::Base
   def self.sanitize_for_search(term, **cols)
     cols = cols.map do |k, v|
       if v.is_a?(Array)
-        #need to sanitize here
-        v.map { |vv| " MATCH #{sanitize_name k}.#{sanitize_name vv} AGAINST ('#{term}' IN BOOLEAN MODE)" }
+        v.map do |vv|
+          #prob can do this just once at end of map
+          ActiveRecord::Base.sanitize_sql([" MATCH #{sanitize_name k}.#{sanitize_name vv} AGAINST (? IN BOOLEAN MODE)", term])
+        end
       else
         "#{sanitize_name k}.#{sanitize_name v}"
       end
     end
 
-    return cols[0]
-    # ActiveRecord::Base.send(:sanitize_sql_array, cols)
+    cols[0]
   end
 
   def self.sanitize_sql_in(ary)
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -8,7 +8,12 @@ class ApplicationRecord < ActiveRecord::Base
 
   def self.match_search(term, **cols)
     sanitized = sanitize_for_search term, **cols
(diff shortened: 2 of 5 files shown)


app/models/application_record.rb
```diff
@@ -26,13 +30,16 @@ def attributes_print
   def self.sanitize_for_search(term, **cols)
```
ArtOfCode-, line 30: The original behaviour of this method was to take a search parameter and list of columns, and turn them into a search against a _single_ fulltext index, i.e.

```ruby
sanitize_for_search 'query', posts: :body
# => "MATCH (`posts`.`body`) AGAINST ('query' IN BOOLEAN MODE)"
```

Do I understand right that what you're trying to do is make this return multiple `MATCH ... AGAINST` clauses for multiple fulltext indexes? At the moment, I don't think this does what you want it to. If you pass in an array of columns, you get the array of clauses:

```ruby
sanitize_for_search 'query', posts: [:body, :title]
# => [" MATCH `posts`.`body` AGAINST ('query' IN BOOLEAN MODE)", " MATCH `posts`.`title` AGAINST ('query' IN BOOLEAN MODE)"]
```

If you only pass a single column, you just get the column's name, no MATCH clause at all:

```ruby
sanitize_for_search 'query', posts: :body
# => "`posts`.`body`"
```

What's probably an easier way to do it (if I understand what you want correctly) is:

 * keep the purpose of the `cols = cols.map` block the same, i.e. map unsanitized column names to sanitized column names; then
 * return a map of that, which turns it into the MATCH clauses.

Something like this:

```ruby
cols = cols.map do |table_name, column_or_array|
  if column_or_array.is_a?(Array)
    column_or_array.map { |column_name| "#{sanitize_name table_name}.#{sanitize_name column_name}" }
  else
    "#{sanitize_name table_name}.#{sanitize_name column_or_array}"
  end
end.flatten

cols.map do |sanitized_name|
  ActiveRecord::Base.sanitize_sql(["MATCH (#{sanitized_name}) AGAINST (? IN BOOLEAN MODE)", term])
end
```

app/models/application_record.rb
```diff
@@ -8,7 +8,11 @@ def self.fuzzy_search(term, **cols)
 
   def self.match_search(term, **cols)
     sanitized = sanitize_for_search term, **cols
-    select(Arel.sql("`#{table_name}`.*, #{sanitized} AS search_score")).where(sanitized)
+
+    mappedCols = sanitized.map { |val| "#{val} AS search_score_#{sanitized.find_index(val)}" }.join(', ')
+    whereClause = sanitized.map { |val| val.to_s }.join(' OR')
```
ArtOfCode-, line 13: Style: Ruby variables are named in `snake_case`, i.e. `mapped_cols` and `where_clause`

app/models/application_record.rb
```diff
@@ -8,7 +8,11 @@ def self.fuzzy_search(term, **cols)
 
   def self.match_search(term, **cols)
     sanitized = sanitize_for_search term, **cols
-    select(Arel.sql("`#{table_name}`.*, #{sanitized} AS search_score")).where(sanitized)
+
+    mappedCols = sanitized.map { |val| "#{val} AS search_score_#{sanitized.find_index(val)}" }.join(', ')
```
ArtOfCode-, line 12: ActiveSupport has a neat extension you can use to simplify this:

```ruby
mapped_cols = sanitized.map.with_index { |val, idx| "#{val} AS search_score_#{idx}" }.join(', ')
```

app/models/application_record.rb
```diff
@@ -8,7 +8,11 @@ def self.fuzzy_search(term, **cols)
...
+    whereClause = sanitized.map { |val| val.to_s }.join(' OR')
+
+    select(Arel.sql("`#{table_name}`.*, #{mappedCols}")).where(whereClause).order('search_score_0 * 2 + search_score_1 * 1 DESC')
```
ArtOfCode-, line 15: Wrap raw SQL statements in `Arel.sql` when you pass them to ActiveRecord:

```ruby
.order(Arel.sql('search_score_0 * 2 + search_score_1 * 1 DESC'))
```

app/controllers/search_controller.rb
```diff
@@ -8,7 +8,7 @@ def search
                        .paginate(page: params[:page], per_page: 25)
 
                if search_data[:search].present?
-                 posts.search(search_data[:search]).user_sort({ term: params[:sort], default: :search_score },
+                 posts.search(search_data[:search]).user_sort({ term: params[:sort] },
```
ArtOfCode-, line 11: Could do with updating to re-add a default for the new weighted search, and to change the `relevance` sort to the same weighted search. Not sure whether our implementation of `user_sort` supports that via `Arel.sql`, though, so that's probably not one for this PR - needs investigation.

ArtOfCode- (changes requested):
Looks like a solid start! I'm not sure it's doing _exactly_ what you want it to right now - some comments below that will hopefully help.

app/models/application_record.rb
manassehkatz: Do we need "* 1" for search_score_1?

manassehkatz: What does "* 1" do? Seems unnecessary - i.e., is that any different from just "search_score_0 * 2 + search_score 1 DESC"?

ArtOfCode-: Probably not - I just copied it out without really thinking about it :)

BoxedFruits: Nope, does nothing really. Was just experimenting on how we would determine search relevancy

app/models/application_record.rb
BoxedFruits: > Do I understand right that what you're trying to do is make this return multiple MATCH ... AGAINST clauses for multiple fulltext indexes?
Yup exactly what I was trying to do. Ran your suggestion and it worked like a charm! 

Would there be a case where we pass in a single column? The methods are first being called here https://github.com/codidact/qpixel/blob/develop/app/models/post.rb#L59 and it looks like it will only ever look at the columns that we tell it. If there is, then the way I am trying to do the search weight would break (not that it was a good way to do it anyways 😛)

cellio:
It looks like we lost track of this.  Does anybody have any updates?  There is a more recent effort looking into moving to elasticsearch, which interacts with searching titles, but it's also marked as a draft.

PR #440, opened 2021-02-15
Allow OP to close own post

### LINKED ISSUE #397: Allow OP to close own question
https://meta.codidact.com/posts/280220

*If it is easy*, let's allow the OP to close a question regardless of abilities, same as you can edit or delete your own post without regard to abilities.  If someone suggests a duplicate and the OP agrees (the case described here) let's make that self-serve, but also, if the OP just realizes that a question is too broad or something but wants to work on it rather than deleting it, let's allow that.

If this is not easy, then let's hold off because we're eventually going to rework both closing and duplicate-marking anyway, so no sense investing heavily now.  I'm filing this because I think it's probably an "|| self" in the "can this person vote to close?" check.  Asking @luap42 for an opinion (not urgent).

BoxedFruits:
Resolves #397.

Checks if the post id is the same if the current user id

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -205,7 +205,7 @@ class PostsController < ApplicationController
   end
 
   def close
-    unless check_your_privilege('flag_close', nil, false) || @post.user.id === current_user.id
+    unless check_your_privilege('flag_close', nil, false) || @post.user.id == current_user.id
       render json: { status: 'failed', message: helpers.ability_err_msg(:flag_close, 'close this post') },
              status: :forbidden
       return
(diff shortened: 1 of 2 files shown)


app/controllers/posts_controller.rb
```diff
@@ -205,7 +205,7 @@ def update
   end
 
   def close
-    unless check_your_privilege('flag_close', nil, false)
+    unless check_your_privilege('flag_close', nil, false) || @post.user.id === current_user.id
```
ArtOfCode-, line 208: You're in Ruby, not JS - only two = needed here :)

ArtOfCode- (changes requested):

app/controllers/posts_controller.rb
BoxedFruits: Force of habit 😂. Will update tomorrow

ArtOfCode- (approved):

PR #445, opened 2021-03-01
Fix TODO links

sanzoghenzo:
Hi, I noticed the TODO list was moved to Trello, this will fix the links in the docs.


ArtOfCode-:
Thanks! That Trello link is actually out of date as well - we keep track of stuff in GH issues at the moment.

PR #457, opened 2021-03-20
Add privacy policy to help center

sau226:
This patch does the following:
* Adds a copy of the privacy policy converted to HTML format
* Makes minor formatting changes (mainly adding colons, please use a text diff tool to validate the changes if desired). These changes were made as part of the HTML format conversion.
* Bumps the last updated date for the policy (as a consequence of the conversion, we probably should do this)
* Switches the privacy policy link in the footer to point to the new help center post for privacy policies
* Deletes/removes the previous help post PDF.

If there are any further changes or reversal of changes that need to be made, please notify me and I can make those changes.

**Note:** Upon deployment of this change, a DB seed will need to be triggered for the privacy policy components to take effect.


db/seeds/posts/privacy-policy.html
```diff
@@ -0,0 +1,65 @@
+<h2 id="how-we-use-and-protect-your-information">How we use and protect your information</h2>
```
ArtOfCode-, line 1: Can we run a search-replace `&rsquo;` => `'` on this file? They don't need to be fancy.

ArtOfCode- (changes requested):

db/seeds/posts/privacy-policy.html
sau226: No problem - change has been made

PR #458, opened 2021-03-21
Switch from unpkg and cdnjs CDNs to jsdelivr CDN

sau226:
This patch switches all (where reasonably possible) JS and CSS packages in QPixel's header to be loaded from the JSDelivr CDN.

This change is made to ensure we are not loading static assets from multiple differing CDNs, which may marginally speed up site loading times. 

A tag has also been added to the HTML header to tell browsers to open a TCP connection before we try to load packages, however the speed up from this may be marginal, and it may be better deployed as an HTML tag.

As usual, if changes are required please request them and I will implement them where possible.


ArtOfCode-:
This is good to merge, just needs @sau226 to merge develop back into this branch to resolve conflicts.

sau226:
Merge conflict now resolved

PR #478, opened 2021-04-21
updating hover text of upvote and downvote

Istiakshovon:
Earlier it was only showing score point which was very confusing. And, new user would know by seeing the hover text what it does.

### DIFF
--- a/app/views/posts/_expanded.html.erb
+++ b/app/views/posts/_expanded.html.erb
@@ -26,11 +26,11 @@
   <div class="post--container <%= 'deleted-content' if post.deleted? %> grid is-nowrap">
     <% if post_type.has_votes %>
       <div>
-        <div class="post--votes has-text-align-center" title="<%= post.score %>">
+        <div class="post--votes has-text-align-center">
           <% existing_vote = my_vote(post) %>
           <% unless post.locked? %>
             <button class="vote-button button is-icon-only-button <%= (existing_vote&.vote_type == 1) ? 'is-active' : '' %>"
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
Thanks! Can we make these just say `"Score: <%= post.score %>"` instead? I don't think it needs any further explanation in the tooltip. (New lines also don't have great cross-browser support.)

cellio:
> Thanks! Can we make these just say `"Score: <%= post.score %>"` instead? I don't think it needs any further explanation in the tooltip. (New lines also don't have great cross-browser support.)

Can we add "(0-1)" or "(scale: 0-1)" after the score? The fractional values might confuse folks expecting "up minus down".

Istiakshovon:
> Thanks! Can we make these just say `"Score: <%= post.score %>"` instead? I don't think it needs any further explanation in the tooltip. (New lines also don't have great cross-browser support.)

`New lines also don't have great cross-browser support.` I didn't have the idea. That's why I wrote it. Cause, sometimes new user get issue to understand what does it do. When I was completely new to SO I was checking all of them cause, I didn't know what they do. Even, I forgot to read hover text. Than, `"Score: <%= post.score %>"` is ok.! :)

Istiakshovon:
> > Thanks! Can we make these just say `"Score: <%= post.score %>"` instead? I don't think it needs any further explanation in the tooltip. (New lines also don't have great cross-browser support.)
> 
> Can we add "(0-1)" or "(scale: 0-1)" after the score? The fractional values might confuse folks expecting "up minus down".

I think no! Cause, 'upvote' and 'downvote' score shows between those buttons. So, I don't think it is necessary.

cellio:
Oh right, the vote breakdown is right there -- so people should be able to interpret the number with the additional hint that it's a score.

Issue #479, opened 2021-04-22
Add revision numbers to history entries


cellio:
This suggestion is related to my answer here: https://meta.codidact.com/posts/281466#answer-281466.

It's sometimes useful to be able to refer to a specific revision of a post -- this answer applies to revision 5 (before the asker made a large change), or there's a meta question asking about an edit, or a flagger wants to point out something problematic.  Could we number the history entries when displaying them?

cellio:
Priority lowered because there *are* timestamps; they're just a little less friendly.

razetime:
Add a pull request for the fix in #484.

PR #484, opened 2021-04-29
Add history timestamp numbers

razetime:
Fix for Issue #479.

### DIFF
--- a/app/views/post_history/post.html.erb
+++ b/app/views/post_history/post.html.erb
@@ -4,10 +4,10 @@
   <%= render 'posts/type_agnostic', post: @post, show_category_tag: true, show_type_tag: true, last_activity: false %>
 </div>
 
-<% @history.each do |event| %>
+<% @history.each.with_index do |event,index| %>
   <details class="history-event">
     <summary>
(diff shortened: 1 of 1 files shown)


app/views/post_history/post.html.erb
```diff
@@ -4,10 +4,10 @@
   <%= render 'posts/type_agnostic', post: @post, show_category_tag: true, show_type_tag: true, last_activity: false %>
 </div>
 
-<% @history.each do |event| %>
+<% @history.each.with_index do |event,index| %>
   <details class="history-event">
     <summary>
-      <strong><%= event.post_history_type.name.humanize %></strong>
+      <strong>#<%= @history.size - index %>: <%= event.post_history_type.name.humanize %></strong>
```
ArtOfCode-, line 10: Just out of curiosity, why size - index instead of index +1?

razetime: The indices need to be reversed to be in chronological order, hence size - index.

ArtOfCode-: Ah, of course

PR #498, opened 2021-05-20
rule

Istiakshovon:


luap42:
Changes to the Code of Conduct, as a central community policy, should only be made after consultation with and agreement by the community; if you assume that these changes are necessary, suggest them on <meta.codidact.com>.

That aside, I don't think this change is necessary. We have "Be nice; be respectful" as general rule and a ban of "Rudeness" more specifically already. Providing an exhaustive list of things that are good or bad is not the intention of our Code of Conduct.

PR #508, opened 2021-06-08
Add keyboard shortcuts for markdown text editor

### LINKED ISSUE #274: Keyboard shortcuts within the post editor
https://judaism.codidact.com/questions/276735
https://meta.codidact.com/questions/277179

Ctrl-B (bold), ctrl-I (italic), etc.  @superplane39 proposed the following in an answer on the first link:

> Ctrl-I - italics
Ctrl-B - bold
Ctrl-K - link
Ctrl-P - preformatted (assuming that's possible without loading the print dialog)
Ctrl-Q - blockquote
Ctrl-U - upload an image

rahulsnkr:
Resolves #274 - Add the following keyboard shortcuts for the markdown text editor as mentioned in the issue

> Ctrl-I - italics
> Ctrl-B - bold
> Ctrl-K - link
> Ctrl-P - preformatted
> Ctrl-Q - blockquote
> Ctrl-U - upload an image

### DIFF
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -49,6 +49,39 @@ $(() => {
     }
   });
 
+  $('.js-post-field').on('keydown', ev => {
+    if (ev.ctrlKey) {
+      switch (ev.keyCode) {
+        case 66:
+          $('[data-action="bold"]').click();
+          break;
+        
+        case 73:
+          $('[data-action="italic"]').click();
+          break;
+
+        case 75:
+          ev.preventDefault();
+          $('[data-modal="#markdown-link-insert"]').click();
+          break;
+
+        case 80:
+          ev.preventDefault();
+          $('[data-action="code"]').click();
+          break;
+
+        case 81:
+          $('[data-action="quote"]').click();
+          break;
+
+        case 85:
+          ev.preventDefault();
+          $('[data-modal="#markdown-image-upload"]').click();
+          break;
+      }
+    }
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/markdown.js
```diff
@@ -49,6 +49,39 @@ $(() => {
...
+        case 66:
+          $('[data-action="bold"]').click();
+          break;
+        
+        case 73:
+          $('[data-action="italic"]').click();
+          break;
+
+        case 75:
+          ev.preventDefault();
```
ArtOfCode-, line 64: Why are only some of these using `ev.preventDefault()`? Do the others work without it?

rahulsnkr: yep that's right, specifically `ev.preventDefault()` is for `Ctrl+P` (opens the print dialog), `Ctrl+K` (focuses on the browser search bar), and `Ctrl+U` (opens the page source). The rest work without it.

cellio: For the other cases, is preventDefault() a no-op or does it not work (because there is no default)?  If the former, I'm wondering if we should do it for all of them in case browsers add more overlapping key bindings in the future.  Disclosure: this is not my area, so if this is a dumb question please let me know.

ArtOfCode-: Ah, of course. I'm thinking of the default being "insert letter", but of course it's not if it's Ctrl. Thanks.

PR #516, opened 2021-06-13
show question on top of editing answer

Istiakshovon:
I think that if "Responding to" or, question shows on top of answer (edit Answer) than it might look good. And, when the feature came I didn't notice that also.

I may set the code wrong way also. Cause, I didn't check main file properly. And, I am completely new to Ruby.

### DIFF
--- a/app/views/posts/_form.html.erb
+++ b/app/views/posts/_form.html.erb
@@ -54,6 +54,14 @@
     </p>
   <% end %>
 
+
+
+  <% if inline_parent && defined?(parent) && parent.present? %>
+    <h2><%= t('posts.responding_to')%></h2>
+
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
If we're going to move this to the top, we should do it everywhere, including on edits.

My personal feeling is that it's better to have this underneath - when you go to write or edit an answer, you expect the textbox and tools to write the answer at the top of the page, then we have the question under that for reference. @cellio thoughts?

cellio:
What does this do, move the question to the top of the page with the editor for the answer below?  What guidance appears where?

ArtOfCode-:
Exactly that, @cellio

PR #527, opened 2021-06-30
broken link

Istiakshovon:
This seems to be broken. According to [him](https://discord.com/channels/634104110131445811/634104110131445815/859770585704693780), it happened for "!". But, I am not sure if it happened for exclamation mark. Otherwise, I would add direct link or, remove the text or, only "#".

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,4 +187,4 @@ See the README.md in the [docker](docker) folder for complete instructions.
 
 <br>
 
-![⬆ Back to Top](#table-of-contents)
+[⬆ Back to Top](#table-of-contents)
--- a/README.md
+++ b/README.md
@@ -56,12 +56,10 @@ For Arch-Based Linux:
 sudo pacman -Syyu
 sudo pacman -Sy gcc
 sudo pacman -Sy make
-sudo pacman -Sy ruby autoconf bison base-devel ruby unixodbc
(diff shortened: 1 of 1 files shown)


sau226:
> But, I am not sure if it happened for exclamation mark. 

This is fixed by the exclamation mark removal. If you test out the button in your PR branch, it should work as originally intended to function. I wrote up the PR that introduced the broken control and must have added it incorrectly somehow.

> Otherwise, I would add direct link or, 

If you use the network inspector or similar feature of your browser, you will notice that no extra internet requests are required to use the back to top feature if properly implemented. If we add a direct link, this will make an extra internet request, which results the same file being loaded/requested twice (which has a small, but in some cases significant impact on internet bandwidth used, due to the images present in it, if cache is not effective in a given scenario).

> remove the text or, 

This could be done, but given the current/potential future size of the readme and presence of a table of contents that enables easy navigation inside it, this is a feature that is best left in for user convenience.

> only "#".

The reference uses text to allow jumping directly back to the table of contents. Using only "#" would scroll the user to the top of the readme, forcing them to scroll back down and (although in a very minor and almost indiscernible way) impacting the ease of use. If the top header section of the readme gets any bigger (e.g. more badges need to be added to it), then this becomes more exacerbated.

PR #528, opened 2021-06-30
Fix misplaced exclamation mark

sau226:
This fixes a misplaced exclamation mark that has blocked the use of a readme "back to top control"


sau226:
This is a duplicate PR, submitted already. Closing.

Issue #466, opened 2021-04-02
Can we show the usercard stats, including rep, on the user profile?


cellio:
Usercards contain several (configurable) stats, including reputation.  Can we find a way to show that same information on the user profile, too?  We show some info in the right column, like number of posts, but not other info like number of edits.  Or rep, which people ask about.

One way to do it would be to add a row of stats under the avatar and above the abilities block.  It could be the same presentation as on the usercard -- icons, numbers, tooltips.  (But I would put rep on the same line, not above it, since there's room and I think that would look tidier.)  Ideally we'd show the same stats that are configured on that community for usercards, though I think it's also fine to show all of them.

It seems odd to have to leave the profile page and go to a post in order to see these stats.

kevinmmansour:
So you want to create for example like:

![image](https://user-images.githubusercontent.com/71846550/124404676-3f9d9400-dd3c-11eb-926c-e9c1bacbf0db.png)

cellio:
That's good as a minimum -- rep is nowhere findable now and that's the biggest omission.  Can we also include the other stats from the usercard?

![screenshot](https://user-images.githubusercontent.com/5557942/124405404-896e8600-dd0c-11eb-9fc0-43f637c90787.png)

The "spec" for the stats in the usercard is here: https://meta.codidact.com/posts/280005.  I'm wondering if we can call the same code that renders that in usercards to also show those stats on the user profile below the "reputation" line.  I do like spelling out "reputation" on the profile rather than just the "rep" from the user card, though.  So I'm guessing you'd call whatever code generates that usercard string with everything except the leading "r." (rep + newline), and then separately get and show the rep like you've done here.  But I haven't seen that code so I'm speculating about how it works.

kevinmmansour:
So far.

This is what I made:

![image](https://user-images.githubusercontent.com/71846550/124633753-e6417c00-de85-11eb-8c19-1d8f2902d7ae.png)

Notice 1: Beside reputation there is no icon since there is no icon for it the Card bt we might choose another icon.

Notice 2: This from my local dev environment/

cellio:
Looks good!  In the absence of an icon, can you align the text vertically (move it over as if there were an icon)?

kevinmmansour:
Sorry for wasting your time. But here is latest preview:
![image](https://user-images.githubusercontent.com/71846550/124658080-9ffb1580-dea3-11eb-85f9-b9a60db547c2.png)

cellio:
[...]

PR #533, opened 2021-07-05
Renaming "rep" to "reputation" in Post User Card, Show the usercard stats on the user profile.

kevinmmansour:
Renaming "rep" to "reputation" in Post User Card. 

image

Show the usercard stats on the user profile. 

image

It will solve #466.

### DIFF
--- a/app/views/users/_common_card.html.erb
+++ b/app/views/users/_common_card.html.erb
@@ -35,7 +35,7 @@
           <span class="user-card--break"></span>
         <% when 'r' %>
           <span class="user-card--detail">
-            <%= user.reputation %> <span title="reputation">rep.</span>
+            <%= user.reputation %> <span title="reputation">reputation.</span>
           </span>
         <% when 'p' %>
           <span class="user-card--detail">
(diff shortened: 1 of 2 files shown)


Istiakshovon:
I don't think that it's a good idea. Reputation text is bigger than usual. I think earlier text was good.

cellio:
On the other hand, it doesn't cost anything to be clear and hover text doesn't work as well for touch interfaces as for desktop -- so yes we'll use hover text sometimes, knowing that limitation, but do we need to here?  "Rep" is insider lingo; does it need to be?

kevinmmansour:
> 
> 
> I don't think that it's a good idea. Reputation text is bigger than usual. I think earlier text was good. The hover text is showing `reputation`. So it's (PR) not very helpful ~ I guess.

I think "reputation" is more clear than "rep". Although as cellio mentioned: The hover text is not showing in touch interfaces devices (e.g Mobile Devices, Or Laptops with touch screen), so, "reputation" is more helpful & clear.

PR #537, opened 2021-07-06
Show question above of thread comment

Istiakshovon:
image

I will make further changes to the page.. Like question collapsing and expanding


cellio:
Reminder: the comment thread could be attached to an answer rather than a question.

I recommend keeping the link back to the full page in any case, so it's easy to go back and look at other comment threads (or answers).

Issue #541, opened 2021-07-07
Starting a thread with a comment that's too short returns 422


cellio:
https://meta.codidact.com/posts/282480

I've reproduced this in production.  On that question I started a new thread, entered "too short" as the comment text, didn't enter a title, and clicked the post button.  I got a 422.  I expected to get "Content is too short (minimum is 15 characters)", like I do when adding a comment to an existing thread.

The 422 is kind of baffling; it suggests logging out and back in, which isn't going to help in this case.  It seems like there's one error-checking call we've missed specifically in the "create new thread" case.

kevinmmansour:
Done! with at least of 15 characters.

PR #552, opened 2021-07-11
Starting a thread with a comment at least of 15 Chars

kevinmmansour:
Starting a thread with a comment at least of 15 Chars.

Error Example:

Screenshot

Built using built in attribute in Text Area element (`minlength`).

Solves #541

### DIFF
--- a/app/views/comments/_new_thread_modal.html.erb
+++ b/app/views/comments/_new_thread_modal.html.erb
@@ -17,7 +17,7 @@
         <%= label_tag :body, 'Your comment', class: 'form-element' %>
         <div class="form-caption">Start the thread with a comment.</div>
         <%= text_area_tag :body, '', class: 'form-element js-comment-field', data: { post: post.id, thread: '-1' },
-                          required: true %>
+                          required: true, minlength:"15" %>
         <span class="has-float-right has-font-size-caption js-character-count"
               data-target=".js-comment-field" data-max="500">0 / 500</span>
 


FoggyFinder:
Is this behaviour consistent with one for an existing comment thread ?

kevinmmansour:
Sorry, but I didn't understand. The problem is when you start a new comment thread you can enter any number of characters under 500 but if you enter less than 15 it just produce 422 so I solved it by force user to enter at least 15 characters.

FoggyFinder:
If you try to submit a short comment to an existing thread you will get the response like that

image

which is a completely different than one that you've suggested

I'd personally like to see something consistent

kevinmmansour:
> 
> 
> If you try to submit a short comment to an existing thread you will get the response like that
> 
> image
> 
> which is a completely different than one that you've suggested
> 
> I'd personally like to see something consistent

Sure, I understood. so, I will wait until I get review from community as which one to use.

cellio:
Thanks for working on this.

We should be consistent.  Is the current behavior entirely client-side, or does that pink error pop-up come from the server rejecting a comment?  If we can catch it client-side we should, and if we're doing it all client-side then we should be able to do a little better than we are now.  Specifically:

- Since you now have to click the button to leave a comment ("return" doesn't post it), should we make the button inactive until the conditions are met?  On hover we could then show "Comment must be at least 15 characters".  (On a touch interface maybe show that text beside the button?)
- In the character count ("13/500"), can we make the first number red until you reach the minimum, like how we use red if you go over 500?
- These considerations should apply for both creating a thread (initial comment) and adding a comment to an existing thread.  Creating a thread is, logically, "create thread name" + "add comment"; let's try to make sure "add comment" behaves the same everywhere.

If all of that is too much scope creep for this effort, then using the current pink error on clicking the "create thread" button would be fine.  Let's either be consistent with what we have or make it better in both contexts so we don't confuse people.  Thanks.

Istiakshovon:
@kevinmmansour The submit button disable when character is over than, 500. So, you can try that if character is less than 15 then, disable button.

PR #553, opened 2021-07-11
Reactions

luap42:
Implementing my proposal from https://meta.codidact.com/posts/277957, this pull request adds *reactions* to QPixel. Reactions are in many ways similar to votes, but they provide a more specific meaning than that a post is good or is bad. The set of available reactions can be edited on a per-site basis, but this PR includes the following as an -- in my opinion -- reasonable default:

* Works for Me
* Outdated
* Dangerous

These reactions are examples of *good* reaction types, because they provide meaningful information to the reader (this post has been confirmed to be working, this post might be outdated and therefore not the proper solution for your specific problem and this post's proposal is dangerous and should be adopted with care and only if you have sufficient knowledge). They are not just an other way to up- or downvote a post. For example, an outdated post might still be worth an upvote if the proposal in it is good, just for a deprecated version of a tool or framework.

Reactions are shown right above a post, clicking on them opens a modal dialog with more information, including who reacted and if they wrote a comment an excerpt of and a link to of that too.

a "Works for me" reaction by two users above a post

the details modal for the previous reaction showing that the second user added a comment

While reactions lower the bar to give feedback and while they make feedback of some types more prominent, they are also not supposed to be a replacement for comments. Therefore they are closely integrated into comment threads. The reaction creation dialog allows you to additionally post a comment; some reactions might also require a comment to be posted (activated by default for the Dangerous reaction).

Reactions can be added by clicking a button below the voting buttons; this will open an drop panel allowing you to select reactions, adding (optionally) a comment and then submitting them. If you already have a reaction of one type, you can retract it from there:

image of the panel described above

Moderators of any site can edit the list of reaction types available on this site. However deleting reactions will be reserved for administrators as it will cascade to all added reactions by users and therefore being an irreversible, destructive action:

moderator dashboard

-----

Basic functionality has been completed, what's making this PR work in progress is:

* editing, adding and deleting reaction types
* testing


cellio:
Looking good so far!  (Based on the description and screenshots here, not code inspection. :-) )  Some questions/thoughts:

- I like that the "react" link is downplayed a little -- below the (larger) voting buttons.  Should leaving a reaction prompt the user to cast a vote in some cases?  You said this is dangerous -- don't forget you can downvote, or you said this works, don't forget you can upvote.  I don't think we want to *require* a vote in order to react (for example, out of date isn't necessarily bad, just out of date), but people have raised concerns about reactions impeding votes, so it's something we should watch for.

- Showing the timestamps on the reactions is very helpful, thanks.  What else can we do to provide a signal when a post was edited after a reaction was made?  What do you think about indicating the revision number a reaction applied to (with history link), with "current" where there hasn't been an edit?

MoshiKoi:
Are reaction types tied to post types? A community might want to have different reactions on different post types, e.g. Code Golf might want to have a set for Q&A posts and a separate set for Challenge Answer posts.

Istiakshovon:
<%if current_user.nil?{
 <button class="react-button button is-muted is-icon-only-button h-fw-bold"
                    id="post-<%= post.id %>-react" aria-label="Add Reaction"
                    data-drop="#post-<%= post.id %>-reactions-panel"
                    data-drop-self-class-toggle="is-active"
                    data-drop-force-dir="down">
                react
            </button>
}%>

ArtOfCode- (dismissed):

PR #561, opened 2021-07-14
Fixing typos and grammer mistakes

kevinmmansour:
Fixing typos and grammer mistakes across all the platform.


app/views/flags/history.html.erb
```diff
@@ -1,7 +1,7 @@
 <h1>Flag History</h1>
 <p>
   Below is a list of all the flags you have cast and the outcome of each of them. Moderators handling your flags may also
-  have left you feedback.
+  have left you a feedback.
```
cellio, line 4: "Feedback" is a mass noun not a counting noun.  "A feedback" sounds odd to me.  "Feedback" or "a response" both work.

cellio:
Aside from one place where I commented these wording changes look fine (and are clearer).  This isn't Python so I *assume* the couple of whitespace changes don't matter.

app/views/flags/history.html.erb
kevinmmansour: OK. rolled back to "Feedback".

kevinmmansour:
> 
> 
> Aside from one place where I commented these wording changes look fine (and are clearer). This isn't Python so I _assume_ the couple of whitespace changes don't matter.

Yes, changes to whitespace don't differ but sometimes they are kind of formatting but I don't see that I changed in whitespace in this pull request.

cellio:
@ArtOfCode-  this looks fine to me.

Issue #542, opened 2021-07-07
Comment notifications: include post title along with thread title


cellio:
https://meta.codidact.com/posts/282436#answer-282436

A notification about a new comment in "General Comments" isn't very informative.  Before threading the notification included the title of the post, and for a new answer the notification includes the title.  I realize we're including the thread title, but without context that's not always very informative.  Can we *also* include the post title?  I realize that long titles + long thread titles could make for very long notifications and maybe we need to truncate one or both in that case, but can we find a way to get the context of the post title into the notification?

cellio:
Fixed:

![Screen Shot](https://user-images.githubusercontent.com/5557942/127692898-2470f760-51fb-4944-b707-5669f17d32ec.png)

Issue #536, opened 2021-07-06
If a tag wiki is long don't show all of it by default on the tag page


cellio:
https://meta.codidact.com/posts/282086

If a tag wiki is long (and there are good reasons for them to be long), you have to skip past the whole description to get to the list of questions.  Sometimes you want the description, but often you just want the list of questions.  Can we show only the first paragraph or so of a long description and put the rest behind an in-page expander like the HTML `details` element?  We don't want people to have to go to a separate page, but maybe the first, say, 500 characters plus a thing to click would be good enough?

We should always show the full *summary*; this is about the longer description.

Here's an example case: https://software.codidact.com/categories/38/tags/3296

Istiakshovon:
I have seen [a tag](https://stackoverflow.com/tags/python/info) (in SO) which wiki is too large. I want to add some context also. 

When you look at our tag you will notice that there's no context like `Tag Info` or something just like this. I guess, we should add something like this
![Uploading Screenshot from 2021-07-17 11-57-39.png…]()

cellio:
The sections aren't labeled, if that's what you mean -- there's a box with the summary and then another box with the longer description, and after that the list of posts (with two options, tag-only and tag+children -- we have hierarchical tags, unlike Somewhere Else).

My suggestion would be to always show the summary (the first part) and, if a longer description exists, show it behind an expander (same idea as the "Post" link at the top of a comment-thread page).  If there is no longer description, then keep the current behavior, which is to show text saying there isn't one and inviting you to create one.  (Example: https://software.codidact.com/categories/38/tags/4015)

Istiakshovon:
![Screenshot from 2021-07-22 12-01-11](https://user-images.githubusercontent.com/58106197/126670856-8bde72a2-f77f-4936-b286-078de93bf0ad.png)

Is it OK? If length of tag wiki is over than, 500 then it will show like this. What do you think of the design?

PR #563, opened 2021-07-15
Pull Request

Istiakshovon:
>Fools write code which only computer can understand. An intelligent programmer write code which human can understand

I don't understand my own code. So, how can a human understand my code 😆


Istiakshovon:
<h2>What new changes will do :</h2>

- disable submit button unless 15 character is added. 
- Character count span color will be red until 15 character is entered.

cellio:
Those sound like good changes, thanks!  As for the thread title, we figured the beginning of the first comment would be clearer than what we might get if we require people to manually enter something -- that's how you get names like "wfuepri25".

Istiakshovon:
https://github.com/codidact/qpixel/issues/542#issue-938409723

app/assets/javascripts/comments.js
```diff
@@ -200,6 +200,19 @@ $(() => {
     const splat = content.split(' ');
     const caretPos = $tgt[0].selectionStart;
     const [currentWord, posInWord] = currentCaretSequence(splat, caretPos);
+    const postId = $tgt.data('post');
+    const characterCount = document.getElementById("character_count_"+postId);
+    const createThreadButton = document.getElementById("create_thread_button_"+postId);
+
+    if (caretPos<15 || caretPos>500){
```
ArtOfCode-, line 207: Spaces around operators, parens, and braces:

```suggestion
    if (caretPos < 15 || caretPos > 500) {
```

app/assets/javascripts/comments.js
```diff
@@ -200,6 +200,19 @@ $(() => {
...
+        characterCount.style.color = "#ff0000";
+        createThreadButton.disabled = true;
+    }else if(caretPos>375){
```
ArtOfCode-, line 210: If-else statements should start new lines:

```suggestion
    }
    else if (caretPos > 375) {
```

app/assets/javascripts/comments.js
```diff
@@ -200,6 +200,19 @@ $(() => {
...
+        characterCount.style.color = "#d7af02";
+    }else{
```
ArtOfCode-, line 212: Likewise here

app/assets/javascripts/comments.js
```diff
@@ -240,3 +253,14 @@ $(() => {
     }
   });
 });
+
+
+window.new_thread = function (post_id){
```
ArtOfCode-, line 258: Avoid adding things to the `window` scope if at all possible. Look at where this function is used - is it only used in this file? If so, then you can just define it for this file. If you need it across multiple files, it'd be better to add this as an API method to `qpixel.js`, where it can come under the `window.QPixel` scope.

Function and parameter names should use `camelCase`, and ES6 function definitions where possible:

```suggestion
const newThread = (postId) => {
```

app/assets/javascripts/comments.js
```diff
@@ -240,3 +253,14 @@ $(() => {
...
+    const thread = document.getElementById("new_thread_modal_"+post_id);
+
+    if (thread.style.display=="none"){
+        thread.style.display = "block";
+    }else{
+        thread.style.display = "none";
+    }
```
ArtOfCode-, line 265: Same comments about spacing around operators and new lines for if-else.

app/views/comments/_new_thread_modal.html.erb
```diff
@@ -18,15 +16,14 @@
...
-        <span class="has-float-right has-font-size-caption js-character-count"
+        <span class="has-float-right has-font-size-caption js-character-count" id="character_count_<%= post.id %>"
               data-target=".js-comment-field" data-max="500">0 / 500</span>
 
         <%= label_tag :title, 'Comment thread title (optional)', class: 'form-element' %>
         <div class="form-caption">You can give your comment thread a title. If you leave this blank, the beginning of the initial comment will be shown.</div>
         <%= text_field_tag :title, '', class: 'form-element' %>
 
-        <%= submit_tag 'Create thread', class: 'button is-filled' %>
+        <%= submit_tag 'Create thread', class: 'button is-filled', id: "create_thread_button_#{post.id}", disabled:true %>
```
ArtOfCode-, line 26: Space around arguments:

```suggestion
        <%= submit_tag 'Create thread', class: 'button is-filled', id: "create_thread_button_#{post.id}", disabled: true %>
```

app/views/posts/_expanded.html.erb
```diff
@@ -447,8 +448,10 @@
                 <% elsif post.comments_disabled %>
                   <p class="has-color-red-500"><i class="fa fa-lock"></i> Comments have been disabled on this post, but as a moderator you are exempt from that block.</p>
                 <% end %>
-                <a href="#" class="button is-muted is-small" data-modal=".js--<%= post.id %>-new-thread-modal">Start new comment thread</a>
+<!--                <a href="#" class="button is-muted is-small" data-modal=".js--<%#= post.id %>-new-thread-modal">Start new comment thread</a>-->
```
ArtOfCode-, line 451: If you don't need the previous definition any more, you can remove this comment.

app/views/posts/_expanded.html.erb
```diff
@@ -447,8 +448,10 @@
...
+                <a class="button is-muted is-small" onclick="new_thread(<%= post.id %>)">Start new comment thread</a>
```
ArtOfCode-, line 452: Related to my other comment about not adding to the `window` scope: instead of using an `onclick` attribute on this element, add a class to it such as `js-new-thread-link`, and add a click handler in the JS file, such as with `$(document).on('click', '.js-new-thread-link', ev => { });`

Remember that you can add `data-*` attributes to this element if you need parameters. Something like this would work:

```suggestion
                <a class="button is-muted is-small js-new-thread-link" data-post="<%= post.id %>">Start new comment thread</a>
```

This is a common pattern in our JS, so you can find examples if you need it. Your JS will probably look something like this:

```js
$(document).on('click', '.js-new-thread-link', ev => {
  const $tgt = $(ev.target);
  const postId = $tgt.attr('data-post');
  // carry on from here as your existing function does
});
```

ArtOfCode- (changes requested):

Istiakshovon:
https://github.com/codidact/qpixel/issues/536

app/assets/javascripts/comments.js
ArtOfCode-: This applies to all instances of this formatting. Try to follow the style of the existing JS in the project - new lines before every new if/else if/else block.

app/assets/javascripts/comments.js
```diff
@@ -255,10 +256,10 @@ $(() => {
 });
 
 
-window.new_thread = function (post_id){
+const newThread = (postId) => {
     const thread = document.getElementById("new_thread_modal_"+post_id);
 
-    if (thread.style.display=="none"){
+    if (thread.style.display == "none"){
```
ArtOfCode-, line 262: Always use `===` unless you're comparing to `null` or `undefined`

app/views/tags/show.html.erb
```diff
@@ -42,7 +42,16 @@
     <% end %>
   </div>
   <div class="widget--body">
-    <%= raw(sanitize(@tag.wiki, scrubber: scrubber)) %>
+    <% if@tag.wiki.present? %>
```
ArtOfCode-, line 45: You're missing a space here - `<% if @tag.wiki.present? %>`

app/views/tags/show.html.erb
```diff
@@ -42,7 +42,16 @@
...
+      <% if @tag.wiki.length < 600 %>
+      <%= raw(sanitize(@tag.wiki, scrubber: scrubber)) %>
```
ArtOfCode-, line 47: Content of if/else blocks should be indented one more level beyond the start of the block. This line and the HTML in the else block here need one more indent level.

app/views/posts/_expanded.html.erb
```diff
@@ -448,8 +448,7 @@
                 <% elsif post.comments_disabled %>
                   <p class="has-color-red-500"><i class="fa fa-lock"></i> Comments have been disabled on this post, but as a moderator you are exempt from that block.</p>
                 <% end %>
-<!--                <a href="#" class="button is-muted is-small" data-modal=".js--<%#= post.id %>-new-thread-modal">Start new comment thread</a>-->
-                <a class="button is-muted is-small" onclick="new_thread(<%= post.id %>)">Start new comment thread</a>
+                <a class="button is-muted is-small js-new-thread-link" data-post="<%= post.id %>">Start new comment thread</a>
```
ArtOfCode-, line 451: You've changed this to a class and removed the `onclick` attribute, which is great, but I don't see anywhere you've linked it up in the JS? Refer to my original review comment for an example - you need a click handler in the JS.

ArtOfCode- (changes requested):

app/assets/javascripts/comments.js
Istiakshovon: Actually, the problem is I never formatted code. :( Sorry!

app/views/posts/_expanded.html.erb
Istiakshovon: I guess, you had changed it earlier. Cause, your suggested text is showing to me. Maybe, you have requested the changes from first/second one. 
Screenshot from 2021-07-25 00-36-34

(https://github.com/Istiakshovon/qpixel/edit/develop/app/views/posts/_expanded.html.erb?pr=%2Fcodidact%2Fqpixel%2Fpull%2F563)

Istiakshovon: I have edited it's working now.. (tested)

app/views/posts/_expanded.html.erb
```diff
@@ -460,5 +462,3 @@
 <% if moderator? || current_user&.has_post_privilege?('flag_curate', post) %>
   <%= render 'posts/post_tools', post: post %>
 <% end %>
-
-<%= render 'comments/new_thread_modal', post: post %>
```
ArtOfCode-, line 464: Where is the new thread modal being added into the page, if not here? I can't seem to see anywhere else.

Istiakshovon: Check in latest expanded.html.erb commit. The new thread modal is on 453 line.
```
<% else %>
                <% if post.locked? %>
                  <p class="has-color-red-500"><i class="fa fa-lock"></i> Comments are disabled on locked posts, but as a moderator you are exempt from that block.</p>
                <% elsif post.comments_disabled %>
                  <p class="has-color-red-500"><i class="fa fa-lock"></i> Comments have been disabled on this post, but as a moderator you are exempt from that block.</p>
                <% end %>
                <a class="button is-muted is-small js-new-thread-link new_thread" data-post="<%= post.id %>">Start new comment thread</a>
              <% end %>
                <%= render 'comments/new_thread_modal', post: post %>
            <% end %>
```

PR #567, opened 2021-07-18
Hide impersonate user link for non developers

sau226:
Please don't expose the impersonate link to non devs, as they will simply get a 404 error if they click it

This PR gates the impersonate user link behind a developer check, and has not been validated on any test system by its author. The route itself is behind a dev check already - this just hides the otherwise broken for non dev UI element.


cellio:
I agree with the goal.  I haven't tested it either.

PR #572, opened 2021-07-27
Update show.html.erb

Istiakshovon:
I have tested [it](https://meta.codidact.com/posts/282885) in most of resolution. For me, it was working in all resolution.


ArtOfCode-:
Thanks for the attempt at this one, but I'd prefer to fix this a different way - center rather than just make it bigger.

PR #574, opened 2021-07-28
Update comments_controller.rb

Istiakshovon:
Earlier there was formatting problem. I was thinking how you wanted to set it. But, couldn't understand that's why I am removing it.

I think you wanted output something just like this : `"There are new comments in a followed thread title "  "on the post post_title"`. But, I can't see why we need quotation mark that way. Your formatting was little bit buggy. e.g. `There are new comments in a followed thread 'Funny that...' on the post ''` I wonder Why " quotation mark had became ' ' (in last example)

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -80,8 +80,7 @@ class CommentsController < ApplicationController
         existing_notification = follower.user.notifications.where(is_read: false)
                                         .where('link LIKE ?', "#{helpers.comment_link(@comment)}%")
         unless existing_notification.exists?
-          follower.user.create_notification("There are new comments in a followed thread '#{@comment_thread.title}' " \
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
This was deliberate. These two lines, as they were, were too long, which was causing rubocop to fail. The backslash `\` between lines is just a continuation - whatever's on the second line continues directly on from the first line. Whether they're written as one line or two, these have the same meaning, but in our case it has to be two to satisfy the linter.

Istiakshovon:
> This was deliberate. These two lines, as they were, were too long, which was causing rubocop to fail. The backslash `\` between lines is just a continuation - whatever's on the second line continues directly on from the first line. Whether they're written as one line or two, these have the same meaning, but in our case it has to be two to satisfy the linter.

But, actually I noticed I got a notification where post title wasn't showing. Could you reproduce?

Issue #581, opened 2021-07-30
Commenting on a existing thread can't return post (question's) title


Istiakshovon:
The bug was mentioned [here](https://github.com/codidact/qpixel/pull/574#issuecomment-888779606). When commenting on a existing thread in answer. That makes a notification which looks like,"There are new comments in a followed thread 'Title' on the post \"" 

![image](https://user-images.githubusercontent.com/58106197/127438515-efeec738-b81a-4e26-a88b-4030ae3b9297.png)

PR #583, opened 2021-07-30
Update comments_controller.rb

Istiakshovon:
https://github.com/codidact/qpixel/issues/581 

Will `@comment.root.title` better than if-else? (I have tested using `@comment.root.title` but, it wasn't working for me.)


ArtOfCode-:
Instead of the if-else block, I'd just use a variable for this.

```ruby
title = @post.parent.nil? ? @post.title : @post.parent.title
follower.user.create_notification("There are new comments in a followed thread '#{@comment_thread.title}' " \
                                  "on the post '#{title}'",
                                  helpers.comment_link(@comment))
```

PR #588, opened 2021-08-01
Pull Request

Istiakshovon:


ArtOfCode-:
Just having the tag wiki there with no introduction seems fine to me. What makes you find it difficult to understand?

Istiakshovon:
> Just having the tag wiki there with no introduction seems fine to me. What makes you find it difficult to understand?

How someone (saying for a new user) can understand what the title actually is.

Istiakshovon:
Copy link is more helpful than link.

ArtOfCode-:
Whether it makes sense is down to how the tag wiki is written. Happy to accept the comment permalink change, but I don't think we need the tag wiki title change. (In general, it's better to split unrelated changes into multiple pull requests to avoid only being able to accept _half_ the pull request. Also helps if you can put something in the commit message other than the default "updated <file>" so that we know what changes are in it.)

PR #594, opened 2021-08-03
Update _comment.html.erb

Istiakshovon:
Usually, earlier it wasn't changing to Copy in Post page. So, I think Copy Link would be better by default

Screenshot from 2021-08-03 04-48-57
Screenshot from 2021-08-03 04-49-09

Earlier, the UI was little bit messy for Android.


ArtOfCode-:
Sorry, I did this one manually before I saw this PR.

Istiakshovon:
@ArtOfCode- No problem! It's OK :D

Issue #565, opened 2021-07-17
dark theme for Codidact


Istiakshovon:
I am tired of using extension. When I use extension (dark theme) for Codidact and, move to a dark theme website then that dark website becomes lighter. Which I totally hate. I always have to change theme for changing tabs. So, I am thinking to build dark theme for Codidact. In SE sites, only SO has dark theme (maybe that's in beta mode). 

Reason of building dark theme : 

- Not to get distracted while writing big answers/questions
- Not to affect eyes.

cellio:
We've gotten other requests for a dark theme, so if somebody wants to work on one that sounds like a benefit.  I don't know a lot about front-end stuff; is this something that should be done at the CoDesign layer?

Istiakshovon:
> We've gotten other requests for a dark theme, so if somebody wants to work on one that sounds like a benefit. I don't know a lot about front-end stuff; is this something that should be done at the CoDesign layer?

I am sure of CoDesign. Actually, I have to design scss file two way (I could let them in a single page). If `background.light{}` and `background.dark{}`. If someone chooses dark theme than, `.light` won't show to him. That's the simple way to do that. But, there's huge scss file that's why it will take longer to be done. I will work on it with patience. :)

cellio:
Please talk with @luap42 about the best way to make this work with CoDesign.  There are some color things already in there; a dark mode would need to play well with them (or they'd need to be adjusted), and there might already be hooks in there for the things you want to do.

ArtOfCode-:
Yes, this is something that should be done from Co-Design. There'll be some things that need to be done in QPixel CSS as well, but most of it should come from CD. Remember there's a media query for this: `@media screen and (prefers-color-scheme: dark) { }`

cellio:
CoDesign repo: https://github.com/codidact/co-design

razetime:
@ArtOfCode- would be worth transferring over to co-design, then?

Istiakshovon:
@razetime some design(header widget and some more) had been in co-design. Those designs don't let me do any changes to qpixel

Istiakshovon:
![Screenshot from 2021-08-02 03-14-21](https://user-images.githubusercontent.com/58106197/127819393-3d824aaf-b4ec-401c-a8ff-f37992c2d8bb.png)

cellio:
Sounds like this needs changes in both CoDesign and qpixel.  Could you create an issue in CoDesign to track it and link the two of them together?  Thanks.

[...]

PR #597, opened 2021-08-04
dark theme

b-Istiak-s:
https://github.com/codidact/qpixel/issues/565

Any suggestion and helps are welcome. Ohh! Wait suddenly I noticed theme isn't working for `/users/edit/profile/` (I can't find problem anywhere). I will make changes dark, light radio button. But, I am not a professional graphics designer. And, I haven't used AI, PS for long days. So, I may forgot lot of things. So, some design for radio button is requested. I will try also (Sorry! I didn't mention earlier). To test the theme I have created [~~a zip~~ a 7z (zip file became larger than 25 MB so, I had to change format)](https://github.com/Istiakshovon/Istiakshovon/blob/main/qpixel.7z). Since, some design were changed to light theme while doing these changes. So, I would request to take a look at light theme also. (I am busy with some other works. So that, I use laptop rarely. That's why I can't work properly)... (Now, have to take a look at light theme. I can see some problems in light theme)(@kevinmmansour  helped with some work also) (Sorry for merging! It had shown in my git but, nothing changed)

Task left in Dark theme : 
- box shadow

Task left in Light Theme:
- have to look at whole theme

<details>
<summary>Dark Theme</summary>

image
image
image
image
image
image
image
Screenshot from 2021-08-04 11-46-56 

</details>

<details>
<summary>Light Theme</summary>

image
image
image
image
image

</details>

After new changes...

<details>
<summary>New Light Theme</summary>

image
image
image
image
image
image
image
image
image

</details>

<details>
<summary>New Dark theme</summary>

image
image
image
image
image
image

</details>

<details>
<summary>New changes</summary>

Screenshot from 2021-08-05 11-57-05
Screenshot from 2021-08-05 11-57-14
Screenshot from 2021-08-05 11-57-24
Screenshot from 2021-08-05 11-57-33

</details>


kevinmmansour:
Could we improve the preference tab for exampleing adding label to make sure that users will understand what is dark and light? Otherwise everything is cool.

Preview

cellio:
Can we make this preference like the other ones?  Not two radio buttons at the top, but a section under either Network Preferences or Community Preferences (hmm, which should it be?) with the layout and controls used for the others.

Could you vote on some of those posts so we can see what the score widget looks like?  And could you take a screenshot of something with pagination, to make sure those controls are legible?

Does this code bypass CoDesign or is there a corresponding PR there?  I don't know a lot about how things are put together, but dark theme shouldn't break if we change/add something in CoDesign.

Thanks.

kevinmmansour:
Since I have got a notification, There are some icons became dark in Light Mode and they should be white, for example: 

128215234-6f8c1c6e-722d-4508-900f-95e763d1a2ce

But, they are white in Light Mode in production sites:

Screenshot 2021-08-04 190633

Istiakshovon:
>Can we make this preference like the other ones? Not two radio buttons at the top, but a section under either Network Preferences or Community Preferences (hmm, which should it be?) with the layout and controls used for the others.

I think you meant to put those radio button system in header, didn't you? Usually, earlier I was doing the same. But, I noticed that mobile UI is little bit different. I was using icon (while putting theme controller in top). And, the header becomes very small for Android. That's why I thought it would be better if I use those theme controller system in preferences.

>Does this code bypass CoDesign or is there a corresponding PR there? I don't know a lot about how things are put together, but dark theme shouldn't break if we change/add something in CoDesign.

As far as I know CoDesign can't break my code. I used `!important` in most of code (If I didn't add !important than, CoDesign could break my things). Currently, I am not using !important for some codes. Since, they are working properly. If we do any changes to CoDesign and, use !important there. Than, CoDesign could break my theme. And, if I write some code there and don't use !important than, those things can't do anything with my code since, my code are on top/above (application.scss) of them.

Istiakshovon:
I guess. The work is completed. Check new images [here](https://github.com/codidact/qpixel/pull/597#issue-703660005). I will change `.zip` file ASAP.

cellio:
On the preference, I meant something like this:

mockup

Add a regular preference, not radio buttons, in the place where regular preferences go.  TBD: should this be a network setting or a per-community setting?

kevinmmansour:
> 
> 
> On the preference, I meant something like this:
> 
> mockup
> 
> Add a regular preference, not radio buttons, in the place where regular preferences go. TBD: should this be a network setting or a per-community setting?

As far as I know, it should be in the network setting. But I'm not an expert at all.

Istiakshovon:
@cellio 
>Add a regular preference, not radio buttons, in the place where regular preferences go. TBD: should this be a network setting or a per-community setting?

Usually, if someone chooses dark theme in Codidact Meta. Than, it should work in other network either. That's what my code says.

cellio:
> > TBD: should this be a network setting or a per-community setting?
> 
> Usually, if someone chooses dark theme in Codidact Meta. Than, it should work in other network either. That's what my code says. But, I don't have experience with sub-domain. So, my code completely depends on server.

Making it a network preference seems reasonable to me.  Could you test with Mathjax and code syntax highlighting to make sure those are ok?  I wondered about per-community in case something specialized doesn't work and people would want to be able to turn it off there, but it's better to just make sure it works well everywhere and then let people have a consistent experience.

I'm not sure what you mean about setting it on Meta.  Just to confirm, somebody can set this preference on any community and it'll apply across the network, right?  You don't have something specific about Meta in the code, right?

(Aside: could you try to use descriptive commit messages instead of "update <filename>"?  That would help me understand what the change is, since I can't always figure it out from the code -- like, maybe you've already done stuff I'm asking for but I can't tell, and I don't want to waste your time with redundant questions.  Thanks.)

Istiakshovon:
>Could you test with Mathjax and code syntax highlighting to make sure those are ok?

I don't have MathJax access in my server. I have create new server to test MathJax. There should be a way to enable MathJax. But, I don't have idea. According to me, MathJax are inside `p` tag. So, I guess they are considered as normal text (without thinking of `mjx-container`).

>it'll apply across the network, right?

According to me, yes. If it doesn't work than, I guess there's no way to do it. Cause, I don't have any idea of server. (I talked about Meta as example, I will use `e.g` next time :) ). 

>I don't want to waste your time with redundant questions.

No! It's OK. Question helps to find out more problems. I have found another problem after seeing the comment. And, I will try to write descriptive commit messages next time...

cellio:
You can enable MathJax via Admin Tools -> Site Settings -> Display and look for the "enable MathJax" setting.  (I assume you're an admin on your own server.  If not, you can change that. :-) )

cellio:
@ArtOfCode-  can you provide some guidance on how to wire up the network preference correctly?  Thanks.

Istiakshovon:
Screenshot from 2021-08-02 02-08-04
Screenshot from 2021-08-06 00-14-56
Screenshot from 2021-08-06 00-15-12


Is there any left which I should test? <3 :)

ArtOfCode-:
So I've only just had a chance to look at this. I appreciate the effort you've put into this, but I'm concerned that this is mostly being done in the "wrong" place - theming CSS and variables should generally be done in Co-Design, not in QPixel-specific CSS. 

I'd also suggest that instead of using a JS trigger for theme changes (and thus restricting dark theme to only those with JS enabled), we should use a media query (`@media screen and (prefers-color-theme: dark)`) - that way we both make the theme available to everyone and avoid the need to add a user preference because it's a browser setting.

Istiakshovon:
@ArtOfCode- i understood your concern. But, these days i couldn't understood how CoDesign was working with qpixel. Today, i have understood what should I do with CoDesign. So, it will take me few weeks to done it in CoDesign. So, other "things" is upto you. 

I didn’t use media query it wasn’t working for me. I was trying this way https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme at first, i tried by creating some HTML and CSS files. It wasn’t working. But, i again tried in qpixel. Although, it wasn’t working. So, i didn’t work that way. My default theme wad forever dark. I don’t know why i couldn’t use dark theme. Thats why i didn’t use media query.

Istiakshovon:
```
<!DOCTYPE html>
<html>
<head>
	<style>
	html {
	    content: ""; /* (ab)using the content property */
	}

	/* Light mode */
	@media (prefers-color-scheme: light) {
	    html {
		content: "light"; /* (ab)using the content property */
	    }
	}

	/* Dark mode */
	@media (prefers-color-scheme: dark) {
	    html {
		content: "dark"; /* (ab)using the content property */
	    }
	}
	</style>
</head>
<body></body>
</html>
```
That's how the code looks like. But, I wonder I always can see light theme not dark.. Is there any way to change theme in Chrome (Or, should I change theme in my system (If yes, than I am using Dark Theme. I am also using Dark Theme in Chrome. But, I don't know if there some other way to do that))?

PR #605, opened 2021-08-07
update text

Istiakshovon:
It should be Copy Link by default cause, it is showing "Permalink" in post page. When user click on that they get Copied then, Copy Link. So, Copy link is better than Permalink I guess.


cellio:
Does it copy the link?  It should only say "copy link" if, like for posts, it copies the link to the clipboard.  (And ideally it'd show you feedback, like it does for posts.)

ArtOfCode-:
Yep, it's a copy link not just a normal link.

Issue #582, opened 2021-07-30
Configurable length limits for content


cellio:
Inspired by: 
https://meta.codidact.com/posts/282950  
https://meta.codidact.com/posts/282973  

We currently hard-code the following length limits:

- post title: 1000 characters
- post body: 30,000 characters
- comment: 500 characters
- (comment thread title? unknown)

There are some questions/comments in those meta posts about whether limits are, or could be, configurable per community.  It seems reasonable to me that a community might want articles to be able to be longer, and maybe now that comments are threaded and thus out of the way, some might want to increase the limit there to avoid the "1/3" pattern.  I don't know if there's a use case for decreasing any of these limits (though one person on meta thinks 1000 characters is too long for titles, so who knows?).

Can we make these length limits configurable per community?  (Per category seems too complex without a strong use case.)  The first two could be part of the configuration of the post type; comment limits would need to go somewhere in site settings.

Istiakshovon:
>post title: 1000 characters
No! The limit is 250 character. Tested [here](https://a.dev.codidact.org/posts/174). Tested with 300 characters also. I am looking at db now(haven't find anything), will edit the comment if I find

cellio:
Related (minimum values):

https://github.com/codidact/qpixel/issues/1092 
https://github.com/codidact/qpixel/issues/1103

PR #607, opened 2021-08-08
Add maximum title length controls

sau226:
We've had a bunch of testers create 200+ character titles for posts on the dev server. I haven't seen this on prod, but to avoid any long term issues its best to place an upper bound on titles.

This patch adds a validator to check the title max length against a site setting value. There is no provision in the code to retroactively enforce this.

Max title length site setting defaults to 150 chars as that is what appears to be the equivalent limit for sites we have previously imported data from.

This also makes some small steps toward resolving issue #582.


Istiakshovon:
Sorry to say! I had done something better way :( . I got busy that's why I couldn't make Pull Request.... I will show what I did..

PR #608, opened 2021-08-08
show number whenever body or, title tends their limit

Istiakshovon:
length isn't showing for title, body and, thread title. I haven't worked with thread title but, post title and body. [User friendly approach](https://meta.codidact.com/comments/thread/3761). 

For post title :

If char is more than 200 and less than 255 than, the color will be yellow like thread body. And, Save Post button will be disable if title is greater than 255. 

For Post Body : 

Post body works same way but, if char is more than 29950 than color will be yellow. And, red when it is more than 30000.

Related [PR by @sau226 ](https://github.com/codidact/qpixel/pull/607)

Screenshot from 2021-08-08 10-13-49
Screenshot from 2021-08-08 10-14-21
Screenshot from 2021-08-08 10-14-28

The text only shows when character.length approaches it's limit


ArtOfCode- (changes requested):
We have an existing way of doing this. Instead of adding new code and repeating yourself, we should use that instead. See [_new_thread_modal.html.erb](https://github.com/codidact/qpixel/blob/edf31dd45613210ec2d5550f2de8c2ab92e33ad9/app/views/comments/_new_thread_modal.html.erb) for an example. In steps:

Add this element below the text field you want to display the character count for:

```html
<span class="has-float-right has-font-size-caption js-character-count-<%= post.id %>"
      data-max="500" data-min="15">
  <i class="fas fa-ellipsis-h js-character-count__icon"></i>
  <span class="js-character-count__count">0 / 500</span>
</span>
```

Change `js-character-count-<%= post.id =>` to use something unique to the field you're putting it on, instead of the post ID. Change `data-max` and `data-min` to the correct values. Change the text `0 / 500` to show the correct count.

Once you've done that, add a `data-character-count` attribute on the text field, containing a CSS selector for the element you just added above. In the new thread modal, that's done in Ruby with:

```ruby
data: { character_count: ".js-character-count-#{post.id}" }
```

Change the selector to match whatever you made the selector on the counter element.

Istiakshovon:
@ArtOfCode- But, there's a problem I don't want to show character count by default or, until it reaches maximum value. Should I use post.js that moment?

ArtOfCode-:
@Istiakshovon In that case, add a `data-display-at="0.75"` attribute to the counter element (same one where `data-max` is). That'll keep it hidden until it's in a warning state.

ArtOfCode- (changes requested):
You don't need the `data-character-count` attribute on the submit buttons, having it on the text field is enough. You've got it for the title field, but you still need to add the attribute to the body field, which is in `app/views/shared/body_field.html.erb`.

ArtOfCode- (approved):

PR #611, opened 2021-08-09
character length

Istiakshovon:
It shouldn't show by default which means the text won't show until your char length approach it's limit...

```
if (displayAt) {
      if (count >= displayAt * max) {
        $counter.show();
      }
      else {
        $counter.hide();
      }
    }
```

~~Somehow, above code isn't hiding. So, I think it shouldn't show either (after merging my PR).~~

<hr/>

It was happening in Dev Server. Maybe, MoshiKoi's PR wasn't deployed

### DIFF
--- a/app/views/posts/_form.html.erb
+++ b/app/views/posts/_form.html.erb
@@ -65,7 +65,7 @@
     </div>
     <div>
       <span class="has-float-right has-font-size-caption js-character-count-post-title"
-            data-max="255" data-min="15" data-display-at="0.75">
+            data-max="255" data-min="15" data-display-at="0.75" style="display:none">
         <i class="fas fa-ellipsis-h js-character-count__icon"></i>
         <span class="js-character-count__count">0 / 255</span>
       </span>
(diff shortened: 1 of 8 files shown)


app/views/posts/_form.html.erb
```diff
@@ -65,7 +65,7 @@
     </div>
     <div>
       <span class="has-float-right has-font-size-caption js-character-count-post-title"
-            data-max="255" data-min="15" data-display-at="0.75">
+            data-max="255" data-min="15" data-display-at="0.75" style="display:none">
```
ArtOfCode-, line 68: Instead of using an inline `style` attribute, add `hide` to the `class` attribute (and same for the other one).

ArtOfCode- (changes requested):

app/views/posts/_form.html.erb
Istiakshovon: `class= hide` Did you mean it?

ArtOfCode-: Yes, but since the `span` element already has a `class` attribute, just add `hide` to the end of the list.

PR #614, opened 2021-08-10
Remove username from account edit form

### LINKED ISSUE #576: Username is editable from both Account and Edit, and with different security
On the user page, you can edit your username on both the Account and Edit tabs.

When editing from the Account tab, you must provide your password to confirm changes. When editing from the Edit tab, you don't need to provide your password.

I think either both of them should require the password or neither of them should, right now it seems like an oversight and one of them is too strict or too lax.

sau226:
This PR removes the username field from the devise edit form (which backs users' account tabs on their profile page). 

This reduces duplication of fields, as username is now centrally managed in the same place where profile images, bios and social links are managed instead of a page with more sensitive controls (such as 2fa, email and password). 

I briefly tested the change on the dev server with a live code edit that I then reverted - since it is an html erb file I didn't go and restart the qpixel process before/after the test.

The matter was discussed in the issue linked by the linking tag below.

GitHub linking tag: Resolves #576.

### DIFF
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -19,11 +19,6 @@
   <% end %>
 
   <div class="form-group">
-    <%= f.label :username, class: "form-element" %>
-    <%= f.text_field :username, class: 'form-element', autocomplete: 'off' %>
-  </div>
-
-  <div class="form-group">
     <%= f.label :password, class: "form-element" %>
     <div class="form-caption">
       Leave blank if you don't want to change it.


ArtOfCode- (approved):
This works as-is, so I'm going to merge it, but just FYI (which you may already have realised) - this only removes the front-end field. The server will still accept a `username` field in the POST request, and will still apply it if it's submitted. To avoid that, you can change the [`configure_permitted_parameters` method](https://github.com/codidact/qpixel/blob/683257bf4ab5ab3fa25b74b10c8cf3ab3bf4784b/app/controllers/application_controller.rb#L31) in `ApplicationController` to remove the `:username` parameter from `account_update`.

Issue #204, opened 2020-10-09
Live update for comment count?


cellio:
https://meta.codidact.com/questions/278059 reports that, after leaving a comment, the comment count ("N comments") doesn't update until the page is refreshed.  I don't know if that's intentional (performance) or an oversight, so this change request is to evaluate whether it makes sense to do "live updates" and, if so, do that.

Istiakshovon:
I think the issue should be closed (It is not needed anymore).

kevinmmansour:
> 
> 
> I think the issue should be closed (It is not needed anymore).

Why it is needed anymore? Did it get solved? I don't think it is solved.

Istiakshovon:
@kevinmmansour  Actually, the problem was happening for normal comment. Now, we have thread comment. Even, thread comment doesn't work that way. So, it have to be closed

PR #618, opened 2021-08-11
Show character length while editing comments

Istiakshovon:
https://meta.codidact.com/posts/282342/282410#answer-282410
https://github.com/codidact/qpixel/issues/204

 I haven't test the code. 
 
 tested... not working... I will check it

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -80,8 +80,8 @@ $(() => {
       <textarea id="comment-content" rows="3" class="form-element is-small" data-character-count=".js-character-count-comment-body" name="comment[content]">${content}</textarea>
       <input type="submit" class="button is-muted is-filled" value="Update comment" />
       <input type="button" name="js-discard-edit" data-comment-id="${commentId}" value="Discard Edit" class="button is-danger is-outlined js-discard-edit" />
-      <span class="has-float-right has-font-size-caption js-character-count-comment-body hide"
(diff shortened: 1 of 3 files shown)


Istiakshovon:
It's working now.. But, I can see some problem right there....

@ArtOfCode- you told me to use `hide` class instead of style="display:'none' ". I noticed that the hide class isn't removing whenever character touches 75% of that.

ArtOfCode-:
Inline `style` attributes aren't a good thing to use in most cases, because they override basically everything which makes it very difficult to change styles anywhere else you want to do so. This is happening because I wrote the code to hide the character counter without taking into account how our `hide` class works, so instead of adding and removing that class I've used jQuery's specific `hide()` and `show()` methods, which work differently. I'm going to merge this as-is for now and hopefully get round to fixing that so we can remove the inline `style` at some later date.

cellio:
I just tried to edit a comment and I don't see a counter.

Istiakshovon:
> I just tried to edit a comment and I don't see a counter.

It will show when you approach limit 
Screenshot from 2021-08-16 15-12-10

cellio:
Ah.  I was confused because for a *new* comment the counter is always there.

Istiakshovon:
> Ah. I was confused because for a _new_ comment the counter is always there.

So, should we show that counter always?

cellio:
I think it would be better if we matched with comment creation, thanks.

PR #621, opened 2021-08-12
copy-paste text in fields

Istiakshovon:
https://meta.codidact.com/posts/283228

Canina already have reported twice that when he copy-paste text in comment field the state of submit button isn't changing. This was happening cause, earlier were only checking `keyup change` but, not copy pasted text. So, after searching little bit I found that there's a way to detect [pasted text](https://infoheap.com/jquery-text-input-field-change-keyup-paste-events/) also. Usually, earlier code (function) wasn't inside any variable or, didn't declare as function so, I just had to copy and paste. There's another way to shorten the code (Just give function a name and, call that on keyup change and paste).

### DIFF
--- a/app/assets/javascripts/character_count.js
+++ b/app/assets/javascripts/character_count.js
@@ -4,7 +4,7 @@ $(() => {
     el.removeClass(icons.join(' ')).addClass(icon);
   };
 
-  $(document).on('keyup change', '[data-character-count]', ev => {
+  $(document).on('keyup change paste', '[data-character-count]', ev => {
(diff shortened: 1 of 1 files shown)


ArtOfCode- (changes requested):
Instead of copying all the existing code to a new event handler, you can just add `paste` to the list of events being listened to on line 7. You only need to create a new event handler if the action to be taken is different, but in this case it's exactly the same.

ArtOfCode- (approved):

PR #627, opened 2021-08-16
Close notification popup on outside click

Istiakshovon:
Fixed https://meta.codidact.com/posts/276941..

Usually, I don't know how you were closing popup when clicking `js-inbox`. That's why I am just writing what worked for me

~~Seems like, clicking on notification button doesn't close it.~~ (fixed)


ArtOfCode-:
This has already been fixed in Co-Design, we don't need to fix it again here.

PR #628, opened 2021-08-16
User can edit, delete or undelete from post page

Istiakshovon:
When I was trying to remove inline from flag I noticed, flag's popup isn't showing in post page. That's why  I am not removing inline from flag now. It needs some more works I think.

Usually, I was trying to find which javascript file is fetching data-drop. But, I didn't find any specific page. I was searching in flags.js that page didn't add is-active class either. That's why I can't take a look for flags. I was trying to add is-active my own way. But, there was some problem. That's why I am not doing that.

### DIFF
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -42,7 +42,7 @@
       <% if with_post_link %>
         <%= link_to 'post', generic_share_link(comment.post) %>
       <% end %>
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
Just removing the check for `params[:inline]` won't work. The JS won't pick up comments that are added after the page is loaded (i.e. any time you open a thread, at the moment). That's down to how Co-Design works, which I seem to remember I've fixed for the next version, but needs building and pushing.

Issue #191, opened 2020-10-08
No character count when editing comments (unlike when creating them)


cellio:
https://meta.codidact.com/questions/276896

When you type a comment, you get a character count and it turns red when you exceed the length limit (so you know you have to cut it down).  When you edit an existing comment, however, you don't get this indicator, so you can exceed the limit and not realize it (and then wonder why your edit isn't being accepted).  Can we have the same counter for edits as we do for creating?  (This would mean that upon edit, it would have to check the existing length, as opposed to new comments that just start at 0.)

galegosimpatico:
[complexity: average](https://github.com/codidact/qpixel/labels/complexity%3A%20average)

cellio:
I am raising the priority because, as it turns out, we didn't fix this as part of the bigger reworking for threaded comments.  This should be easy and it's a source of frustration and confusion for users; can we fix it?

PR #629, opened 2021-08-16
Show counter always

Istiakshovon:
https://github.com/codidact/qpixel/pull/618#issuecomment-899615434


cellio:
Fixes https://github.com/codidact/qpixel/issues/191.

PR #634, opened 2021-08-17
change background color of code area in dark theme

b-Istiak-s:
It's useless without [this](https://github.com/codidact/co-design/pull/60).

I am making the PR without testing. I know which color I chose. And, I may not test it cause, I am not going to use Debian for few days. :( 

Code area (pre tag) and, user-card-meta's background

**Make sure that you are going to commit to develop CoDesign's PR before merging it.**

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -16,11 +16,6 @@
 
 @import 'variables';
 
-@media (prefers-color-scheme: dark){
-  background-color: rgb(30, 30, 30);
-  color: white;
-}
-
 h1, h2, h3, h4, h5, h6 {
   font-family: $font-stack-display;
 }
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -203,12 +203,18 @@ img {
 .widget.is-tertiary .widget--body + .widget--header {
   border-top: 1px solid #9daeb7;
 }
-
-pre {
-  background: #f0f0f0;
-  border: 0;
+@media (prefers-color-scheme: light) {
+  pre {
+    background: #f0f0f0;
+    border: 0;
+  }
 }
-
+@media (prefers-color-scheme: dark){
+  pre {
+    background: #2c2a2a;
+    border: 0;
+  }
+}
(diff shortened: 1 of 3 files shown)


ArtOfCode-:
Seems like this should be part of the existing Co-Design PR?

Istiakshovon:
> Seems like this should be part of the existing Co-Design PR?

I have made changes to 3 pages. So, I am including all of them one by one. 

https://github.com/Istiakshovon/qpixel/blob/6ed0026acd3cbbc65725dde6ad2698c198c5ef91/app/assets/stylesheets/application.scss#L212-L220
I had made changes to it in typography. https://github.com/codidact/co-design/blob/46e7b4c3e1db101d8b91d080cd0a58b1d7b97def/src/common/_typography.scss#L114-L122

But, you had also coded in application.scss that's why I had to make changes to the page also. Otherwise, it might affect design. Since, application is saying to make #f0f0f0 forever. I know one will be ignored and, another will work. But, as far as I can tell that is CoDesign will be ignored that's why I made changes to it either.

I had made changes for body https://github.com/Istiakshovon/qpixel/commit/1ee83f6fabc2922da310fc5b4ee4047a57ea246f#diff-0869e3045a3b5ca0e44ed193977d1568d686ca2373f4b93a8a868e4099fc932dR19-R22
Unfortunately, I forgot to write body. But, it's available in CoDesign that's why I had deleted it again. https://github.com/codidact/co-design/blob/db4527d664bc90c340390703cdd066e7344ca3a1/src/common/_layout.scss#L67-L72 I had forgot to write color earlier that's why I committed there again.

Another one is user-meta-card, 
Screenshot from 2021-08-21 16-33-37
According to inspect element, it is only available in application.scss if it is available in CoDesign than, I don't know cause, it wasn't available in my work place.

As far as I remember from my another [PR](https://github.com/codidact/qpixel/pull/597), I have to make change to something else either which is user-list-user (only for hover). Here's the line. https://github.com/Istiakshovon/qpixel/blob/6ed0026acd3cbbc65725dde6ad2698c198c5ef91/app/assets/stylesheets/users.scss#L19

Issue #633, opened 2021-08-17
Restrict user to use $$$$ for title


Istiakshovon:
I was trying to use $$$$ in title to see it works or not. Usually, SE doesn't allow user to use $$$$ in title. 
![Screenshot from 2021-08-17 17-48-38](https://user-images.githubusercontent.com/58106197/129694784-98facdaa-b1c9-4153-986f-ff8558be30d3.png)
Using $$$$ is completely bad so, I think we should restrict that.

PR #642, opened 2021-08-23
let user don't use $$Math$$

Istiakshovon:
I am not sure how can I update text of error. 

Error : translation missing: en.You shouldn't use $$

I don't have any idea why translation missing is showing. I guess, `, type: @post_type.name` not required

Resolves : https://github.com/codidact/qpixel/issues/633

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -52,7 +52,7 @@ class PostsController < ApplicationController
     if @post.title?
       if @post.title.include? "$$"
         flash[:danger] = I18n.t 'posts.no_block_mathjax_title'
-        redirect_back fallback_location: root_path
+        render :new, status: :bad_request
         return
       end
     end
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -51,7 +51,7 @@ class PostsController < ApplicationController
 
     if @post.title?
       if @post.title.include? "$$"
-        flash[:danger] = helpers.i18ns('You shouldn\'t use $$', type: @post_type.name)
+        flash[:danger] = helpers.i18ns('Title can not contain "$$"', type: @post_type.name)
         redirect_back fallback_location: root_path
         return
       end
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -49,6 +49,14 @@ class PostsController < ApplicationController
     @post = Post.new(post_params.merge(user: current_user, body: helpers.post_markdown(:post, :body_markdown),
                                        category: @category, post_type: @post_type, parent: @parent))
 
+    if @post.title?
+      if @post.title.include? "$$"
+        flash[:danger] = helpers.i18ns('You shouldn\'t use $$', type: @post_type.name)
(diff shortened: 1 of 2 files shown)


app/controllers/posts_controller.rb
```diff
@@ -49,6 +49,14 @@ def create
     @post = Post.new(post_params.merge(user: current_user, body: helpers.post_markdown(:post, :body_markdown),
                                        category: @category, post_type: @post_type, parent: @parent))
 
+    if @post.title?
+      if @post.title.include? "$$"
+        flash[:danger] = helpers.i18ns('Title can not contain "$$"', type: @post_type.name)
```
ArtOfCode-, line 54: `helpers.i18ns` is a wrapper round `I18n.t` that substitutes text in before it's sent for translation. This call is trying to substitute `":type"` in the string for `@post_type.name`, but you haven't got and don't need `:type` in the string, so you can remove this entirely.

The argument to `i18ns` (and `I18n.t`) should be the _translation key_, not the string you want translated. The actual string needs to go in a locale file.

Let's also be a bit clearer about the error message.

```suggestion
        flash[:danger] = I18n.t 'posts.no_block_mathjax_title'
```

Then you also need to add another entry in `config/locales/strings/en.posts.yml`, like the following:

```yml
  no_block_mathjax_title: >
    Title cannot contain block-level MathJax.
```

ArtOfCode- (changes requested):

config/locales/strings/en.posts.yml
```diff
@@ -64,3 +64,5 @@ en:
       In a URL of "https://yoursite.codidact.com/help/topic", the "topic" is the slug.
     responding_to: >
       Responding to:
+    no_block_mathjax_title: >
+     Title cannot contain block-level MathJax.
```
ArtOfCode-, line 68: You need one more space at the start of this line. YAML is very picky about indentation.

ArtOfCode- (changes requested):

app/controllers/posts_controller.rb
```diff
@@ -49,6 +49,14 @@ def create
     @post = Post.new(post_params.merge(user: current_user, body: helpers.post_markdown(:post, :body_markdown),
                                        category: @category, post_type: @post_type, parent: @parent))
 
+    if @post.title?
+      if @post.title.include? "$$"
+        flash[:danger] = I18n.t 'posts.no_block_mathjax_title'
+        redirect_back fallback_location: root_path
```
ArtOfCode-, line 55: Sorry, missed this one - last one! Can we make this a `render :new, status: :bad_request` instead, so that we keep the post details submitted.

```suggestion
        render :new, status: :bad_request
```

ArtOfCode- (changes requested):

app/controllers/posts_controller.rb
Istiakshovon: Did you mean if user enter's "$$" than, we let them post that?

ArtOfCode-: No, that doesn't actually post it. Using `redirect` means that the form will be cleared. Using `render` means that the text will stay so it can be edited.

Issue #643, opened 2021-08-23
Add "Linked Answer" header or similar to the answer when using a direct answer link


MoshiKoi:
We've already gotten two reports of confusion because of answer-floating, ([here](https://meta.codidact.com/posts/283346) and [here](https://software.codidact.com/posts/283667))

Clearly, this behavior is confusing users. I suggest adding something like "Linked Answer" above the answer to make it clear to users that the specific out-of-order answer is being specially considered.

Here's a quick Inspect Element mockup I made to illustrate:

!["Linked Answer" header above answer](https://user-images.githubusercontent.com/54333972/130522435-04f6dbe0-907f-4a37-82cc-0e336da2b324.png)

PR #644, opened 2021-08-23
Show info for floated posts

luap42:
closes: #643 

untested (edited only in GitHub), might test out later unless someone does before me.

### DIFF
--- a/app/views/posts/_expanded.html.erb
+++ b/app/views/posts/_expanded.html.erb
@@ -60,7 +60,7 @@
       <% if !is_top_level && !params[:answer].nil? && params[:answer] == post.id.to_s %>
         <div class="notice is-info">
           <p>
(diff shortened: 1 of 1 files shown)


app/views/posts/_expanded.html.erb
```diff
@@ -56,6 +56,16 @@
     <% end %>
 
     <div class="post--content has-padding-2">
+      
+      <% if !is_top_level && !params[:answer].nil? && params[:answer] == post.id.to_s %>
+        <div class="notice is-info">
+          <p>
+            You are accessing this answer with a direct link; it is therefore shown above all other answers regardless of it's score.
```
cellio, line 63: Nit: "its", not "it's".

luap42: fixed.

Issue #630, opened 2021-08-16
Create Post button appears even when you don't have permission to post there, and you can still draft posts there


MoshiKoi:
Even though I do not have permission to post in the [Meta Blog category](https://meta.codidact.com/categories/30), the "Create Post" button still appears, and I can click on it to create a draft of a post in that category.

When clicking "Save post in Blog", it gives a red warning notice that says

>The following errors prevented your post from being saved:
>
> - You don't have a high enough trust level to post in the Blog category.

If I do not have permission to post in that category, could the button be removed so that it's more obvious before that notice is given?

cellio:
I *think* this is easy but I'm not sure.  If it's easy let's prioritize it; it's an infrequent case but can be very frustrating for those who hit it.

We do a check on submit (and still should, in case of shenanigans), but can we *also* do a check either on page view (gray out the button) or, if that's too expensive, on button click, so somebody doesn't write a whole blog post only to find out *after* that the blog category is restricted?

PR #651, opened 2021-08-25
User can't see form when their trust level is low

Istiakshovon:
Resolves : https://github.com/codidact/qpixel/issues/630

### DIFF
--- a/app/views/posts/new.html.erb
+++ b/app/views/posts/new.html.erb
@@ -17,7 +17,9 @@
     </div>
   <% end %>
 
-  <% if @post.errors.any? %>
+  <% if !@post.errors.any? %>
+   <%= render 'form', post: @post, post_type: @post_type, category: @category, submit_path: request.path %>
+  <% else %>
     <div class="notice is-danger is-filled">
       <p>The following errors prevented your post from being saved:</p>
       <ul>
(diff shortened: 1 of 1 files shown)


app/views/posts/new.html.erb
```diff
@@ -1,30 +1,40 @@
...
+      </ul>
     </div>
-  </div>
-<% end %>
+  <% end %>
 
-<% if @post.errors.any? %>
+  <%= render 'form', post: @post, post_type: @post_type, category: @category, submit_path: request.path %>
+<% else %>
   <div class="notice is-danger is-filled">
```
ArtOfCode-, line 33: We don't really need the lead text and bullet points here - this whole `<div>` can just be a single error message:

```erb
<div class="notice is-danger is-filled">
  You don't have a high enough trust level to post in the <%= @category.name %> category.
</div>
```

ArtOfCode- (changes requested):

app/views/posts/new.html.erb
Istiakshovon: How can a user know that they don’t have high enough trust level for posting in category without a error message? I was thinking that tips of posting in blogs is part of form so, it shouldn’t show for low trust level users. 🤔 😴

PR #652, opened 2021-08-26
Button is disabled without any explanation when title is too short (fixed)

Istiakshovon:
first half of this one https://github.com/codidact/qpixel/pull/641

### DIFF
--- a/app/assets/javascripts/character_count.js
+++ b/app/assets/javascripts/character_count.js
@@ -7,7 +7,6 @@ $(() => {
   $(document).on('keyup change paste', '[data-character-count]', ev => {
     const $tgt = $(ev.target);
     const $counter = $($tgt.attr('data-character-count'));
-    const $formElement = $tgt.attr('data-field-name');
     const $button = $counter.parents('form').find('input[type="submit"]');
     const $count = $counter.find('.js-character-count__count');
     const $icon = $counter.find('.js-character-count__icon');
@@ -47,7 +46,7 @@ $(() => {
       if ($button) {
         $button.attr('disabled', true).addClass('is-muted');
       }
-      document.getElementById($formElement).style.setProperty("border", "2px solid red", "important");
+      $tgt.css("border", "2px solid red", "important");
     }
     else {
       $counter.removeClass('has-color-red-500 has-color-yellow-700 has-color-primary');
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/character_count.js
```diff
@@ -46,13 +47,15 @@ $(() => {
       if ($button) {
         $button.attr('disabled', true).addClass('is-muted');
       }
+      document.getElementById($formElement).style.setProperty("border", "2px solid red", "important");
```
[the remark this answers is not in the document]
Istiakshovon: Somehow class isn't working for me. https://discord.com/channels/634104110131445811/672517380152361047/881069687897739354 I had messaged there. So, I am not making changes to it. If it works for you than, you can commit...

app/assets/stylesheets/application.scss
```diff
@@ -16,6 +16,10 @@
 
 @import 'variables';
 
+.red-700-border-2px-width{
```
ArtOfCode-: For a good semantic name, I'd suggest naming CSS classes in a way that describes what they're _for_, rather than what they _do_. You're using this one for an "invalid data" state, so you could call it `.invalid-input` or `.failed-validation` or something similar.

Also as a general rule: when adding new CSS, add it to the most specific file and append to the bottom. `application.scss` is the least specific file; it's just general application-wide styling. If this is only used on posts, you could append it to `posts.scss`, or if it's for comments, to `comments.scss`. (It doesn't actually matter if you use it elsewhere - all CSS is available everywhere, not just specific files in specific places, so just pick whatever it's _primarily_ for).

Istiakshovon: I directly used it in application.scss cause it is needed it lots of pages. users, comments, posts that's why I thought all of them call the class from application.scss

app/assets/stylesheets/comments.scss
```diff
@@ -143,3 +143,7 @@
   padding: 0.7em;
   display: none;
 }
+
+.failed-validation {
+  border: 2px solid #d21919 !important;
+}
```
ArtOfCode-, line 149: Sorry, I don't think I explained clearly enough in my previous comment. You only need to add new styles to _one_ file. Your style will be available everywhere regardless of what controller it's under.

ArtOfCode- (approved):

PR #656, opened 2021-08-28
Show if closed as duplicate... Show if question was reopened... Activity on answers

Istiakshovon:
Resolves : https://physics.codidact.com/posts/283783

Maybe, it's completed : https://meta.codidact.com/posts/279562 I am not sure if anything left for last_activity.

I didn't add locked_at cause, it was looking like when mod lock a post it doesn't bump. Hence, last_activity doesn't change. So, that's not required I think. (Closing doesn't available in answer so, I didn't add it either). If there's anything left then, comment.

Why I didn't fetch children under case?
When I was trying to fetch children under case it was automatically printing those arrays instead of last_activity. Something like returning those arrays instead of 'posted an answer' (or, something just like this).

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -98,6 +98,26 @@ class Post < ApplicationRecord
   end
 
   def last_activity_type
+    last_activity_on_post = last_activity
+    last_edited = children.map(&:last_edited_at)
+    last_edited_children = nil
+    last_edited.each do |x| if x==last_activity_on_post
+                              last_edited_children = x
+                            end
+    end
+    last_deleted = children.map(&:deleted_at)
+    last_deleted_children = nil
+    last_deleted.each do |y| if y==last_activity_on_post
+                              last_deleted_children = y
+                            end
+    end
+    last_created = children.map(&:created_at)
+    last_created_children = nil
+    last_created.each do |z| if z==last_activity_on_post
+                               last_created_children = z
+                             end
+    end
+    
     case last_activity
     when closed_at
       if duplicate_post
@@ -115,6 +135,12 @@ class Post < ApplicationRecord
       'deleted'
     when last_edited_at
       'edited'
+    when last_created_children
+      'posted an answer'
+    when last_deleted_children
+      'deleted an answer'
+    when last_edited_children
+      'edited an answer'
     else
       'last activity'
     end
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
I'm gonna decline this one, for now at least. Showing activity on answers requires us to load all answers to every question we're checking. On the category index page, that means we have to load every question plus all answers, just to show the questions. That doesn't make a huge amount of sense, and isn't going to be great for performance.

Istiakshovon:
So, first two commit helpful? `Show if closed as duplicate... Show if question was reopened`

cellio:
Instead of querying for activity on all answers at page-load time, which I agree would be too expensive, can we store, on a top-level post, the last child activity (if any), and use that when deciding what to show in the post list?  We only need the most recent, not everything from all children.  Doing things to children would have to update this field on the parent, which seems practical?

I don't know how we're storing the time and author of last activity now, though we obviously are, so maybe we could store the "what" at the same time, where "what" is things like answered, answer-edited, answer-deleted, etc?  If we also stored the post ID, we could make the last-activity link be a specific link too, without extra computation at page-load time.

I realize I'm talking around the c-word and cache invalidation is one of the two hard problems in computer science, but since we only ever overwrite this data (don't need to roll it back), maybe that's ok?

Istiakshovon:
@cellio currently, we are storing data of last_activity in DB. And, that shouldn’t be too hard to store last_activity of recent. All the things we have to do that is saving recent_activity in answer (ID of that post). If we save that then we can directly fetch a single answer only. 

But i worry of old post. Old posts didn’t have the field in DB. So i was thinking if it actually affects them. I think command-line may helpful that moment but, i don’t have any idea.

ArtOfCode-:
@Istiakshovon Yep, showing if the question was duped or reopened would be fine.

ArtOfCode-:
@cellio Yep, we could look at doing that. I'm pretty sure we already update last activity on the parent when an answer is changed, so extending that a bit shouldn't be too hard.

cellio:
We must be doing *something*, because you can see "last activity $time $user" when the last activity involves an answer.  That's why I thought it might not be hard to store a little *more* information there -- which answer (to enable linking), what action.  Sounds like I should file a separate issue so it's not tangled in this one?

PR #660, opened 2021-08-31
accurate length when user clicks on edit button

Istiakshovon:
Read [the thread's comment](https://meta.codidact.com/comments/thread/4145#comment-13137)

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -67,7 +67,6 @@ $(() => {
     const $commentBody = $comment.find('.comment--body');
     const commentId = $comment.attr('data-id');
     const originalComment = $commentBody.clone();
-    const $counter = $tgt.attr('data-body');
 
     const resp = await fetch(`/comments/${commentId}`, {
       credentials: 'include',
@@ -75,7 +74,6 @@ $(() => {
     });
     const data = await resp.json();
     const content = data.content;
-    const length =  $counter.length;
 
     const formTemplate = `<form action="/comments/${commentId}/edit" method="POST" class="comment-edit-form" data-remote="true">
       <label for="comment-content" class="form-element">Comment body:</label>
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/comments.js
```diff
@@ -83,7 +85,7 @@ $(() => {
       <span class="has-float-right has-font-size-caption js-character-count-comment-body"
             data-max="1000" data-min="15">
         <i class="fas fa-ellipsis-h js-character-count__icon"></i>
-        <span class="js-character-count__count">0 / 1000</span>
+        <span class="js-character-count__count">`+length+` / 1000</span>
```
ArtOfCode-, line 88: In a string wrapped with <code>`</code> backticks, you can use string interpolation, like Ruby:

```suggestion
        <span class="js-character-count__count">${content.length} / 1000</span>
```

app/assets/javascripts/comments.js
```diff
@@ -67,13 +67,15 @@ $(() => {
...
     const commentId = $comment.attr('data-id');
     const originalComment = $commentBody.clone();
+    const $counter = $tgt.attr('data-body');
 
     const resp = await fetch(`/comments/${commentId}`, {
       credentials: 'include',
       headers: { 'Accept': 'application/json' }
     });
     const data = await resp.json();
     const content = data.content;
```
ArtOfCode-, line 77: Here's where we get the comment content. It's fetched on line 71, the `await fetch`, and pulled out of the response data into the `content` variable here.

app/views/comments/_comment.html.erb
```diff
@@ -43,7 +43,7 @@
         <%= link_to 'post', generic_share_link(comment.post) %>
       <% end %>
       <% if user_signed_in? && (comment.user == current_user || current_user.is_moderator) && params[:inline] != 'true' %>
-        <a href="#" class="js-comment-edit">edit</a>
+        <a href="#" data-body="<%= comment.content %>" class="js-comment-edit">edit</a>
```
ArtOfCode-, line 46: See above - you don't need to add the content to a `data` attribute.

PR #663, opened 2021-09-02
show parent if available

Istiakshovon:
Resolves : https://meta.codidact.com/posts/283812

UI : 

Screenshot from 2021-09-02 18-20-01
<hr/>

Screenshot from 2021-09-02 18-20-08

Reason of pulling it : I have recently face the same problem again which was annoying. Cause, I was doing math but it's annoying to open new tab every times. That's why the pull request without any up or down vote

### DIFF
--- a/app/views/comments/thread.html.erb
+++ b/app/views/comments/thread.html.erb
@@ -2,13 +2,13 @@
 
 <% if @post.parent.present? %>
   <details>
-    <summary>Question</summary>
+    <summary>Parent</summary>
     <%= render 'posts/expanded', post: @post.parent %>
   </details>
 <% end %>
 
 <details>
-  <summary> <% if !@post.parent.present? %> Thread on question <% else %> Thread on answer <% end %> </summary>
+  <summary>Post</summary>
   <%= render 'posts/expanded', post: @post %>
 </details>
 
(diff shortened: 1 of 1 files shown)


app/views/comments/thread.html.erb
```diff
@@ -1,7 +1,14 @@
...
+<% if @post.parent.present? %>
+  <details>
+    <summary>Question</summary>
+    <%= render 'posts/expanded', post: @post.parent %>
+  </details>
+<% end %>
+
 <details>
-  <summary>Post</summary>
+  <summary> <% if !@post.parent.present? %> Thread on question <% else %> Thread on answer <% end %> </summary>
```
ArtOfCode-, line 11: This sort of indicates that the content behind the dropdown is the thread, rather than a post. I'd suggest leaving the wording as just "Post". With the change I suggested above that should be clear enough.

app/views/comments/thread.html.erb
```diff
@@ -1,7 +1,14 @@
 <% pingable = get_pingable(@comment_thread) %>
 
+<% if @post.parent.present? %>
+  <details>
+    <summary>Question</summary>
```
ArtOfCode-, line 5: Not all post types with parents are questions; we can dynamically add others. I'd suggest calling this "Parent post" (or you could fetch the post type and display its name, but "parent" covers all cases).

ArtOfCode- (changes requested):

Issue #664, opened 2021-09-03
The height of textarea of comment-content (editing comment field) should be increased


Istiakshovon:
![Screenshot from 2021-09-03 13-48-25](https://user-images.githubusercontent.com/58106197/131952035-54915d3d-3573-4634-8f76-be963c25b6a7.png)

When editing my comments, I saw that I can't see 4 lines properly also. Currently, we can write 1000 characters but, we can only see 3 lines properly while editing comment. There's some padding-top of `comment body` characters. Maybe, if we decrease the padding then the comment field may be better.

ArtOfCode-:
There's a balance to be hit here. We don't want comment input fields to be _too_ big, because that encourages people to write lots in comments when really if there's that much to say it should probably be an answer. We could probably increase to 4 rows to start with, maybe increasing to 5 later if that's not enough.

PR #667, opened 2021-09-06
4 rows for comment body (for editing)

Istiakshovon:
https://github.com/codidact/qpixel/issues/664#issuecomment-913190317


ArtOfCode-:
I think there's somewhere else this needs changing too, in the initial view as well as the edit JS

Istiakshovon:
I think you are talking about it 

https://github.com/Istiakshovon/qpixel/blob/patch-8/app/views/posts/_form.html.erb#L173-L176

Issue #546, opened 2021-07-08
Fully expand a comment thread in-page


cellio:
The most frequent request we've heard for our new threaded comment system is the ability to expand a comment thread in the post page, rather than having to navigate to the thread page (and lose the context of the post, other posts on the page, and other comment threads).

We still want the thread pages, but we should also have a way to access the entire thread from the post, including leaving a new comment.  Currently if you expand a thread you see the first five comments and then a link to the thread.  I'd like to add an option: along with a link to the thread page, add a "show all comments" link.  If you click that you get all of them, whether that's six or sixty.  Having to click through twice (once to see the thread at all, once to see all of it) adds the friction that we were looking for even without going to a new page.

I think it's ok to follow the Toyota principle here ("you asked for it, you got it") and just load all the comments in place, rather than wrapping the thread in a scrollable container.  Make sure that "collapse" link stays available and people can recover.

Finally, because you'll get the textbox to add a comment if you expand the whole thread, either you should get it "for free" if you expand a thread with fewer than five comments, or in the case where there are no more comments to show, the link should be "add comment", not "show all comments", because you're already looking at all of them and that could be confusing.

Istiakshovon:
I said in my earlier comment a thread comment might have 50 comments also. If we allow to expand in post page. Than, chrome might crash in low quality's mobile. Instead of expanding whole thread I think it would be better if we expand 10 comments per click... :thinking:

cellio:
I didn't realize that expanding comments could crash a browser.  Can you say more about what the issue is?  Does your browser crash if you expand a long Reddit thread or SE comment pile?  Options to expand "next 10" and "all" could work.  I don't want to force people to keep clicking if they want everything (maybe because their next step is going to be ctrl-F for that thing they remembered *somewhere* in that pile), but if loading a big pile is a usability issue on a phone, I do want to account for that.

Istiakshovon:
> I didn't realize that expanding comments could crash a browser. Can you say more about what the issue is? Does your browser crash if you expand a long Reddit thread or SE comment pile? Options to expand "next 10" and "all" could work. I don't want to force people to keep clicking if they want everything (maybe because their next step is going to be ctrl-F for that thing they remembered _somewhere_ in that pile), but if loading a big pile is a usability issue on a phone, I do want to account for that.

[...]

PR #668, opened 2021-09-06
Expand thread in post page

b-Istiak-s:
Said all the things right [here](https://github.com/codidact/qpixel/issues/546#issuecomment-913581697). Not sure if the method is acceptable or not...

Screenshot from 2021-09-09 09-00-25

Showing "Go to thread page" at bottom cause, @cellio had said that user won't like to move to top again. So, showing it with Add reply.

Oops, flag button wasn't hide.

~~Wait a minute... Currently, my code is fetching every comment twice.(fixed)~~

Currently, I can flag from inline preview. But, If I try to show tools same as flag then it will work for few items not for all cause, there's lot of tools which also contains modal. 

resolves : https://github.com/codidact/qpixel/issues/670
resolves : #546

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -45,7 +45,7 @@ $(() => {
     $inlineHeader.removeClass('hide');
     $jsExpandThread.addClass('hide');
     $inlineComments.removeClass('hide');
-    $comments.removeClass('hide');
+    $comments.addClass('hide');
     //$commentsAdditionalTools.removeClass('hide');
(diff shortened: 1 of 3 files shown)


cellio:
What is the current status of this PR?  Needs a review?

b-Istiak-s:
> What is the current status of this PR? Needs a review?

I think @ArtOfCode- and @luap42 thinks it shouldn’t be done this way. As a programmer (from another sector), I think it should be done some other way even I was thinking it when working on it. Latter, i just gave a try.

PR #723, opened 2021-11-14
Add email sender name

sau226:
This PR makes changes to the devise configuration files to add a proper sender display name on our emails.

Currently, when you open the email in free webmail services (think GMail or Outlook), the sender shows as the value configured in these files currently. As this is just a raw email, it makes for a less than ideal inbox experience.

The new behavior means that a display name shows up in the sender field. For instance, GitHub emails are sent from a noreply address using the display name "GitHub" in the sender field/line. 

In this PR, we add a display name of "Codidact" as the QPixel default. Changes to the example and non example configurations can be made by those seeking to customize the display name if they wish to do so.


ArtOfCode-:
Most of our individual mailers use overridden default `from` addresses, so I suspect this won't actually affect the majority of emails - if we want a sender name on them, we'll need to modify the mailer files.

sau226:
I tried this on the dev server with code editing and it works for the basic devise injected routes (e.g. signup/account unlock via devise/password reset).

I'll make some changes that I'm not sure how best to test, but that should patch up those overrides for us to also work as well.

ArtOfCode-:
Don't know why checks aren't running here, but I think Rubocop is failing - can you run locally to check?

sau226:
> Don't know why checks aren't running here, but I think Rubocop is failing - can you run locally to check?

We don't run CircleCI on fork builds by the looks of it somehow, so this is expected behavior.

I triggered rubocop and it looks like a fix was auto generated.

The remaining issues are in the donations mailer and couldn't be auto fixed - I guess we need to double check the length of the reply to and from fields for those mailers. Would you have any suggestions for handling those?

luap42:
@sau226 could you copy the "other issues" error messages from Rubocop in here (minus any private infos about you of course)?

sau226:
Sure, the error looks something like the below:

```
Offenses:

app/mailers/donation_mailer.rb:7:121: C: Layout/LineLength: Line is too long. [122/120]
    mail from: 'Codidact Donations <donations-support@codidact.com>', reply_to: 'Codidact Support <support@codidact.com>',
                                                                                                                        ^^
app/mailers/donation_mailer.rb:17:121: C: Layout/LineLength: Line is too long. [122/120]
    mail from: 'Codidact Donations <donations-support@codidact.com>', reply_to: 'Codidact Support <support@codidact.com>',
```

Everything else relating to mailers should be patched by the auto fix tool, at least per my understanding

PR #737, opened 2021-12-06
Exclude post history with nil user

### LINKED ISSUE #736: It's possible to visit deleted post's history if you aren't logged in
https://software.codidact.com/posts/284206/history

not sure if it's visible to who is logged in also. But I could visit the history not post while not logged in. A similar problem  (https://meta.codidact.com/posts/284937) was solved recently, but I can't find that right now...

b-Istiak-s:
Maybe it fixes #736 , I got hint from [the commit](https://github.com/codidact/qpixel/commit/6373c10279f75b39f557399e2c21c6c9219141ea). both were completely relevant so I believe it might fix #736


ArtOfCode-:
A PostHistory event with a nil user is a different concept to a post with a nil user. This won't solve it, sorry!

PR #743, opened 2021-12-16
Add ability check to deleted post history

### LINKED ISSUE #736: It's possible to visit deleted post's history if you aren't logged in

sau226:
This does what it says in the PR title, to prevent people from accessing post history for deleted posts if they shouldn't be able to see the deleted post.

I reused code from https://github.com/codidact/qpixel/blob/2a627321dc65e92a81215e688218d4e56bb965b4/app/controllers/posts_controller.rb#L133-L135.

Is there any chance we could get small bits of common logic like that and set up a library/package or something to avoid duplication?

GitHub linking tag: resolves #736

### DIFF
--- a/app/controllers/post_history_controller.rb
+++ b/app/controllers/post_history_controller.rb
@@ -1,6 +1,11 @@
 class PostHistoryController < ApplicationController
   def post
     @post = Post.find(params[:id])
+    
+    if @post.deleted? && !current_user&.has_post_privilege?('flag_curate', @post)
+      return not_found
+    end
+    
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
As for common logic, I'm pretty sure there's already a helper method to see if a user can access a post.

PR #744, opened 2021-12-16
Remove duplicate keybind for flags

sau226:
This got picked up when I ran the code through a local SonarQube instance - if you ask me, it's a good example of how small bugs/"nits" can get picked up by automated services, demonstrates how having an automated service in a scan only/non blocking mode is useful for codebases such as ours and shows that while a code analyzer is not a must or a hard block (at least not in any hypothetical use by Codidact), it should not be overlooked too easily either.


ArtOfCode- (approved):

sau226:
Is there anything blocking this PR from merging since there is already 1 approval on it?

PR #760, opened 2022-01-14
moves dashboard link to header making it easier to access

Trilarion:
removed dashboard link from the communities drop down
added it to the header row after help (and before notifications)

makes it easier for people to know about and access the dashboard

see also this post by Alexei: https://meta.codidact.com/posts/285449

quite small change, but good training for me to get better into qpixel

not done (but could be done): remove mobile signin button from header row (is also on profile page)

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -45,6 +45,9 @@
         <%= link_to help_center_path, class: 'header--item' do %>
           <i class="fas fa-fw fa-question-circle"></i> Help
         <% end %>
+        <%= link_to dashboard_path, class: 'header--item' do %>
+          <i class="fas fa-fw fa-th" title='Dashboard'></i>
+        <% end %>
       <% end %>
       <% if user_signed_in? %>
         <a href="#" class="header--item inbox-toggle is-visible-on-mobile" data-header-slide="#js-inbox">
@@ -105,12 +108,6 @@
       </div>
     <% end %>
   </div>
-  <hr class="has-margin-top-2 has-margin-bottom-2" />
-  <div>
-    <%= link_to dashboard_path, class: 'button is-muted has-float-right' do %>
-      Dashboard
-    <% end %>
-  </div>
 </div>
 
 <div class="header-slide is-large" id="search-slide">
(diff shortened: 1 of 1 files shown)


Trilarion:
Wait, I haven't thought about mobile. There might not be enough space on the header row. Will put the link into the hamburger menu then.

cellio:
All the other stuff in the top bar has icons; do you have something for this?  (That could solve the mobile issue too, for sizes that are large enough for an icon but not the word.)

I don't have a suggestion for what the icon could be.  I'm hoping we're not the first project to need one and there are some conventions out there.

Trilarion:
Fixed. Dashboard link in header row on desktop and in right menu on mobile.

Trilarion:
@cellio I used this icon (https://fontawesome.com/v4.7/icon/th). I looks like https://imgur.com/gk9Pda5

I you have another in mind that can easily be changed.

cellio:
Works for me, thanks.  I didn't have something particular in mind, just a desire to have something.

app/views/layouts/_header.html.erb
```diff
@@ -44,6 +44,9 @@
       <% unless @community.is_fake %>
         <%= link_to help_center_path, class: 'header--item' do %>
           <i class="fas fa-fw fa-question-circle"></i> Help
+        <% end %>
+		<%= link_to dashboard_path, class: 'header--item' do %>
```
ArtOfCode-, line 48: Odd indentation on this line. Are you using hard tabs? Most editors have an option to use soft tabs (i.e. convert tabs to spaces when you hit tab, which is what we use)

Trilarion:
Yes, indentation was with tabs hardcoded. I used an editor that I do not normally use for developing and somehow it had tabs hardcoded. Usually I also use spaces instead and it should be fixed now.

app/views/layouts/_header.html.erb
Trilarion: Is now fixed

cellio:
@ArtOfCode- any other review feedback or is this ready now?

Trilarion:
As a continuation, on mobile, we could decide not to show the rep next to the user avatar in the top bar. That would free some space and gives the opportunity to put another icon or two the like the dashboard and maybe search. After all rep is not really a functionality of the site, just a stat. On the other hand, there is already the communities dropdown and the dashboard is very similar. That is more a design decision. Just wrote it down here in order not to forget it.

cellio:
I think we need to establish a "priority order" for things to show; it's not just mobile vs not but phones vs tablets vs small desktop windows vs large windows... We should decide what the most important things are to show, in order, and at any given size, we show as many as fit.  For now, in a small window I think I'd put the dashboard on the hamburger menu, but the larger question deserves a broader discussion, maybe on the dev Discord server.

Issue #759, opened 2022-01-14
Docker compose setup gives: No community record matching Host='localhost:3000'


RobertFloor:
**Describe the bug**
First congratulations on this nice project. I wanted to run it locally so I followed using docker-compose procedure in the docker/README.md file to get qpixel running locally. However, I end up with this in the browser: ```No community record matching Host='localhost:3000'.``` Probably I am missing a setting but I cannot find the correct place to fix the communities in the docker or variables file.

**To Reproduce**
1 Start with the develop branch and run these commands 
```
$ chmod +x docker/local-setup.sh
$ docker/local-setup.sh
$ docker compose build
$ docker compose up 
```
2. In either safari or chrome go to http://localhost:3000 the website it now says 
```
No community record matching Host='localhost:3000'
```

**Expected behavior**
I expect the webserver to start and show the webpage as described after step 4 in the readme. 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mac OS 12.1
 - Browser Safari 15.2

**Logs**

```
qpixel-uwsgi-1  | => Booting Puma
qpixel-uwsgi-1  | => Rails 5.2.6 application starting in development
qpixel-uwsgi-1  | => Run `rails server -h` for more startup options
qpixel-uwsgi-1  | Puma starting in single mode...
qpixel-uwsgi-1  | * Version 4.3.10 (ruby 2.6.6-p146), codename: Mysterious Traveller
qpixel-uwsgi-1  | * Min threads: 0, max threads: 16
qpixel-uwsgi-1  | * Environment: development
qpixel-uwsgi-1  | * Listening on tcp://0.0.0.0:3000
qpixel-uwsgi-1  | Use Ctrl-C to stop
qpixel-uwsgi-1  | Started GET "/" for 172.18.0.1 at 2022-01-14 15:26:23 +0000
qpixel-uwsgi-1  |    (1.1ms)  SET NAMES utf8mb4 COLLATE utf8mb4_unicode_ci,  @@SESSION.sql_mode = CONCAT(CONCAT(@@sql_mode, ',STRICT_ALL_TABLES'), ',NO_AUTO_VALUE_ON_ZERO'),  @@SESSION.sql_auto_is_null = 0, @@SESSION.wait_timeout = 2147483
qpixel-db-1     | mbind: Operation not permitted
[...]

PR #762, opened 2022-01-17
fixed error in docker compose setup

RobertFloor:
This pull request fixes https://github.com/codidact/qpixel/issues/759, the docker compose setup was not working. It ended up not finding the correct community. 

I found three improvements:

1 Add this capability to the mysql container to solve the` mbind operation not permitted `error (https://docs.docker.com/config/containers/resource_constraints/)
cap_add:
      - SYS_NICE
      
2 Included the` local dev port` in the container as an environment variable

3 Rebuild the cache after creating the community

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,11 +4,13 @@ services:
     build:
       context: "."
       dockerfile: docker/Dockerfile.db
-    volumes:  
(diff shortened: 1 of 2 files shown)


sau226 (approved):
LGTM

Issue #146, opened 2020-08-25
Let's limit or truncate super-long user names


cellio:
Our tester has produced the following:

Comment: 

![comment w/10-line user name](https://user-images.githubusercontent.com/5557942/91112578-7b15ff00-e651-11ea-841a-05d168dcc23d.png)

Post: 

![long user name prevents usercard formatting on post](https://user-images.githubusercontent.com/5557942/91112596-89641b00-e651-11ea-9255-4439d6b30918.png)

And the user page; note the size of the horizontal scrollbar:

![very wide page](https://user-images.githubusercontent.com/5557942/91112717-de079600-e651-11ea-99af-1beee216539f.png)

We should probably prevent that from rendering like that.  I think we do have a reasonable-ish limit on the length of user names, but special character encodings might be in play here.  I'm not sure what limitations make sense; maybe we can instead truncate where necessary?  Is that possible?  Can we, in the UI, give an element a bounding box and say "you must fit within this space", so the rest of the page displays ok?

chindraba-work:
Rather than try to limit the visual length of a user name, which seems problematic to implement without "knowing" all the possibilities for extending the visual withing a constrained character limit, just place the display of user names in a constrained box with overflow hidden. Reduces the effect of the occasional use of Zalgo text as well.

kevinmmansour:
Since we are in 2021, so we have to add limit for long usernames or add container for them but I like to limit them?

I think about 15 Characters is fine or maybe 20 Characters .

cellio:
Here is another place where this causes problems:

![users list, long name in left column pushes right out of the frame](https://user-images.githubusercontent.com/5557942/149264822-547ce58e-ffc6-4cb2-8fa8-764bea2ba9de.png)

A long user name in the left-most column pushes the other two columns over.  But we're not falling back to two columns if three don't fit; instead, the third (right-most) column is wider than its container and so the names are truncated.

I don't think we should try to limit names by visual display; with Unicode that'd be hard.  I think we should limit how much space we give a user name in all the places we show it, truncating the name if necessary.

Trilarion:
The user name in the screenshot q4q0... is 36 characters long. Not sure that the current limit on the length of user names is and what a reasonable length limit would be. I can probably check SEDE and see how long the smallest 95% of usernames on SO are as a reasonable value.

[...]

PR #769, opened 2022-02-08
prevent overflowing of user cards for long user names by wrapping them

Trilarion:
My solution to #146 as discussed there

- no truncation of user names (not clear where to set the cut and full user name read out by public/assets/community/codegolf.js might have unintended side consequences)
- biggest problem is overflow of layout, especially on the users list
- wrapping is a good solution for now (may be replaced by more sophisticated one in the future)
- however, wrapping must occur at different widths, depending whether the user card is shown on the users list (quite small) or below a post (relatively large)
- fixed widths are okay, because the whole layout is based on fixed values, values chosen here have been carefully optimized, they are the max values of what is possible within the solution
- changed CSS (user-card--link), templates common_card (additional parameter) and user (to indicate this is in the users list and a smaller max width is desired)

Might not fully solve the issue but it's not a simple problem. User names with 50 characters max length and potentially wide characters will take a lot of space. Most user names on StackOverflow (99.997%) are less than 30 characters long. I asked if there are [user names with more than 40 characters on Codidact](https://collab.codidact.org/posts/285625) but did not get an answer. I think we could safely reduce user name lengths to 40 characters though to alleviate the display problem (where is the sense in allowing longer names if we do not display them anyway).

### DIFF
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -2,8 +2,13 @@
 
 .user-list {
   display: grid;
-  grid-template-columns: repeat(2, 1fr);
   gap: 1em;
+
+  grid-template-columns: auto;
+
+  @media screen and (min-width: $screen-md) {
+    grid-template-columns: repeat(2, 1fr);
+  }
 
   @media screen and (min-width: $screen-lg) {
     grid-template-columns: repeat(3, 1fr);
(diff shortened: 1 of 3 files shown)


Trilarion:
Oh wait, I didn't think about mobile. It will need to be adapted for it. The users list on mobile is two columns currently, but should only be one column. On mobile the user name should wrap but the width of the link should be all that is available.

PR #771, opened 2022-02-11
Fix matomo tracking

sau226:
This PR goes ahead and updates matomo script implementations to more closely mirror the default.

Note that this resets a window function, tosses some of the async wrapping and sets a defer bit on the matomo side. Hopefully this ensures that the tracker will work going forward.

I haven't tested it, but everything is heavily based off what matomo itself provides, so there shouldn't be any significant code issues.


sau226:
@ArtOfCode- Can you please take a look at this? 

I've tried to delete some of the async wrappers to allow the code to handle an effective defer request, but testing this out after Matomo is upgraded again (https://matomo.org/faq/on-premise/update-matomo) would greatly help, to see if any of the "stability fixes" offered in the newer versions help solve the matter organically.

PR #775, opened 2022-02-17
[WIP] Fix problems with db:seed

Trilarion:
Work in progress

- Post type question could not be created (no answer_type_id possible if answer as post type wasn't created before). Solution is to create Answer first and then the other post types using Answer
- Other problems with db:seed will be fixed later (one with User system and one with a post)


Taeir:
@cellio The indended fixes in this PR have been fixed in #822 . I believe this can be cosed.

PR #788, opened 2022-04-11
Allow staff to edit and edit mod tags

sau226:
A while back, I got a staff flag on a QPixel site as a ordinary user (the default participate/participate everywhere ability only). 

Being someone who occasionally helps with mod tagging things on said site, this works out OK until you realize that staff with default abilities can not edit mod tags (they can suggest changes to include mod tags, and when they get edit ability they can no longer change mod tags). Obviously, staff who concurrently hold a moderator flag or another flag including moderator powers do not have this issue.

This PR does the following (keep in mind the changes affect non moderator/admin staff users only - moderator/admin staff users are not affected):
* Changes the post editing logic to give non moderator/admin staff edit access on all content, with the exception of suggested edits (if there is a suggested edit, editing is blocked, just like for normal users and they can not review the suggested edit).
* Changes the moderator tag validation logic so that staff with the edit posts ability can add or remove them, even if they are not a moderator/admin.

Because this PR changes our usual paradigm regarding the staff role (that the staff role is at a technical level just a label to show on an account/profile), I'm happy to discuss changes to this PR directly, such as removing the staff-based edit access change. 

More advanced changes like gating all or part of this logic behind a site setting are acknowledged, but due to the many sites such a setting would need to be updated on, this PR won't address it (please just merge it or defer until you have your own changes to implement the advanced change, then merge this PR and apply your changes on top before deploying).


cellio:
We have a page under Admin Tools for Global Site Settings.  I don't see any settings there (is that correct, that we don't have any yet)?  This seems like something one might want to do on a network-wide basis.

What I'd like to see: global (network) setting for things Staff status grants (multi-selectable).  Initial options are: edit mod tags on posts and see deleted content (the things covered by this request).  I'm suggesting we make this extensible because we, or others running their own networks, might want more options in the future.  I can see an active network wanting staff to be able to revoke abilities to mitigate problems happening right now, for instance.  We don't need to build any extra stuff, just have a place to put it.

sau226:
I'm marking this PR as deferred and in draft status until the (much bigger) global/network settings are introduced to enable this code to be implemented.

Is there any way we can add a deferred label to this PR, so we know to revisit it at some point?

Issue #272, opened 2020-11-06
Link back to mod dashboard from individual mod pages


cellio:
https://meta.codidact.com/questions/278995

> Right now the moderator dashboard has a bunch of links to different moderator tools like Flag Queue and Featured Links. However from those tools, there is not a link back to the dashboard, you either have to click the link in the sidebar or go back.
>
> Could we get a link to /mod at the top of the different moderator tool pages?

Seems like this should be pretty easy and I agree it would be helpful.

cellio:
Ok, most pages do have it in the right column; I just happened to pick the one that doesn't (featured posts) when looking at the report.  Lowering priority.

luap42:
Moderator pages could probably need some "unification in theme", in which this issue could be resolved too.

cellio:
I think we want to put the following on each of the nine pages under app/views:

```
<%= link_to moderator_path, class: "has-font-size-small" do %>
  All Moderator Tools
<% end %>
```

(I pattern-matched that from app/views/admin/index.html.erb)

Some of the mod tools are under app/views/moderator; some (like flags) are elsewhere.

I am a Ruby newbie.  Rather than cut/pasting that nine times, should I do something with includes or inheritance or something? 

We could also make the analogous change for the 13 pages under "admin", but there are way fewer admins than mods, so that'd be lower priority.

PR #803, opened 2022-05-30
Cellio/272 mod tools backlink

cellio:
Add a "return to mod tools" link to the mod-tool pages, based on the link that already existed on the flag-queue page.  Two noteworthy cases:

1. I added it to the Reports page along with the rest, but that page produces a 500 error in my dev environment (with or without my change) so I can't test it.  The error seems to be about time zones.  (Of course.  It's always time zones.  Let's abolish them.)
2. The mod-tools page includes a link to create a new help topic.  This goes to the regular "create post" page.  I couldn't quite figure out how to check there for the help-specific case, so I didn't change anything there.  It's probably a rare mod action, so I'm proceeding with the rest anyway.

Fixes https://github.com/codidact/qpixel/issues/272.

### DIFF
--- a/app/views/close_reasons/index.html.erb
+++ b/app/views/close_reasons/index.html.erb
@@ -1,4 +1,7 @@
 <% content_for :title, "Close Reasons" %>
+<%= link_to moderator_path, class: 'has-font-size-small' do %>
+  &laquo; Return to moderator tools
+<% end %>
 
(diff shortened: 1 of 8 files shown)


cellio:
I don't know why that first commit is showing up here.  That's a separate PR: https://github.com/codidact/qpixel/pull/798.

PR #806, opened 2022-06-11
Upgrade ruby to 2.7.6

sau226:
We are currently running ruby `2.7.6` on our dev server. Currently, the auto-deploy job in CircleCI is failing because a ruby version mismatch error is thrown after git operations occur, as this repo expects ruby `2.6.6` to be running.

As dev has been moving along just fine with this bump, this PR goes ahead and upgrades ruby from `2.6.6` to `2.7.6`. An upgrade of the ruby version on prod will need to occur before this deploys.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   test:
     docker:
-      - image: circleci/ruby:2.6.6-node
+      - image: circleci/ruby:2.7.6-node
       - image: circleci/mysql:8.0.20
         command: [--default-authentication-plugin=mysql_native_password]
         environment:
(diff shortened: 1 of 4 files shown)


Taeir:
> @Taeir What Ruby version do you need to upgrade Rails most effectively? I have this PR open to upgrade to Ruby `2.7.6`, but I'm happy to close this/change the version to upgrade to if that fits better in your Rails upgrade process.
> 
> Only real concern is the potential for breaking changes (assuming there are any) as part of a Ruby major upgrade, but the rest should be comparatively straightforward.

Rails 5.2 works on Ruby 2.6 but will trigger warnings in 2.7. Rails 6 and 6.1 support both 2.6 and 2.7. So for the rails upgrade it does not matter, and we can keep the PRs separate :)

Issue #809, opened 2022-06-21
"Push to network" checkbox is shown to mods when editing help (but only admins can do this)


cellio:
When editing a help topic, there's a checkbox for "push to network" to propagate a change.  Only network admins can actually do this, but a moderator informed me that moderators can see it too.  Testing on a dev server confirmed that a moderator checking that box doesn't actually propagate the change, but it would be better for the UI to not offer it in the first place, just like we don't show admin tools to non-admins.

Can we put an admin check around the display of that part of the help editor?

MoshiKoi:
Has this been fixed? The PR that says it fixes it has been merged, but this is still open.

PR #810, opened 2022-06-22
show 'push to network' control only to admins

cellio:
A moderator was surprised and I think a little alarmed to see an option, when editing help, to push the change to the whole network.  It wouldn't have worked, but this change adds an admin check before displaying that control.

Moderator sees:
mod

Admin (still) sees:
admin

### DIFF
--- a/app/views/posts/_form.html.erb
+++ b/app/views/posts/_form.html.erb
@@ -154,7 +154,8 @@
     </div>
 
     <% if edit_comment %>
-      <div class="form-group">
+    <% if current_user.is_admin %>
+    <div class="form-group">
         <div class="checkbox-setting">
           <div class="checkbox-setting--desc">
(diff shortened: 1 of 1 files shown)


cellio:
Fixes https://github.com/codidact/qpixel/issues/809 .

app/views/posts/_form.html.erb
```diff
@@ -154,7 +154,8 @@
     </div>
 
     <% if edit_comment %>
-      <div class="form-group">
+    <% if current_user.is_admin %>
```
ArtOfCode-, line 157: Probably better to use the [`admin?` helper](https://github.com/codidact/qpixel/blob/648a0fc85f494ab2a4700082b55a9a6c93275cc8/app/helpers/application_helper.rb#L9) here.

ArtOfCode- (changes requested):

PR #822, opened 2022-08-10
Fix seeding

### LINKED ISSUE #638: Fix category/post-type seeding
> **luap42 🍭**: @ArtOfCode- that is sonething for you as I don't know how to properly fix it: the rep per post type and category update broke category seeding as that depends on the old data model.
>
> <sub>&mdash; [Discord](https://discord.com/channels/634104110131445811/752215864354537524/878335134661746698)</sub>

Needs fixing, here's a reminder for me to do it.

Taeir:
Previously, seeding the database runs into multiple issues.

1. It does not seed models in the correct order, so some don't get added because their dependencies were not seeded yet.
2. It does not create a "Question" post type because it fails validations. "Question" post type sets `has_answer` to true, and needs to refer to the "Answer" post type, so we need to first create the answer post type before we can create the "Question" post type.
3. Creating a user succeeds, but an error is reported.
4. The initially seeded categories do not have tag sets associated with them.

For the first, I have added a bit of code to prioritize models on which others depend.

For the second, I have flipped the order of the Post types in the file and added ids. The import scripts (and potentially some code) rely on the fact that id for the "Question" post type is 1 and that the id for the "Answer" post type is 2, therefore I added the ids to ensure they are set that way.

For the third point, I added a confirmed at date for the admin user to ensure that devise does not try to send a confirmation mail. Otherwise it crashes in `devise/mailer/confirmation_instructions.html.erb` on line 11 since there is no `RequestContext` while seeding.

For the last point, I added `.id` such that it actually gets the ids and not the instances.

Fixes #638

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,9 +8,20 @@ else
   find_glob = 'db/seeds/**/*.yml'
 end
 
-Dir.glob(Rails.root.join(find_glob)).each do |f|
+# Get all seed files and determine their model types
+files = Dir.glob(Rails.root.join(find_glob))
+types = files.map do |f|
   basename = Pathname.new(f).relative_path_from(Pathname.new(Rails.root.join('db/seeds'))).to_s
-  type = basename.gsub('.yml', '').singularize.classify.constantize
+  basename.gsub('.yml', '').singularize.classify.constantize
+end
+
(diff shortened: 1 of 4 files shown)


Taeir:
I noticed another problem, tagsets were not getting linked to the initially seeded categories. I've added the fix for that too.

Issue #833, opened 2022-08-14
Better searching


Taeir:
**Is your feature request related to a problem? Please describe.**
Currently, search is performed using a database match query. While this is a very fast way of searching, it does not provide great results. As it is currently set up, only exact word matches are found. This can be good or bad:

- Bad: Searching for 'test' does not find a post with the word 'testing'
- Good: Searching for 'hell' does not find a post with the word 'hello'

Additionally, it seems the current solution does not search in titles, even though those usually contain the most important information.
- Bad: Search does not search in the title (#387)

When a user is not able to find their question, they will look elsewhere or ask a new one. This may decrease user interaction with the platform and increase the likelihood of negative interactions (closed because of duplicate). It should not be underestimated how important proper search is.

**Describe the solution you'd like**
Searching effectively is a rather complex problem. Rather than reinvent the wheel, I suggest to use an existing full-text analyzing solution. These search systems have an understanding of languages and can detect word matches when other tenses or even synonyms are used. Additionally, weights can be set to give more weight to matches in the title, etc.

**Describe alternatives you've considered**
There are a few different alternatives available:

**Elasticsearch**
Pros:
- Widely used
- Free to use
- Good ruby and rails integration
- Easy to integrate with current logic

Mehs:
- Partially open source (was fully open source, but seems to be moving away from that model slowly)

Cons:
- Hard to configure correctly when working at a large scale

**Apache Solr/Lucene**
Pros:
- Widely used
- Free to use
- Properly open source
- Good ruby and rails integration

Cons:
- Requires more work to integrate with current filtering logic (`tag:...`, `votes:10`, etc.)

There may be other systems out there, but these two seem to be the most used which have proper rails integration.

**Additional context**
While SolR may be a somewhat better choice for the long run, the current code style lends itself to elasticsearch much better (as far as I can tell). I'm relatively familiar with ElasticSearch, so I will be providing a pull request with an implementation for it. More work will be required to build additional features such as highlighting (indicate the words in the match), search in associated entities (increase search rank of posts where the answers match highly with the search term), and if you want to go that direction another consideration should be made which of the two to use.

[...]

PR #834, opened 2022-08-14
Better searching - Elasticsearch support

Taeir:
This PR adds (opt in) support for Elasticsearch.

To enable, the Global Site Setting `ElasticsearchEnabled` must be set to true.

Items left to do:
- [ ] Update documentation on how to configure Elasticsearch (build index, resync/import data). (perhaps make rails tasks for that?)
- [ ] Add support to other models/searches? (less necessary)
- [ ] Add support for associated items? I.e. comments or answers.

Some design questions:
- Elasticsearch can be enabled and disabled without server restart. However, when enabling it, an elasticsearch index resync is necessary to ensure that it is not out of sync with the database (otherwise you get error 500 pages for posts which were created while it was inactive, and searches will be working with outdated posts). In that sense, it does not need to support dynamic changing of the setting without restarts, and the code could perhaps be simplified (no mocking approach, just don't load ES at all if disabled). 

See #833


ArtOfCode- (commented):
All looks good to me so far.

Issue #826, opened 2022-08-13
Self-hosted instances


Taeir:
**Is your feature request related to a problem? Please describe.**
When attempting to set up a self-hosted instance of QPixel for the Delft Univeristy of Technology, a number of issues popped up.

The key ones:
* Some things which seem configurable in the application actually do not work when configured in a way that is different from how things are hosted on codidact.com.
* A lot of values are hardcoded to point to codidact.com (links to subpages, email sent from addresses, links in email addresses, etc.).

For the latter point, of course we would expect some mention and attribution of Codidact ("hey we made this" "come help us"), but for many other things we would expect these to be configurable instead.

**Describe the solution you'd like**
The main question is how codidact sees the future of QPixel. We have made/are making the required code changes to make it work as a self-hosted instance under a domain of our own. We could keep these changes separate, but this will provide difficulties in keeping up to date with changes made by the codidact team, and the changes we are making may also be useful to other users.

On the other hand, adding the required configuration requires additional work. For now, we just replaced a lot of the hardcoded values in the code. Of course, other institutions can do the same, and we could argue that this is the "intended" way to create a self-hosted instance. This does however raise the bar for difficulty on setting up such an instance, and may lead institutions to other platforms.

cellio:
See also https://github.com/codidact/qpixel/issues/770 and https://github.com/codidact/qpixel/issues/747 .  It wasn't our intention to force other instances to link to us all over the place; it kind of grew like that and we haven't fixed it yet.  I would rather have this be in configuration, not code forks.

Since you've just gone through and changed them, can you share your list for where you found hard-coded assumptions?  The ones I know about are:

- page footers: Codidact project links, and "other Codidact communities" (you probably want the name of your network in that string, not ours)
- links in seeded help topics
- email addresses and links, including the link in the confirmation email for new accounts
- the "Codidact" block in the right column and its outbound links
- donate button

Thanks.

Taeir:
Sorry, I didn't find #770 when looking for this problem in the issues list. I think that is pretty much it indeed. This is a full table of the items and their locations:

| Page / component                          | File                                                       | What                                                                                | Variable suggestion                                      | Importance                                                  |
|-------------------------------------------|------------------------------------------------------------|-------------------------------------------------------------------------------------|----------------------------------------------------------|-------------------------------------------------------------|
[...]

PR #840, opened 2022-08-18
Make tour more page configurable per instance

Taeir:
The tour's "more" page contains a long explanation of codidact, communities and how meta works. For instances not hosted as part of Codidact, this does not make sense.

This PR does the following:
- Add (global) site settings to control the content of the tour more page
- Update indentation of tour pages to match with other pages in the codebase.
- Add missing codidactyl.png asset (very important!)

Note: This requires seeding the database (or at least the site settings part of it) again. Since it skips things that are already there, that should prove no issue.

Related to #826 (fixes part of)


cellio:
Sounds like a great idea, thanks!  One question for review (@ArtOfCode- ): is it ok to add the Codidactyl image in the repo or does it get special licensing, and if we can't add it, what should we do for the tour?  (My unofficial opinion is that it's ok to add the Codidactyl's "personality" into the repo; this isn't the blue logo that we might want to protect more, and the tour needs a "host" of some sort.  But I defer to people who know more about this stuff than I do.)

Art, normally I'd ask @luap42 to review this (as the author of the tour) but given availability constraints, I'm asking you to do it.

ArtOfCode-:
I see no issues with including the Codidactyl in the repo, that should be fine.

Taeir:
I updated some of the description strings of the site settings just now and force pushed that (to not clutter up the history).

PR #841, opened 2022-08-18
Add tagset for tour

### LINKED ISSUE #838: Tour tagset not created/initialized
**Describe the bug**
When doing the tour, you are asked to add tags. However, it is not possible to add these tags (at least on meta.codidact.com).

**To Reproduce**
1. Create a new account such that you get the tour notification or go to https://meta.codidact.com/tour/
2. Follow the tour until you need to add tags.
3. Observe that you cannot add tags (the results could not be loaded).

**Expected behavior**
The user is able to add the tags as suggested in the tour.

**Screenshots**
image

**Additional context**
The tour looks for a tagset with id -1. This is not created in seeding, and does not seem to be present on all or some of the codidact sites. Creating an empty one will fix the issue.

Taeir:
Fixes #838

### DIFF
--- a/db/seeds/tag_sets.yml
+++ b/db/seeds/tag_sets.yml
@@ -1,2 +1,4 @@
 - name: Main
 - name: Meta
+- name: Tour
+  id: -1


cellio:
One-line seed change looks fine to me (but I haven't tested).

PR #842, opened 2022-08-18
Use same badge as elsewhere on the site

### LINKED ISSUE #816: Having two different symbols for moderators is confusing
https://meta.codidact.com/posts/286545

The symbol we show next to a user name for moderators is configurable (any Unicode symbol).  It's not the same as the icon on the user profile in the right column, above the abilities list; that one is baked into the code (I believe).  Can we use the same symbol for mods everywhere?

Taeir:
* Replaces the mod character with a fontawesome icon.
* Also display admin separate from Moderator on the user profile (was already done for the name).

Example with `shield-alt`:
Profile:
image

Post:
image

Fixes #816


Taeir:
Marking as draft until we conclude on what we actually want :)

ArtOfCode- (approved):

PR #843, opened 2022-08-19
Add SAML support

### LINKED ISSUE #825: SAML Single Sign On support
Support for SAML sign-in solutions such that self-hosted instances of QPixel can use alternative sign in methods. For example, larger institutions often have their own single sign on service. Many universities for example are connected to the [SurfConext network](https://www.surf.nl/en/surfconext-global-access-with-1-set-of-credentials).

Alternatively to SAML, Active Directory is also a commonly used sign-in system by institutions. Adding support for this would require a similar setup as for SAML and would raise the same questions. It does seem to be slightly less popular (though I have not done extensive research).

**Additional context**
It is not technically difficult to add support for SAML into QPixel. The `devise_saml_authenticatable` gem integrates  fully with the already used devise sign-in framework. A minimal amount of additional code and configuration will make actual sign ins/sign ups work.

The main work lies in the required configuration options (in the interface? in a file?), how to present the additional sign in methods to the user and the details of interaction with existing logic. Some examples:

* Can a site opt out of normal sign-in/sign-up (i.e. be SAML only)?
* Can a site use both normal login and SAML login together?
* Can users signed in through SAML change their email address (which could break the SAML user link)?
* Can users signed in through SAML set a password to log in without SAML?

I have already implemented required changes to support the minimal viable product of signing in using SAML (multi-login support). I can also create a tutorial on how to set this up for an instance if you are interested. I may contribute some code if I can make it integrated enough to add without breaking anything when SAML is not used. I'm mainly looking forward to hearing your thoughts and ideas on where you would like to see from such an integration.

Taeir:
This PR adds SAML support (Single Sign On providers).

### UI
* The SSO Sign In button either replaces the Sign Up + Sign In buttons, or is added in conjunction in case MixedSignIn is enabled.
* Add a notice for the sign-in and registration page to sign in with SSO (SSO + MixedSignIn only)

### Design questions/decisions
* This PR adds the SsoProfile model which stores the information from SAML in the case it exists. At the moment this is a single field, so it could also be added to the user model instead. However, for complex scenarios additional fields may need to be added, where it would make sense to have these separated into their own model.
* I expose the first_factor class variable from SessionsController such that I can access it from the SamlSessionsController. Ideally I would let the SamlSessionsController extend the SessionsController, but unfortunately ruby does not have multiple inheritance.
* If a user has an SSO Profile, they can **only** sign in through SSO. By default they cannot change this, but there is a site setting to allow users to disconnect from SSO. If they disconnect (sso profile is destroyed), they are sent a password reset email and they can sign-in normally once they set a new password (Mixed sign in only).
* Stopping impersonation (developer only feature) requires entering a password. If the user comes from SSO, they are signed out instead and redirected to the SSO sign in page (if we are lucky they are still signed in and it just works, otherwise they basically have to type their password again just like previously).

### 2FA
By default, SSO users will not be asked for 2FA (even if it is enabled on their account or they are global admin and the setting to enforce it is enabled), since it is the job of the SSO provider to ensure safety (it can do 2FA). There is a setting to change this behavior to allow SSO users to set 2FA (and this will also start enforcing it for global admins if that setting is enabled).

### Routes
* This adds devise routes under `/users/saml/...`. These routes are only active if the site setting `SsoSignIn` is enabled (they dynamically become active or inactive). This means that no additional endpoints (security risks) are present whenever SsoSignIn is disabled.
* This adds routes under `/users/disconnect-sso` which can be used for SSO users to disconnect their account from the SSO service.

### Database migration
* The database migration adds the `sso_profiles` table `(saml_identifier)`, but does not alter any existing tables. No existing tables or data is modified.

### Configuration
* Six new global site settings are added:
  * SsoSignIn -> (default false) controls whether sign in with SSO is enabled
  * MixedSignIn -> (default false) if sso is enabled, normal sign in becomes unavailable unless if mixed sign in is enabled. (ideally we would have a setting where you can choose between 3 options (Normal, SSO, Mixed), but this is not currently possible).
  * Enable2FAForSsoUsers -> (default false) whether SSO users are allowed to use 2FA.
  * EnableMobileSignInForSsoUsers -> (default false) whether SSO users can use the mobile sign in option.
  * AllowSsoDisconnect -> (default false) whether SSO users can disconnect their account from SSO and switch it over to a normal account.
* The added configuration should suit both basic and more complex scenarios. The comments in the configuration already provide a reasonable explanation on how to configure it (`config/initializers/devise.rb` at the bottom and the `config/attribute-map.yml`).

### Documentation
Added as wiki page: https://github.com/codidact/qpixel/wiki/Setting-up-SAML---Single-Sign-On

### Tasks to do
* [x] Integration with 2FA
* [ ] Test (manually at least) whether 2FA + SSO Sign In works
* [x] Disable normal sign_in/register/confirm routes if SSO sign in is enabled and mixed sign in is not.
* [x] Add tutorial somewhere (docs repo? wiki?)
* [x] Figure out something for impersonation password request (sign out?)

Closes #825


Taeir:
@cellio Do you have a suggestion for where I can add the documentation for this? I seem to be able to add wiki pages, but the wiki currently says that there is no wiki and that all documentation is in the codidact/docs repo (which contains outdated information?). I can also add it as a markdown file, but I think it is nicer to keep the documentation more separate from the code base.

cellio:
@Taeir I see that the installation guide is in this repo (https://github.com/codidact/qpixel/blob/develop/INSTALLATION.md), so maybe we should put the SAML doc there too?  Pinging @ArtOfCode-  for possibly a better answer.  (This is my first OS project so I'm not as familiar with the norms as you and Art are.)

Another possibility (deferring to Art) would be to use this repo's wiki, replacing that landing page that says "go use the now-out-of-date docs wiki" with something current.

ArtOfCode-:
Wiki makes sense to me. We could also copy the readme/installation guide there as well.

cellio:
I updated the landing page on [this repo's wiki](https://github.com/codidact/qpixel/wiki).  I don't currently know how to grant write access and I'll be going offline soon until Saturday night; @ArtOfCode-  if you're around can you take care of that?  Thanks.

app/views/devise/shared/_links.html.erb
```diff
@@ -1,20 +1,25 @@
-<%- if controller_name != 'sessions' %>
-  <%= link_to "Sign in", new_session_path(resource_name) %><br />
+<%- if controller_name != 'sessions' && controller_name != 'saml_sessions' %>
+  <%- if SiteSetting['MixedSignIn'] || !SiteSetting['SsoSignIn'] %>
```
ArtOfCode-, line 2: suggestion: This `SiteSetting['MixedSignIn'] || !SiteSetting['SsoSignIn']` seems to be repeated a lot, both here and in other files. Worth creating `sso_sign_in_enabled?` and/or `devise_sign_in_enabled?` helper methods for?

config/attribute-map.yml
```diff
@@ -0,0 +1,18 @@
+# Add your SAML attribute mapping here.
```
ArtOfCode-, line 1: lint: Minor one - unless it needs to be kebab-case for some other reason, file names should generally be snake_case.

ArtOfCode- (commented):
Thanks! Love how well structured and documented this code is so far.

app/controllers/application_controller.rb
```diff
@@ -327,7 +327,11 @@ def authenticate_user!(_fav = nil, **_opts)
       respond_to do |format|
         format.html do
           flash[:error] = 'You need to sign in or sign up to continue.'
-          redirect_to new_user_session_path
+          if SiteSetting['MixedSignIn'] || !SiteSetting['SsoSignIn']
+            redirect_to new_user_session_path
+          else
+            redirect_to new_saml_user_session_path
+          end
```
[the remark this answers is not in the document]
Taeir: Correct.

app/views/devise/shared/_links.html.erb
Taeir: Good suggestion, I do wonder how to make the helper method work in the devise emails, since they don't descend from application mailer nor from application controller, and I'm not sure if they load all helpers by default.

config/attribute-map.yml
ArtOfCode-: Okay, not an issue then. Actually working trumps style concerns!

app/views/devise/shared/_links.html.erb
ArtOfCode-: Have a look in app/views/donations_mailer and admin_mailer - looks like some helpers might be supported out of the box, but if not you can use `ApplicationController.helpers.helper_method_name` to access them.

Taeir:
> * What kind of IDP system are you hoping to target SAML to? If its something like Shibboleth/Okta/OneLogin, we may as well as provide example configs/screenshots to make the setup a little more hassle free.

We use it with our institution's SAML system (Delft University of Technology), which is probably a simpleSamlPHP instance with custom settings. We also have a test SAML server with the same configuration but mock accounts which we use for the development of our applications to test SAML integrations. We are already using it in production, but this PR polishes up the code and makes it more configurable to other scenarios + provides instructions. I agree that it would be good to provide examples of linking with Okta or OneLogin, as those are probably a more common use case.

> * SAML IDPs can be just a passthrough to LDAP systems. Should we recommend to users that want to keep using LDAP to set up a SAML service instead and forego direct LDAP integration in QPixel?

Are there currently institutions you are aware of that use LDAP login with QPixel? My institution indeed has the SAML IDP as a passthrough to their LDAP directory. A direct link to LDAP in QPixel is possible as well, but less "trustworthy"/convenient since you don't get the organizational login screen, autofill credentials, additional 2FA options, and the only-sign-in-once experience. Adding LDAP login support to QPixel is possible as well, but more of a hassle due to the additional views required to make that work with mixed sign in (usually it applies the ldap login on the normal sign in page). I have not worked before with the devise_ldap_authenticatable gem, but I suspect it to be somewhat similar.

> * (my opinion only for this last one) It's a bad idea to "enforce" 2fa if SAML is the only "active" login method. If it uses a "native" or "mixed" (can log in "natively" or via SAML) mode, then fine, but the intent of an SSO system is to generally outsource all the "auth bits" as well. Feel free to gate this enforcement component behind a setting (probably default the "service side 2fa" to off) if this is something that you really want to keep.

With enforcement you mean for mod+? I think all other users have an opt in to use 2FA. If you use mixed sign in, I don't think it makes sense to disable 2FA site wide, and the question is whether this makes sense as a separate setting. Instead, we could:
* Don't enforce 2FA for mods if they are SAML users, require 2FA with SAML if users enabled it for their accounts.
* Disable 2FA for SAML sign ins and list this with the enforcement of 2FA settings. Normal 2FA will apply when signing in through a different method.

The second option would be easy to add, as it only requires removing some logic from the `saml_sessions_controller`.

Taeir:
@sau226 
> Re testing, we have a dev system we can probably make a backup of + push this PR to when ready. Just have some thoughts for you in the meantime:

I don't really understand what you mean by this. You mean a staging setup for testing purposes?

Taeir:
Added some new restrictions regarding 2FA:
* (By default) 2FA is not checked for SSO users even if enabled
* (By default) SSO users cannot enable 2FA in the UI
* (By default) 2FA is not enforced for global administrators when they sign in using SSO, even if this setting is enabled

There is a global site setting to configure whether SSO users can use 2FA. If this is set, if behaves like for normal sign in.

Taeir:
Added some new restrictions:
* SSO users cannot change their email or password (as these are provided by the SSO service)
* SSO users cannot sign in through normal authentication even when mixed authentication is allowed.
* SSO users can now disconnect their account from SSO if the server supports mixed sign in (if the global site setting for it is enabled, it is not by default)

Taeir:
@ArtOfCode- This is ready for re-review. I will still finish the tutorial on how to set it up with Okta, but the rest should be fine now.

ArtOfCode- (approved):

PR #846, opened 2022-08-23
Add missing help images

Taeir:
These images are linked in the help pages (and quite helpful), but are not available in the repo. It would be good if they are included such that they work on newly created instances (assuming they are in the public domain).


cellio:
@Taeir good catch, thanks.  Images in seeded help topics should be in the repo so other deployments get them too.  There's one I'd like to omit, pending the rest of this comment.

Question for @ArtOfCode- : I notice one of those images is the Codidact org chart, which made me go look at the help.  What would it take to pull the Codidact policy topics out of the seeded help and make sure they're just on _our_ network?  The seeds could contain stubs for "put your TOS here" and "put your 'about our network' topic here", so other instances can customize that.  We could rename "Codidact FAQ" to "About the Codidact Project" with just the first paragraph (slightly edited to not refer to our network), and drop the rest for other deployments.  What would it take to untangle this?  

I can prepare a PR for the modified topics if merging that wouldn't mess up our existing deployment.  And -- to get back to this current PR -- we could skip adding the Codidact team picture here.

ArtOfCode-:
@cellio that's the gist of it, just change the seeds. That _shouldn't_ affect our current deployment because it's already in place.

cellio:
@Taeir could you remove the one with the Codidact team org chart, which I don't think we want to include in every deployed version?  In https://github.com/codidact/qpixel/pull/849 I'm removing the use of it in the seeds.

Taeir:
@cellio I removed it.

PR #851, opened 2022-08-24
Fix CI

Taeir:
We were using legacy CI images, for which the updated ruby 2.7 image does not exist (any more?), and these are being phased out. We need to use the tag `cimg` instead to switch to the new images (which should be faster but do the same).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,15 +2,15 @@ version: 2.1
 jobs:
   test:
     docker:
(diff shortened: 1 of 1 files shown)


Taeir:
NOTE: The ci does not run rubocop or tests whenever code is not edited, which is how this issue could have happened unnoticed in the first place. Perhaps we should change it to also run whenever the circle ci config is updated.

PR #852, opened 2022-08-25
Search tags in name and excerpt

Taeir:
This PR adds searching in tag excerpts when choosing tags for a post. Creating a new tag instead of a matched tag is moved to the bottom to prevent accidental creation of duplicate tags, and it is indicated with a footer.

image

We added this feature because we found it useful for our instance, but please determine whether it fits in your vision for how tags are intended to be used.


cellio:
I like!

ArtOfCode- (approved):

PR #853, opened 2022-08-25
Display post body in lists

Taeir:
This PR adds a short preview of the body of a post to any list which displays posts (category index, search, user posts, etc.)

image

For this, the text size of the title was increased to give a better reading experience. The size matches the codesign standard for H3 tags. The content was set to slightly larger text of 14 px to avoid it getting too big.

The length of the preview is configurable as `PostBodyListTruncateLength` in the settings of each site.

### DIFF
--- a/app/assets/stylesheets/posts.scss
+++ b/app/assets/stylesheets/posts.scss
@@ -40,6 +40,14 @@ h1 .badge.is-tag.is-master-tag {
   font-size: 14px;
 }
 
+.post-list--title {
+  font-size: 19px;
+}
+
+.post-list--content {
+  font-size: 14px;
+}
+
 .post--action-dialog {
   display: none;
 
--- a/app/views/posts/_list.html.erb
+++ b/app/views/posts/_list.html.erb
@@ -35,6 +35,11 @@
       <%= post.duplicate_post && post.post_type.is_closeable && post.closed ? "[duplicate]" : "" %>
       <% end %>
     </div>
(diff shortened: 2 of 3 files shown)


cellio:
Short previews are handy in many contexts, yes.  (Search results seem especially helpful.)

I assume `PostBodyListTruncateLength=0` means no previews (current behavior)?

Also, would this preserve the "N answers" and "last activity" parts from the current view, or do you see this replacing those?  What about tags?

Taeir:
> I assume `PostBodyListTruncateLength=0` means no previews (current behavior)?

That is the intention, but that part is yet to be added 😉 

> Also, would this preserve the "N answers" and "last activity" parts from the current view, or do you see this replacing those? What about tags?

We have only added the body description, we don't remove any of the existing parts from the view. The reason you don't see tags in the screenshot is that all of them are answers. I took it from my profile page to prevent showing any content made by our students (those posts are in my dev database because we were testing our SO import there, and they are quite useful to check what things actually look like), and I very rarely post questions.

This is what it looks like for a post as shown in the Q&A category:
image

After an edit:
image

The last activity will also still show up if there was any activity on the item (i.e. an answer?), and I think it looks fine that way.

> Edit: oh, just noticed it's a draft; sorry for jumping the gun.

No problem, I think its good to get a "pre-review" to determine whether it is worthwhile to add the additional changes necessary.  Thanks!

Taeir:
Disable if set to 0 (or less than 0)

ArtOfCode- (approved):

PR #856, opened 2022-08-29
Only create the tour tagset for the first community

Taeir:
Fixes issue with seeding whenever there are multiple communities.


Taeir:
@sau226 This should fix the issues you encountered.

PR #857, opened 2022-08-30
Restricts `answers:n` to posts that have answers

### LINKED ISSUE #797: Can a search for "answers:n" limit itself to post types that can have answers?
https://meta.codidact.com/posts/286454/286456#answer-286456

Searching for `answers:0` returns results that _are_ answers (because they don't have answers).  That's not, strictly speaking, _wrong_, but it's a little confusing.

Can a search that includes the "answers" keyword restrict itself to searching post types that can have answers?  Currently that's just questions, but we shouldn't bake in "is a question" because there could be other top-level post types in the future that have answers.

If the search already specifies a post type (like question) then we should use it; in a future where there's another kind of answerable post, we wouldn't want to expand the search.  So I think the logic should be:

    if search includes `answers:`
        then if search includes `post-type:`
            then do that
        else add "post types that can have answers" to search
    else... (rest of search logic)

MoshiKoi:
Should fix #797

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -77,7 +77,9 @@ module SearchHelper
         next unless value.match?(valid_value[:numeric])
 
         operator, val = numeric_value_sql value
+        post_types_with_answers = PostType.where(has_answers: true)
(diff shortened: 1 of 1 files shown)


Taeir (approved):
Looks good!

PR #858, opened 2022-08-30
Filters out pinned posts from hot posts

### LINKED ISSUE #812: Don't show a post as both featured and a hot post
**Is your feature request related to a problem? Please describe.**
It is currently possible for the hot post logic to pull content that is also featured. This can result in awkward situations like that in https://meta.codidact.com/posts/277057, where a post shows up in both the hot posts and featured posts modules.

In https://github.com/codidact/qpixel/blob/develop/app/controllers/application_controller.rb#L252, adding code to filter out active pinned links that are being shown to users. 

Action can be taken to make more posts eligible for the hot post feature and users can then wait for the cache to expire, causing the featured post to no longer be a hot post.

MoshiKoi:
Fixes #812 

I'm not quite sure why pins with undefined post ids (i.e., featured links rather than featured posts) needed to be filtered as well, but it doesn't work if I don't.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -264,7 +264,7 @@ class ApplicationController < ActionController::Base
     # Post not pinned => NOT IN pinned_post_ids = NOT NULL = NULL => Not Selected
     # Post pinned => NOT IN pinned_post_ids = NOT TRUE = FALSE => Not Selected
     # I.e., if pinned_post_ids contains null, the selection will never return records
-    pinned_post_ids = @pinned_links.pluck(:post_id).select(&:present?)
+    pinned_post_ids = @pinned_links.map(&:post_id).reject(&:nil?)
   
     @hot_questions = Rails.cache.fetch('hot_questions', expires_in: 4.hours) do
       Rack::MiniProfiler.step 'hot_questions: cache miss' do
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -256,7 +256,14 @@ class ApplicationController < ActionController::Base
     end
 
     # We need to filter out pinned links that aren't posts
-    # because for whatever reason, where.not(id: nil) will return an empty query
+    # If pinned_post_ids contains nil, then the query formed is
+    # SELECT WHERE NOT IN (... NULL)
+    # In SQL, when the rhs of IN contains NULL, it returns NULL if not present, rather than false
+    # Secondly, NOT NULL evaluates to NULL
+    # These two counterintuitive behaviors mean that if NULL is in pinned_post_ids,
+    # Post not pinned => NOT IN pinned_post_ids = NOT NULL = NULL => Not Selected
+    # Post pinned => NOT IN pinned_post_ids = NOT TRUE = FALSE => Not Selected
+    # I.e., if pinned_post_ids contains null, the selection will never return records
     pinned_post_ids = @pinned_links.pluck(:post_id).select(&:present?)
   
     @hot_questions = Rails.cache.fetch('hot_questions', expires_in: 4.hours) do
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -254,12 +254,18 @@ def pull_pinned_links_and_hot_questions
...
+    # because for whatever reason, where.not(id: nil) will return an empty query
+    pinned_post_ids = @pinned_links.pluck(:post_id).select(&:present?)
+  
     @hot_questions = Rails.cache.fetch('hot_questions', expires_in: 4.hours) do
       Rack::MiniProfiler.step 'hot_questions: cache miss' do
         Post.undeleted.where(last_activity: (Rails.env.development? ? 365 : 7).days.ago..DateTime.now)
             .where(post_type_id: [Question.post_type_id, Article.post_type_id])
             .joins(:category).where(categories: { use_for_hot_posts: true })
             .where('score >= ?', SiteSetting['HotPostsScoreThreshold'])
+            .where.not(id: pinned_post_ids)
```
Taeir, line 268: I have not tested this, but to optimize the query it would be better if we do not pluck.

Also, your finding of pinned links which do not have a post id breaking the query is quite strange. Especially since "WHERE id IS NOT NULL" should not affect the query.

```suggestion
            .where.not(id: @pinned_links.select(:post_id))
```

Taeir (commented):
I left a comment about your change. Perhaps you could try it to see if it works?

app/controllers/application_controller.rb
MoshiKoi: Unfortunately, that doesn't work, and I really have no idea why.

The debugger shows

image

```
[#<PinnedLink id: nil, post_id: 28>, #<PinnedLink id: nil, post_id: nil>, #<PinnedLink id: nil, post_id: 21>]
```

(Of course, I didn't add the to_a in the code itself, it was just so I could see what came up)

Taeir: Thanks for checking. I had a look at `@pinned_posts` and it seems to be cached in redis, which may be what causes the interesting behavior. I think then your solution is the best one indeed.

MoshiKoi: So yes, filtering out the null values from the featured links is unfortunately necessary

cellio: Sigh. Could you drop a comment in there for the next person who comes across this code and thinks it can be simplified?

MoshiKoi: @cellio Done 👍

app/controllers/application_controller.rb
```diff
@@ -254,12 +254,25 @@ def pull_pinned_links_and_hot_questions
...
+    # We need to filter out pinned links that aren't posts
+    # If pinned_post_ids contains nil, then the query formed is
+    # SELECT WHERE NOT IN (... NULL)
+    # In SQL, when the rhs of IN contains NULL, it returns NULL if not present, rather than false
+    # Secondly, NOT NULL evaluates to NULL
+    # These two counterintuitive behaviors mean that if NULL is in pinned_post_ids,
+    # Post not pinned => NOT IN pinned_post_ids = NOT NULL = NULL => Not Selected
+    # Post pinned => NOT IN pinned_post_ids = NOT TRUE = FALSE => Not Selected
+    # I.e., if pinned_post_ids contains null, the selection will never return records
+    pinned_post_ids = @pinned_links.pluck(:post_id).select(&:present?)
```
ArtOfCode-, line 267: We should use `map` here instead of `pluck` - performance is usually better. I'd also (opinion alert) use `reject` rather than `select` - it narrates what you're trying to do a bit more clearly.

```suggestion
    pinned_post_ids = @pinned_links.map(&:post_id).reject(&:nil?)
```

ArtOfCode- (changes requested):

PR #859, opened 2022-08-30
Sticky navigation bar preference

### LINKED ISSUE #814: Option for sticky top bar
https://meta.codidact.com/posts/286303

Sticky top bars are controversial (not just here).  Some people love them and some people hate them.  Currently our top bar is not sticky; the linked post is a request for a user option to make it sticky.

I'm marking this as low priority because part of the utility (on other sites) is seeing new notifications right away -- but we don't have live notifications yet (you have to reload the page).  The links could still be useful, granted.

MoshiKoi:
Resolves #814 by adding a network preference to make the header navigation bar sticky.

I used inline styles because there doesn't seem to be anything in co-design for sticky items, but if there is then it should probably be changed to use that instead.

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,7 +1,11 @@
 <% logo_path = SiteSetting['SiteLogoPath'] %>
 <% mobile_logo_path = SiteSetting['MobileLogoPath'] %>
+<% sticky_header = user_preference('sticky_header', community: false) == 'true' %>
 
-<header class="header is-small has-margin-0 <%= SiteSetting['SiteHeaderIsDark'] ? 'is-dark' : '' %>">
+<header class="header is-small has-margin-0 <%= SiteSetting['SiteHeaderIsDark'] ? 'is-dark' : '' %>"
+    <% if sticky_header %>
+      style="position:sticky;top:0;"
+    <% end %>>
   <div class="container header--container">
     <div class="header--brand">
       <% if mobile_logo_path.present? %>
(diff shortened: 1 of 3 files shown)


Taeir (commented):
I'm happy to see how much CSS has improved since when I learned it. This would have been so annoying to get right 10 years ago :P 

I have left one suggestion to use a class instead of an inline style, since inline style does not seem to be used anywhere in the application.

Taeir:
For context, a sticky header stays fixed to the top and appears above the content (image from scrolling down the page a bit):

image

app/views/layouts/_header.html.erb
```diff
@@ -1,7 +1,11 @@
 <% logo_path = SiteSetting['SiteLogoPath'] %>
 <% mobile_logo_path = SiteSetting['MobileLogoPath'] %>
+<% sticky_header = user_preference('sticky_header', community: false) == 'true' %>
 
-<header class="header is-small has-margin-0 <%= SiteSetting['SiteHeaderIsDark'] ? 'is-dark' : '' %>">
+<header class="header is-small has-margin-0 <%= SiteSetting['SiteHeaderIsDark'] ? 'is-dark' : '' %>"
```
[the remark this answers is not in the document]
MoshiKoi: Done 👍

config/config/preferences.yml
```diff
@@ -67,3 +67,9 @@ auto_follow_comment_threads:
   description: >
     Automatically follow any comment thread you participate in.
   default: 'true'
+
+sticky_header:
+  type: boolean
+  description: >
+    Make the Codidact navigation bar sticky
```
ArtOfCode-, line 74: Can we change this description to "Make the top navigation bar sticky."? With all the work that's being done to remove Codidact-specific things, I want to make sure we're not adding more now.

ArtOfCode- (changes requested):

config/config/preferences.yml
MoshiKoi: Done 👍

PR #865, opened 2022-08-31
Create markdown-it plugin to escape LaTeX

### LINKED ISSUE #435: Can we prevent collisions between Markdown and Mathjax?
See https://github.com/codidact/docs/issues/26 and the meta posts linked there.  Some Mathjax apparently has to be escaped because it's within Markdown, which confuses people periodically.  We can document it (the linked issue is in the docs repo), but if we can make it behave the way people are expecting, that'd be better.

MoshiKoi:
Should fix #435. However, I haven't fully tested this yet, so feedback is welcome.

### DIFF
--- a/app/assets/javascripts/latex_escape.js
+++ b/app/assets/javascripts/latex_escape.js
@@ -40,11 +40,15 @@ latexEscape.inlineRule = ruleset => (state, silent) => {
 latexEscape.blockRule = ruleset => (state, startLine, endLine, silent) => {
   const start = state.bMarks[startLine] + state.tShift[startLine]
 
-  if (!state.src.startsWith(ruleset.open, start)) { return false; }
+  ruleset.match.lastIndex = start;
+  const match = ruleset.match.exec(state.src);
+  if (!match) { return false; }
 
-  let nextLine = startLine + 1;
+  const end = start + match[0].length
 
-  while (nextLine < endLine && !state.src.endsWith(ruleset.close, state.eMarks[nextLine])) {
+  let nextLine = startLine;
+
+  while (nextLine < endLine && state.eMarks[nextLine] != end) {
     ++nextLine;
   }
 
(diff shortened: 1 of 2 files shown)


MoshiKoi:
It's been mostly tested and seems to work fine, so I'm opening it for review.

MoshiKoi:
Okay, after fixing the most stupid mistake ever (return failure if validate is true), I have *actually* gotten all of my test posts rendering correctly.

app/assets/javascripts/latex_escape.js
```diff
@@ -0,0 +1,113 @@
...
+    // $$123$$ some text
+    //
+    // not math $$456$$
+    validate: match => match.indexOf('$$', 2) === match.length - 2.
+  },
+  '\\[': {
+    match: /\\\[[^]+?\\\]$/my   // \[...\]
+  },
+  'begin-end': {
+    match: /\\begin\{(.+)\}[^]+\\end\{\1\}$/my // \begin{...}...\end{...}
```
Taeir, line 111: Why is this one greedy in contrast to the others which are reluctant?

Taeir (commented):
I learned a lot of new regex things reviewing this PR 😂 
Reluctant matching and how `[^]` works in Javascript regex.

I'm a bit curious though why the "sticky" flag is set on the regexes and what the effect of that is on your regex. Does that ensure it doesn't match inside of another match?

app/assets/javascripts/latex_escape.js
MoshiKoi: No special reason, I probably just forgot. I'll change it

MoshiKoi:
@Taeir 
> I'm a bit curious though why the "sticky" flag (`/y`) is set on the regexes and what the effect of that is on your regex. Does that ensure it doesn't match inside of another match?

> The `y` flag indicates that the regex attempts to match the target string only from the index indicated by the [`lastIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex) property (and unlike a global regex, does not attempt to match from any later indexes).

([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky))

PR #867, opened 2022-09-01
Make tag selection clearable

### LINKED ISSUE #252: Can't remove a tag's parent
https://meta.codidact.com/questions/278607

Once a tag has been made a child of another tag, there appears to be no way to revert the child back to a freestanding (top-level) tag.  When editing such a tag there's this menu to choose a parent:

screenshot

The menu has a list of tags to select from.  But "none" or blank isn't an option.

In general we want actions, especially moderator actions, to be reversible.

Complexity: if it's just a matter of adding the option and, if it's chosen, removing a relation, this should be easy.  If there's more in the back end that depends on tag hierarchies, I wouldn't know how to evaluate that.  Hence, unassessed.

MoshiKoi:
Resolves #252

### DIFF
--- a/app/assets/javascripts/tags.js
+++ b/app/assets/javascripts/tags.js
@@ -66,7 +66,8 @@ $(() => {
           };
         },
       },
-      templateResult: template
+      templateResult: template,
+      allowClear: true
     });
   });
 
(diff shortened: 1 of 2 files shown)


Taeir (approved):
Looks good :)

cellio:
@ArtOfCode- this is small and Taeir has already reviewed it; adding you for final dispensation.

Issue #783, opened 2022-03-23
Comment count includes deleted comments (user confusion)


cellio:
https://meta.codidact.com/posts/285825

A thread link includes the number of comments in the thread.  This number includes all comments, but should not include deleted comments for people who can't see them.  The linked post includes screenshots where the thread said 4 comments, the user expanded the thread, and it shows one comment + "showing 4 of 4 comments".  Only on the full thread page do we show "skipping 3 deleted comments".

Ideally, we would show a number that reflects what *you* can see, and include those "skipped deleted" steps in the in-page view. For people who can see deleted comments (mods/admins), we would either show the true number or show something like "1 (+ 3 deleted)".

MoshiKoi:
Is this still an issue? #875 fixed this in favor of including the "skipping 3 deleted comments" message on the in-page view.

PR #869, opened 2022-09-01
Include deleted comments in inline thread

### LINKED ISSUE #696: For inline comment view, indicate where there are deleted comments
https://meta.codidact.com/posts/283807

When viewing a comment thread, if there are deleted comments you see something like "skipped 3 deleted comments".  However, if you view a thread in-page, you see the first five surviving comments without an indication that there are also deleted ones.  It would be better to show the indicator there too, especially if we're going to give people the option to expand the entire thread in-page (a change that's in progress on another branch).

MoshiKoi:
Resolves #696

While it shows deleted comments (as "Skipping n deleted" comments), it only counts undeleted comments, so if there are 1 deleted and 1 undeleted comment, it show "Showing 1 comment" at the bottom. Not sure whether it's better like this or not though.

If it's better to include the deleted comments in the count, then it would render #783 moot

### DIFF
--- a/app/views/comments/thread.html.erb
+++ b/app/views/comments/thread.html.erb
@@ -67,7 +67,7 @@
   <% if params[:inline] == 'true' %>
     <% count = 0 %>
     <% comments = comments.take_while do |comment| %>
-      <% count += 1 if !comment.deleted %>
+      <% count += 1 unless comment.deleted %>
       <% count <= 5 %>
     <% end %>
   <% end %>
--- a/app/views/comments/thread.html.erb
+++ b/app/views/comments/thread.html.erb
@@ -65,7 +65,7 @@
   <% skipped_deleted = 0 %>
   <% comments = @comment_thread.comments %>
   <% if params[:inline] == 'true' %>
-    <% comments = comments.where(deleted: false).take 5 %>
+    <% comments = comments.take 5 %>
   <% end %>
   <% comments.each do |comment| %>
     <% if comment.deleted && !(current_user&.is_moderator && params[:show_deleted_comments] == "1") %>
--- a/app/views/comments/thread.html.erb
+++ b/app/views/comments/thread.html.erb
@@ -65,7 +65,11 @@
   <% skipped_deleted = 0 %>
   <% comments = @comment_thread.comments %>
   <% if params[:inline] == 'true' %>
-    <% comments = comments.take 5 %>
+    <% count = 0 %>
+    <% comments = comments.take_while do |comment| %>
+      <% count += 1 if !comment.deleted %>
+      <% count <= 5 %>
+    <% end %>
   <% end %>
   <% comments.each do |comment| %>
     <% if comment.deleted && !(current_user&.is_moderator && params[:show_deleted_comments] == "1") %>


cellio:
I think the clearest thing is to show where there were deleted comments ("skipping 3 deleted" etc) and only include undeleted ones in the count (both the number you see for the collapsed thread and the "showing N comments" line in the expanded thread). Let's not tell people there are e.g. four comments to be seen when it turns out there's only one visible.

In other words: the number should always be the number of survivors, but we should also show where there were how many deletions because that can be relevant for users and is needed by mods.

cellio:
Question: suppose there's a thread with 10 comments, and numbers 2-6 are deleted.  What shows in the thread now, and what shows after this change?  I think after this change you'll see #1 and "4 deleted", right?  Are we currently showing people #1 and 7-10?  Is that a regression?  If so, do we care?

An argument could be made for showing 1, "5 deleted comments", and 7-10, a total of 5 visible comments (which is what the collapsed thread would report).  Would that make sense?  Is it hard?

MoshiKoi:
> Question: suppose there's a thread with 10 comments, and numbers 2-6 are deleted. What shows in the thread now, and what shows after this change? I think after this change you'll see #1 and "4 deleted", right? Are we currently showing people #1 and 7-10? Is that a regression? If so, do we care?

Currently, we would show 1, 7, 8, 9, 10, yes. (The first 5 undeleted comments)
After the change, it would show 1 and "Skipping 4 deleted comments"

Whether this is a regression is whether this behavior is undesirable.

> An argument could be made for showing 1, "5 deleted comments", and 7-10, a total of 5 visible comments (which is what the collapsed thread would report). Would that make sense? Is it hard?

I think it'd make sense, and I don't think it would be too difficult. I'll try and see if I can implement it.

cellio:
Thanks for taking a look.  I think for most users (who cannot see deleted comments), showing 5 comments and indicating that there were also some deleted ones is the desired behavior -- they mostly want to see the comments, after all.  The indicator for deleted comments is for context, for example if an extant comment no longer makes sense -- at least you can see that there was something there that this was a reply to.

MoshiKoi:
Is this the kind of behavior you were talking about? 

image

cellio:
> Is this the kind of behavior you were talking about? Ignore the fact that I can't count apparently and mislabeled the comments

Yes, thanks!  Sanity check: "show them anyway" appears only for people who can see them (like in the full thread page)?

MoshiKoi:
> Sanity check: "show them anyway" appears only for people who can see them (like in the full thread page)? (That is, regular user sees "skipping 4 deleted comments" only, and mods see "show them anyway" with the link.)

It uses the exact same erb template, so yes

MoshiKoi:
It doesn't actually *work* though, since it just redirects the page to (e.g.) `/posts/28?show_deleted_comments=1`, which 1. refreshes the page, so the thread gets closed and 2. doesn't actually show the hidden comments. This will need to be fixed.

app/views/comments/thread.html.erb
```diff
@@ -65,7 +65,11 @@
   <% skipped_deleted = 0 %>
   <% comments = @comment_thread.comments %>
   <% if params[:inline] == 'true' %>
-    <% comments = comments.where(deleted: false).take 5 %>
+    <% count = 0 %>
+    <% comments = comments.take_while do |comment| %>
+      <% count += 1 if !comment.deleted %>
```
ArtOfCode-, line 70:
```suggestion
      <% count += 1 unless comment.deleted %>
```

MoshiKoi:
Closing in favor of #875

PR #872, opened 2022-09-01
Improve subscriptions

### LINKED ISSUE #69: "Latest questions" subscription seems to be sending me everything, not just "latest"
Yesterday I got email from new subscriptions for the first time and it included lots of questions.  That didn't seem wrong for a first message.

Today I got email from those subscriptions and it still includes...a lot of questions.  My subscription for Site Suggestions on Meta includes all three questions, the oldest of which was asked 12 days ago.

Shouldn't "new" mean "new since the last email"?  (Or, more precisely, "new within the frequency you specified", even if it doesn't precisely align with when the email is assembled and sent.)

### LINKED ISSUE #269: Send email subscriptions only when there's new activity
https://languages.codidact.com/questions/278773

> Since I wanted to engage in the language community, but there isn't a lot of questions yet, I set up a daily mail subscription to notify me for new content. I expected "daily" to mean that I get mails for new activity at max once per day, but I am getting an email even if there hasn't been new noticeable activity in a while.
>
> I'd like to suggest to change the behaviour of the notification emails to only be sent if there is new activity in a community.

I remember finding this frustrating in the early days of another community, enough that I turned off email notifications.  As an active user I can check the dashboard instead, but more casual users would benefit from email notifications when there is new content (but not otherwise).

Taeir:
Changes by @MrHug 

* Add button to subscribe to a tag to the tag overview page (tag subscriptions were already implemented, but we could not find the supposed location where one could subscribe to a tag anywhere)
* Only include posts in subscriptions which were made since the last email was sent (or in case of first email, since the subscription was created)
  * This behavior is not present for interesting subscriptions, since those can get to interesting status well after they were posted. A better solution still needs to be "invented" to make those work well.
* Mention in the email that max 25 posts are included when the limit is reached, to avoid the idea that those 25 are all posts created since the last email.

Fixes #69
Fixes #269

### DIFF
--- a/app/views/subscription_mailer/subscription.html.erb
+++ b/app/views/subscription_mailer/subscription.html.erb
@@ -3,6 +3,10 @@
   <% if @subscription.name.present? %>
     "<%= @subscription.name %>"
   <% end %>
+
+  <% if @questions.count == 25 %>
+    This e-mail only shows the 25 most recent items matching your filter.
+  <% end %>
 </h2>
 
 <% @questions.each do |question| %>
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -14,12 +14,14 @@ class Subscription < ApplicationRecord
     case type
     when 'all'
       Question.unscoped.where(community: community, post_type_id: Question.post_type_id)
-              .where('created_at >= ?', last_sent_at)
+              .where(Question.arel_table[:created_at].gteq(last_sent_at || created_at))
     when 'tag'
       Question.unscoped.where(community: community, post_type_id: Question.post_type_id)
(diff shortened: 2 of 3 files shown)


app/views/subscription_mailer/subscription.html.erb
```diff
@@ -3,6 +3,10 @@
   <% if @subscription.name.present? %>
     "<%= @subscription.name %>"
   <% end %>
+
+  <% if @questions.count == 25 %>
+    This e-mail only shows the 25 most recent items matching your filter.
+  <% end %>
```
ArtOfCode-, line 9: Probably want to copy this to subscription.text.erb too.

Taeir: Added in #881

Issue #750, opened 2022-01-02
Code has syntax highlighting in comment thread pages but not inline comments


cellio:
https://software.codidact.com/posts/284928

A comment on Software Dev (where code highlighting is enabled) uses code and the highlighting shows on the thread page: https://software.codidact.com/comments/thread/4992 (third comment).  However, if you expand the same comment thread on the question page, the code is not highlighted: https://software.codidact.com/posts/284910/284911#answer-284911.

I would not have been surprised if syntax highlighting didn't work in comments at all; maybe we don't want to pay that extra cost in performance and/or complexity.  That it works in one of the two places but not the other makes me think this is instead an oversight, not an intentional decision.

Ideally we'd resolve the inconsistency in the generous direction, by enabling it in the expanded thread.  Let's not remove it from the thread view; if there's some reason that we can't also do it in the in-page view, let's document that in an answer to the bug report from Software Meta.

Oaphi:
I think this one has been fixed in #1647 - @cellio, can you confirm that's the case?

<img width="655" height="396" alt="Image" src="https://github.com/user-attachments/assets/e672240b-019e-4acf-a069-2d18fc681b85" />

PR #875, opened 2022-09-03
Comments improvements

MoshiKoi:
Aggregating #873 #869 and #862 since they're all on the same files and I don't want them to clobber each other (I actually had merge conflicts when aggregating)

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -18,7 +18,7 @@ $(() => {
 
     evt.preventDefault();
 
-    $tgt = $(evt.target);
+    const $tgt = $(evt.target);
     openThread($tgt.closest('.post--comments-thread-wrapper')[0], $tgt.attr("href"));
   });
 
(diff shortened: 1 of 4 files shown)


MoshiKoi:
Fixes #750 : Triggers highlight.js on thread expansion, so inline threads have syntax highlighting as well.

image

Resolves #696 : See discussion on #869 and screenshots below:

image

Showing deleted comments works inline too now:

image

app/assets/javascripts/comments.js
```diff
@@ -17,19 +17,31 @@ $(() => {
     if (evt.ctrlKey) { return; }
 
     evt.preventDefault();
-    const $tgt = $(evt.target);
 
-    const resp = await fetch($tgt.attr("href") + '?inline=true', {
+    $tgt = $(evt.target);
```
Taeir, line 21: I think this variable can remain as a `const`:

```suggestion
    const $tgt = $(evt.target);
```

Taeir (commented):
Nice improvements to how things work :) I made one very small suggestion.

ArtOfCode- (approved):

PR #876, opened 2022-09-03
Resize images

### LINKED ISSUE #462: Size of user images varies on users list
https://meta.codidact.com/posts/281072

On the users list, some avatars are larger than others even when they use the same aspect ratio.  For example, in the screenshot linked in that meta post, we can see differently-sized unicornified avatars.  The size also isn't tied to the user; the same avatar can look larger on some communities and smaller on others.  This makes me think it's something about the layout of the users list in particular, and maybe the variation is positional rather than personal.  But I haven't spotted a pattern yet.

### LINKED ISSUE #1133: Codidact loads images in full size
**Describe the bug**
Even when displaying a user icon in a tiny fashion (e.g. as 48x48 in an image card on a post), the user is sent to load the full whatever-uploaded-size (max 2MB) image.

**How to reproduce**
Look at image load times on the users page, and see that the browser is struggling to shrink all the images quickly enough (and if its your first time at that community, it's probably also needing to fetch quite a lot of data).

**Expected behavior**
Qpixel could serve a much smaller version of the image.

**Additional context**
This issue would be solved by #876 . However, in the current setup for the codidact communities, QPixel generates a 301 permanent redirect for image files directly to the S3 bucket. This means that the browser will attempt to fetch from codidact servers once, gets redirected and then always directly from s3.

For regular users, most of the time the browser will pull the images out of its cache, but especially first time viewing a users profile picture can pull in the large image file.

The pull request changes that behavior to server side generate a smaller sized image that fits the display dimensions. So if it is displayed as 48x48 pixels, no need to send more than that to the client. This reduces request sizes a lot. The smaller variant will be stored as well (in Codidacts case, in S3).

Now it does incur additional storage costs (though the smaller images are likely significantly smaller than the source images). Additionally, it changes the redirection behavior to temporary redirects which will require the client to send more requests to codidact servers for these images which will then get redirected to S3. The question is if this is sufficient or will be a problem for expenses.

Taeir:
Images were not being resized, even when displayed as tiny 40x40 squares. For codidact, the image size limit is 2 MB which somewhat alleviates this problem, but for our self hosted instance we upped max image size to 10 MB for posts (which also affects avatars). As a result, if someone uploads a 6 MB avatar, any page which displays it loads the 6 MB giant avatar and then the browser has to scale it down to display it as a tiny image. This causes huge load times and server traffic for pages like the users overview.

In this PR, images are resized by ActiveStorage to the size matching the requested size (without altering aspect ratio). These resized images are created once when requested and are then cached on the storage provider (in your case, Amazon AWS) and not resized again.

As a bonus, images are also upscaled to the requested size if they are smaller, which should resolve #462 and ensure equal image size for all users.

Finally, the profile page would not maintain aspect ratio of the image (in contrast to all other locations which display the avatar). This is fixed in 879040d.

Fixes #462
Fixes #1133

### DIFF
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -61,6 +61,7 @@
 
   img {
     width: 100%;
+    object-fit: contain;
   }
 }
 
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -4,7 +4,7 @@ module UsersHelper
     if deleted_user?(user)
       user_auto_avatar_url(letter: 'X', color: '#E73737FF', size: size, format: :png)
     elsif user&.avatar&.attached?
-      uploaded_url(user.avatar.blob.key)
+      url_for(user.avatar.variant(resize_to_fit: [size, size]).processed)
     else
       user_auto_avatar_url(user, size: size, format: :png)
     end


Taeir:
I just noticed that this does no longer go through the `upload_remote_url` endpoint, which does some transformation on the s3 bucket. The main question is if this method is necessary at all.

ArtOfCode-:
Looks good to merge once #854 is in.

MoshiKoi:
> Looks good to merge once #854 is in.

#854 has been merged; are we waiting on anything else or is this ready?

Taeir:
> > Looks good to merge once #854 is in.
> 
> #854 has been merged; are we waiting on anything else or is this ready? (I see this is still a draft)

@MoshiKoi This alters the permanent redirect behavior to s3 that is currently present (4c05959c).

Taeir:
The permanent redirects were put into place because the image requests were disproportionately slow. Investigate the timings to see if this is still the case.
Another avenue is to ensure that users still get a 301 on the image request by changing the logic a bit.

PR #879, opened 2022-09-08
Make float_notice parameter optional

### LINKED ISSUE #878: Regression: comment threads produce error 500 and can't expand in-page
This is a post on an instance that was updated today (https://github.com/codidact/qpixel/commit/1b07c45df485a0baf64004bdd679d8008dcd017b): https://a.dev.codidact.org/posts/418.  Clicking on the comment link does nothing (doesn't expand it); visiting the thread page produces a 500 error.

Comment threads are working on codidact.com, which was last updated six months ago (https://github.com/codidact/qpixel/commit/2f9c419bd8bec98db8c84edd58dfd4ed508eb5f9).  That's a lot of commits to search, but I suspect it's recent because otherwise the people setting up a recent new deployment would have run into it.  There were some PRs in the last week or two that touched comments, so I'd start there.

MoshiKoi:
Fixes #878

### DIFF
--- a/app/views/posts/_expanded.html.erb
+++ b/app/views/posts/_expanded.html.erb
@@ -5,6 +5,7 @@
     float_notice : whether to display the float notice
 %>
 
+<% float_notice ||= float_notice%>
 <% category ||= defined?(@category) && @category.id == post.category_id ? @category : post.category %>
 <% post_type ||= defined?(@post_type) && @post_type.id == post.post_type_id ? @post_type : post.post_type %>
 
--- a/app/views/posts/_expanded.html.erb
+++ b/app/views/posts/_expanded.html.erb
@@ -5,7 +5,7 @@
     float_notice : whether to display the float notice
 %>
 
-<% float_notice ||= float_notice%>
+<% float_notice ||= false %>
 <% category ||= defined?(@category) && @category.id == post.category_id ? @category : post.category %>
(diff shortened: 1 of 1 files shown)


app/views/posts/_expanded.html.erb
```diff
@@ -24,7 +24,7 @@
     </h1>
   <% end %>
 
-  <% if float_notice %>
+  <% if params[:float_notice] %>
```
ArtOfCode-, line 27: Does this work? This looks like it'd be looking for a parameter, not a variable. There are optional variables defined in a bunch of other views (including the post partial), if you want an example.

Taeir:
I fixed this one for our instance by adding
```ruby
<% float_notice ||= false %>
```

at the top. That is ruby speak for "if float_notice is nil (not set) or false, make it false instead".

app/views/posts/_expanded.html.erb
MoshiKoi: It does work; I've tested it. I'm not sure what you mean by parameter vs variable.

MoshiKoi:
> I fixed this one for our instance by adding
> 
> ```ruby
> <% float_notice ||= false %>
> ```
> 
> at the top. That is ruby speak for "if float_notice is nil (not set) or false, make it false instead".

Yeah this works. My original one doesn't do what I wanted to oops

PR #888, opened 2022-09-12
Allow moderators to directly create tags

### LINKED ISSUE #180: Allow moderators to directly create tags
The usual way to create a tag is to add it to a post.  This is the path we usually want people to take -- create the tag when you need it, to avoid clutter from speculative tags that never get used.  However, there are two use cases I'd like us to support:

1. Up-front creation of "special" tags.  While the meta ones (bug, feature-request, etc) are now seeded, we support the idea of special tags more broadly.  If a community wants to have special tags in other categories, they have to first use them and only then add them to the category definition, which isn't always convenient.  (Also, older communities don't have the meta seed; I just manually created "support" on Writing and there are only two status- tags there so far.)

2. *Planned* tag organization.  The Software Development community wanted a set of language tags up front, as a way of channeling people to the right tags, and IIRC Art added them directly to the database.  Other communities might, after discussion on their metas and otherwise after due consideration, want to create certain key tags, especially when building out tag hierarchies.

Allowing moderators to create tags before use would address both of these cases.  An intuitive place to hook it in would be a (mod-only) "create tag" link on the tags list, though it's ok if it goes in mod tools if that's easier.

eip-ewi:
Closes #180

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -1,8 +1,8 @@
 class TagsController < ApplicationController
-  before_action :authenticate_user!, only: [:edit, :update, :rename, :merge, :select_merge]
+  before_action :authenticate_user!, only: [:new, :create, :edit, :update, :rename, :merge, :select_merge]
   before_action :set_category, except: [:index]
   before_action :set_tag, only: [:show, :edit, :update, :children, :rename, :merge, :select_merge, :nuke, :nuke_warning]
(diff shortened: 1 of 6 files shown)


Taeir:
I posted this as eip-ewi to fix the "requiring approval for every merge request" stuff.

Taeir:
The new button is placed like this on the tags overview page:
image

The creation part looks like this:
image

Issue #756, opened 2022-01-05
Filters phase 2: UI for defining and editing


cellio:
Depends on: https://github.com/codidact/qpixel/issues/755

Add a way to see the definition of the current filter (or define a new one).  This interface should be accessible to users at all levels of technical experience (i.e. not just a place to type a query). https://meta.codidact.com/posts/285370 has an initial mockup; it's almost certainly wrong, but that's the _general_ idea --options in the widget with options that depend on the post type.  (One thing that's missing in that sketch is a way to specify the type of top-level posts, relevant only in categories that have more than one, which is rare today.)

Add "save as" (define a new filter) and "save" (edit an existing one).  For built-in filters, only "save as" -- you can clone them but you can't redefine them.

When a user saves a filter, give an option for "just this category", "just this community", or "global".   This controls where else you'll see it as an option in the filter selector.  Use cases: Software Dev Q&A + Code Review, all the metas.

Add some sort of indicator if you have unsaved changes.

cellio:
This issue is being worked on by @MoshiKoi , who I can't assign it to for some strange reason.

cellio:
Resolved by https://github.com/codidact/qpixel/pull/976.

Issue #755, opened 2022-01-05
Filters phase 1: predefined filters for post list and UI for choosing one (set as default for that community + category)


cellio:
Filters proposal on meta: https://meta.codidact.com/posts/285370

In the first phase, we want to define some built-in filters, implement them on top of search (expanding the search language if necessary), and add some UI for choosing a filter for a category post list.  The choice should be sticky, and "no filter" should be an option (to clear the choice).  I think this bit of UI should be a widget in the right column; in future phases it will need more vertical space than just a selector, so don't put it at the top of the posts list.

The filter framework should not care what kind of post list it's looking at; assume this could be used later for profile pages or search results.  But the UI now is only for category post lists.

Remember that a post list can contain post types other than questions; filters will need to specify post type.  Ideally, the UI would gray out or not offer filters that don't apply -- for example, "open questions" would not be offered on the Meta blog, which uses the article post type.  This might mean that a filter has a "prerequisite" (post type) and a "body" (the actual definition), so that the code can easily check which filters are available.

These are my notes on what *I* think is a sound approach, but people with deeper knowledge of the code can obviously override.

What initial filters should we offer?  Not necessarily a complete list, but some ideas:

- Open questions: post_type:q status:open   (we'll need to add something like a status option)
- Unanswered questions: post_type:q status:open answers:0 
- Exclude low scores (needs a better name): post_type:top-level score:0.4   (do we have the concept of top-level post available? or do we need to enumerate them?)

A nice bonus would be a filter for "my tags" that shows only your favorite tags, but I don't know how to express that in search or what it would take to do it.  We'll need it; I don't know if it's part of the first phase.

cellio:
This issue is being worked on by @MoshiKoi , who I can't assign it to for some strange reason.

cellio:
Work is on https://github.com/codidact/qpixel/pull/890 .

PR #890, opened 2022-09-13
Filters

MoshiKoi:
Proof of concept, looking for feedback

### DIFF
--- a/app/assets/javascripts/filters.js
+++ b/app/assets/javascripts/filters.js
@@ -39,7 +39,7 @@ $(() => {
         }
       });
 
-      // Clear the preset when the user enters in a filter manually
+      // Enable saving when the filter is changed
       $form.find('.form--filter').each((i, filter) => {
         $(filter).on('change', _ => {
           $saveButton.prop('disabled', false);
@@ -59,12 +59,10 @@ $(() => {
         await QPixel.setFilter($tgt.val(), filter);
         $saveButton.prop('disabled', true);
       });
-    });
-  });
-  $('.filter-clear').on('click', evt => {
-    const $tgt = $(evt.target);
-    const $form = $tgt.closest('form');
 
-    $form.find('.form--filter').val(null).trigger('change');
+      $form.find('.filter-clear').on('click', _ => {
+        $form.find('.form--filter').val(null).trigger('change');
+      });
+    });
   });
 });
--- a/app/assets/javascripts/filters.js
+++ b/app/assets/javascripts/filters.js
@@ -6,9 +6,13 @@ $(() => {
 
       $tgt.select2({
         data: Object.keys(filters),
+        tags: true,
       }).on('select2:select', evt => {
         const filterName = evt.params.data.id;
         const preset = filters[filterName];
+
+        // Name is not one of the presets, i.e user is creating a new preset
+        if (!preset) { return; }
 
         for (const [name, value] of Object.entries(preset)) {
           $form.find(`.form--filter[name=${name}]`).val(value);
@@ -20,6 +24,15 @@ $(() => {
         $(filter).on('change', _ => {
           $tgt.val(null).trigger('change');
         });
+      });
+
+      $('.filter-save').on('click', evt => {
+        const filter = {};
+
+        for (const el of $('.form--filter')) {
+          filter[el.name] = el.value;
+        }
+        QPixel.setFilter($tgt.val(), filter)
       });
     });
   });
--- a/app/assets/javascripts/filters.js
+++ b/app/assets/javascripts/filters.js
@@ -12,16 +12,17 @@ $(() => {
         ? ' <span has-font-size-caption">(New)</span>'
         : '';
       return $(name + systemIndicator + newIndicator);
-    }  
+    }
 
     $('.js-filter-select').each((i, el) => {
       const $tgt = $(el);
       const $form = $tgt.closest('form');
+      const $saveButton = $('.filter-save');
 
       $tgt.select2({
         data: Object.keys(filters),
         tags: true,
-        
+
         templateResult: template,
         templateSelection: template
       }).on('select2:select', evt => {
@@ -31,6 +32,8 @@ $(() => {
         // Name is not one of the presets, i.e user is creating a new preset
         if (!preset) { return; }
 
+        $saveButton.prop('disabled', true);
+
         for (const [name, value] of Object.entries(preset)) {
           $form.find(`.form--filter[name=${name}]`).val(value);
         }
@@ -39,17 +42,22 @@ $(() => {
       // Clear the preset when the user enters in a filter manually
       $form.find('.form--filter').each((i, filter) => {
         $(filter).on('change', _ => {
-          $tgt.val(null).trigger('change');
+          $saveButton.prop('disabled', false);
         });
       });
 
-      $('.filter-save').on('click', evt => {
+      $saveButton.on('click', async evt => {
+        if (!$form[0].reportValidity()) {
+          return;
+        }
+
         const filter = {};
 
         for (const el of $('.form--filter')) {
           filter[el.name] = el.value;
         }
-        QPixel.setFilter($tgt.val(), filter)
+        await QPixel.setFilter($tgt.val(), filter);
+        $saveButton.prop('disabled', true);
       });
     });
   });
--- a/app/assets/javascripts/filters.js
+++ b/app/assets/javascripts/filters.js
@@ -35,23 +35,27 @@ $(() => {
         const preset = filters[filterName];
 
         // Name is not one of the presets, i.e user is creating a new preset
-        if (!preset) { return; }
-
+        if (!preset) {
+          $saveButton.prop('disabled', false);
+          $deleteButton.prop('disabled', true);
+          return;
+        }
         $saveButton.prop('disabled', true);
+        $deleteButton.prop('disabled', preset.system);
 
         for (const [name, value] of Object.entries(preset)) {
           $form.find(`.form--filter[name=${name}]`).val(value);
         }
-      })
+      });
+      $saveButton.prop('disabled', true);
+      $deleteButton.prop('disabled', true);
     }
 
     initializeSelect();
 
     // Enable saving when the filter is changed
-    $form.find('.form--filter').each((i, filter) => {
-      $(filter).on('change', _ => {
-        $saveButton.prop('disabled', false);
-      });
+    $form.find('.form--filter').on('change', _ => {
+      $saveButton.prop('disabled', false);
     });
 
     $saveButton.on('click', async evt => {
@@ -67,6 +71,7 @@ $(() => {
       // Reinitialize to get new options
       await initializeSelect();
       $saveButton.prop('disabled', true);
+      $deleteButton.prop('disabled', false);
     });
 
     $deleteButton?.on('click', async evt => {
@@ -75,12 +80,15 @@ $(() => {
         // Reinitialize to get new options
         await initializeSelect();
         $saveButton.prop('disabled', true);
+        $deleteButton.prop('disabled', true);
       }
     });
 
     $form.find('.filter-clear').on('click', _ => {
       $select.val(null).trigger('change');
       $form.find('.form--filter').val(null).trigger('change');
+      $saveButton.prop('disabled', true);
+      $deleteButton.prop('disabled', true);
     });
   });
 });
--- a/app/assets/javascripts/filters.js
+++ b/app/assets/javascripts/filters.js
@@ -3,13 +3,14 @@ $(() => {
     const $select = $(el);
     const $form = $select.closest('form');
     const $saveButton = $form.find('.filter-save');
+    const $deleteButton = $form.find('.filter-delete');
 
     async function initializeSelect() {
       const filters = await QPixel.filters();
-  
+
       function template(option) {
         if (option.id == '') { return 'None'; }
-  
+
         const filter = filters[option.id];
         const name = `<span>${option.text}</span>`;
         const systemIndicator = filter?.system
@@ -20,22 +21,22 @@ $(() => {
           : '';
         return $(name + systemIndicator + newIndicator);
       }
-    
+
       $select.select2({
         data: Object.keys(filters),
         tags: true,
-  
+
         templateResult: template,
         templateSelection: template
       }).on('select2:select', evt => {
         const filterName = evt.params.data.id;
         const preset = filters[filterName];
-  
+
         // Name is not one of the presets, i.e user is creating a new preset
         if (!preset) { return; }
-  
+
         $saveButton.prop('disabled', true);
-  
+
         for (const [name, value] of Object.entries(preset)) {
           $form.find(`.form--filter[name=${name}]`).val(value);
         }
@@ -64,6 +65,15 @@ $(() => {
       // Reinitialize to get new options
       await initializeSelect();
       $saveButton.prop('disabled', true);
+    });
+
+    $deleteButton?.on('click', async evt => {
+      if (confirm(`Are you sure you want to delete ${$select.val()}?`)) {
+        await QPixel.deleteFilter($select.val());
+        // Reinitialize to get new options
+        await initializeSelect();
+        $saveButton.prop('disabled', true);
+      }
     });
 
     $form.find('.filter-clear').on('click', _ => {
--- a/app/assets/javascripts/filters.js
+++ b/app/assets/javascripts/filters.js
@@ -1,69 +1,74 @@
 $(() => {
-  QPixel.filters().then(filters => {
-    function template(option) {
-      if (option.id == '') { return 'None'; }
+  $('.js-filter-select').each((_, el) => {
+    const $select = $(el);
+    const $form = $select.closest('form');
+    const $saveButton = $form.find('.filter-save');
 
-      const filter = filters[option.id];
-      const name = `<span>${option.text}</span>`;
-      const systemIndicator = filter?.system
-        ? ' <span has-font-size-caption">(System)</span>'
-        : '';
-      const newIndicator = !filter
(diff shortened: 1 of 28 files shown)


cellio:
Addresses https://github.com/codidact/qpixel/issues/755 .

app/controllers/users_controller.rb
```diff
@@ -64,6 +64,36 @@ def preferences
...
   end
 
+  def filters
+    system_filters = Rails.cache.fetch 'system_filters' do
+      User.find(-1).filters.to_h { |filter| [filter.name, filter.json] }
+    end
+
+    respond_to do |format|
+      format.json do
+        if current_user
```
ArtOfCode-, line 74:
```suggestion
        if user_signed_in?
```

app/models/filter.rb
```diff
@@ -0,0 +1,15 @@
+class Filter < ApplicationRecord
+  belongs_to :user, required: true, class_name: 'User'
```
ArtOfCode-, line 2: You should be able to drop the `class_name` argument here - Rails should infer it from the `:user` name. `required` also isn't strictly necessary: any reference that isn't explicitly optional is required by default.

MoshiKoi:
I've reached a good point to stop working on further features for this pull request, I think. I've completed most of the stuff in #755 and #756, now it's just testing (maybe spin this branch up on the dev server for people to mess about with?)

Here are the changes:

## Side widget

On the side of category feeds is the filter widget, where users can enter filter options. On submitting, the category will be filtered.

image

## Search filters

Should be fairly intuitive. Filters the results of a search. NOTE: This is *in addition* to any search queries - if the user types "score:>0.6" and filters by a minimum score of "0.5", then it will only show results with score above 0.6. The reverse applies too.

image

## System presets 

System presets are initially populated by seeding. You can take a look in https://github.com/MoshiKoi/qpixel/blob/filters/db/seeds/filters.yml for the format

Note to Administrators: There's no UI tool right now to edit the system presets after the initial population of the database. If you want to modify them, they're just normal filter records assigned to the System user (user_id: -1)

## User-created filters

If a user types in an option that isn't already the name of a preset, saving the filter will create a new filter for them. They can also modify an existing option by hitting save after changing a field.

## Management

There is now a tab on the user page where users can create/edit/delete their filters.

image

## Future Improvements

While not strictly necessary, there are some "nice to haves" I can think off for future work.

 - Don't allow users to edit System presets on the client side. (I already do validation on the backend to make sure they have permissions, but this is just a nice UX thing). Right now, it just fails silently if they try to submit an edit to a System filter.
 - Admin tool to edit System presets. I feel like they change rarely enough that doing without for now shouldn't be too much of a problem, and I already have a seed file which can populate it initially.
 - Make filters that can't apply to a category greyed out (e.g., "min answers" filter option on a category with only Articles). This is a "nice to have", but it's not really necessary. 
 - Add tag includes/excludes to filters. This is somewhat involved from a technical perspective, requiring either changes to the tags API or a new API, so I'm deferring it.
 - Add "favorite tags" option as a filter (relies on the above)

cellio:
Thanks for all your work on this!  This is great!

This comment is based on your comment (not on seeing it live).  I would prefer that users not be able to redefine system filters for themselves; that would remove the ability to get back to the defaults.  The system filters should be immutable, _but_ users should be able to create new filters based on existing ones.  What I'm imagining is: you choose a system filer, edit some properties (maybe your "positive" includes slightly negative, like a min of 0.45), and get a "save as" option where you specify a new name.  That way you have both your version and the system version available.

This means there needs to be a place to type that new name.  I don't care if that's part of the widget or a modal that comes up when you click the button.

On a different subject, can we put the filter parameters behind an expander, so you initially just see the selector for predefined filters, but you can click the expander to see the rest?  I think most people, most of the time, will use a small number of filters and not want to see all the details all the time, which pushes other content in the sidebar down.  (If my description isn't clear enough, let me know and I'll draw a picture.)

MoshiKoi:
> This comment is based on your comment (not on seeing it live). I would prefer that users not be able to redefine system filters for themselves; that would remove the ability to get back to the defaults.

It just silently fails if they try to redefine system filters right now.

> On a different subject, can we put the filter parameters behind an expander

No problem 👍

MoshiKoi:
It doesn't really look as nice though....

image

cellio:
Do you also have a screenshot of the collapsed version?

MoshiKoi:
Her it is collapsed image

cellio:
That looks fine to me.  You could actually make it a little more compact; "Filters" and "predefined filters" are redundant, so we could just have "filters", the selector, and the options behind the expander.  The fact that there's a list means there are some predefined ones, after all. :-)

Can we only light up "save" if there are unsaved changes?  We don't want to lure people into hitting that a bunch of times because they think they have to in order to use a filter.

MoshiKoi:
> Can we only light up "save" if there are unsaved changes?

If I actually select something, then it greys it out - though i suppose i could grey it out on initialization as well

MoshiKoi:
It turns out tag filters were much simpler than expected; I just misunderstood the Select2 documentation

image

MoshiKoi:
> It turns out tag filters were much simpler than expected; I just misunderstood the Select2 documentation

It turns out my initial complexity assessment was actually not that far off... it was complicated in a different way

cellio:
@ArtOfCode- , Moshi said in chat that this is ready for an initial code review.  I see there are merge conflicts which obviously have to be resolved, but can someone can take a look at the code for any other issues in the meantime?

app/controllers/users_controller.rb
```diff
@@ -64,6 +64,100 @@ def preferences
...
+    if params[:system] == true
+      if current_user&.is_global_admin
+        as_user = User.find(-1)
+      else
+        return render json: { status: 'failed', success: false, errors: ['You do not have permission to delete'] },
+                      status: 400
+      end
+    end
+
+    filter = Filter.find_by(user: as_user, name: params[:name])
```
Taeir, line 141: As I understand it, it is possible to delete the filter of the system user (used as default filter for category?), but not to set it with the `set_filter` method above?

app/views/search/_filters.html.erb
```diff
@@ -0,0 +1,75 @@
...
+                                                            selected: @active_filter[:name]),
+          include_blank: true, class: "form-element js-filter-select", id: nil,
+          data: { placeholder: "" } %>
+    </div>
+    <button type="button" class="filter-clear button is-medium is-danger">Clear</button>
+    <% if allow_apply %>
+      <%= submit_tag 'Apply', class: 'button is-medium is-outlined', name: nil %>
+    <% end %>
+    <% if user_signed_in? %>
+      <button type="button" class="filter-save button is-medium is-filled">Save</button>
```
Taeir, line 19: Should we display the button but grayed out (disabled) and with a title set to indicate "this is not possible because you are not signed in"?

Taeir (commented):
I think this is some quite great work @MoshiKoi ! There are quite a few annoying/tricky cases in the search (and other places) that you dealt with in a pretty nice way :)

I do have quite a few review comments, some of those questions about the workings, others small improvements or suggestions. For a final version, I also think we should write some tests, to verify that basic filters work as intended.

I have not yet checked your JavaScript code, I kept to the Ruby and HTML for now. I will take a look at that at a later time if I get to it.

app/assets/javascripts/filters.js
```diff
@@ -0,0 +1,148 @@
+$(() => {
+  $('.js-filter-select').each(async (_, el) => {
+    const $select = $(el);
+    const $form = $select.closest('form');
+    const $formFilters = $form.find('.form--filter');
+    const $saveButton = $form.find('.filter-save');
+    const $isDefaultCheckbox = $form.find('.filter-is-default');
+    const categoryId = $isDefaultCheckbox.val();
+    var defaultFilter = await QPixel.defaultFilter(categoryId);
```
ArtOfCode-, line 9: consider (es+): doesn't need to be a `var` just because it's from `await` (if that was the reason) - `const` if it's a set-once value, `let` if it will be changed later.

app/assets/javascripts/filters.js
```diff
@@ -0,0 +1,148 @@
...
+
+        $isDefaultCheckbox.prop('checked', defaultFilter === $select.val());
+        computeEnables();
+
+        // Name is not one of the presets, i.e user is creating a new preset
+        if (!preset) {
+          return;
+        }
+
+        for (const [name, value] of Object.entries(preset)) {
```
ArtOfCode-, line 90: consider (es+, non-essential): `Object.entries(preset).forEach((k, v) => {` etc.

app/assets/javascripts/qpixel_api.js
```diff
@@ -263,6 +263,95 @@ window.QPixel = {
...
+      credentials: 'include',
+      headers: {
+        'Accept': 'application/json'
+      }
+    });
+    const data = await resp.json();
+    return data.name;
+  },
+
+  setFilterAsDefault: async (category_id, name) => {
```
ArtOfCode-, line 298: request (style): there are a few variables in snake_case in this file - unless there's a reason they need to be, can we change to camelCase?

app/controllers/users_controller.rb
```diff
@@ -64,6 +64,100 @@ def preferences
...
+        render json: filters_json
+      end
+    end
+  end
+
+  def set_filter
+    if user_signed_in? && params[:name]
+      filter = Filter.find_or_create_by(user: current_user, name: params[:name])
+
+      filter.update(min_score: params[:min_score], max_score: params[:max_score],
```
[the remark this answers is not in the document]
ArtOfCode-: No obvious security risks that I can see here, but the validation concerns are... valid (ha!). Worth considering strong params here, i.e. `filter.update(permitted_params)`, where `permitted_params` is a method that calls `params.require(...).permit(...)` etc.

app/controllers/users_controller.rb
```diff
@@ -64,6 +64,100 @@ def preferences
...
+                    min_answers: params[:min_answers], max_answers: params[:max_answers],
+                    include_tags: params[:include_tags], exclude_tags: params[:exclude_tags],
+                    status: params[:status])
+
+      unless params[:category].nil? || params[:is_default].nil?
```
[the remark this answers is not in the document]
ArtOfCode-: This one might be more down to personal preference? I like an unless statement in Ruby - I find they make sense if you read them in natural language to yourself.

"Unless params[:category] is nil or params[:is_default] is nil"

vs

"If params[:category] is not nil and params[:is_default] is not nil"

app/helpers/search_helper.rb
```diff
@@ -1,4 +1,85 @@
...
+  def filter_to_qualifiers(filter)
+    qualifiers = []
+    qualifiers.append({ param: :score, operator: '>=', value: filter.min_score }) unless filter.min_score.nil?
+    qualifiers.append({ param: :score, operator: '<=', value: filter.max_score }) unless filter.max_score.nil?
+    qualifiers.append({ param: :answers, operator: '>=', value: filter.min_answers }) unless filter.min_answers.nil?
+    qualifiers.append({ param: :answers, operator: '<=', value: filter.max_answers }) unless filter.max_answers.nil?
+    qualifiers.append({ param: :include_tags, tag_ids: filter.include_tags }) unless filter.include_tags.nil?
+    qualifiers.append({ param: :exclude_tags, tag_ids: filter.exclude_tags }) unless filter.exclude_tags.nil?
+    qualifiers.append({ param: :status, value: filter.status }) unless filter.status.nil?
+    qualifiers
```
ArtOfCode-, line 39: consider (quality): is it better to iterate through the properties of `filter` and add to `qualifiers` for each that exists, rather than repeated `append/unless` statements?

app/assets/javascripts/filters.js
MoshiKoi: Yes, this should be `let`; thank's for catching that

app/assets/javascripts/filters.js
MoshiKoi: My usage of `for...of` intead of `.forEach` is mostly personal preference since I think it's more readable. Is there a benefit to the `.forEach` version that I'm missing?

app/assets/javascripts/qpixel_api.js
MoshiKoi: Will do 👍

app/assets/javascripts/filters.js
ArtOfCode-: Nope, likewise that's just preference, that's why it's a consider :)

config/config/preferences.yml
```diff
@@ -17,6 +17,7 @@ keyboard_tools:
   description: >
     Enable keyboard shortcuts. Press ? for a list of shortcuts.
   default: 'true'
+  global: true
```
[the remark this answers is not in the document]
MoshiKoi: Done 👍

app/views/search/_filters.html.erb
MoshiKoi: Not sure - I personally prefer it the way it is now, but I'm open to changing it

app/helpers/search_helper.rb
MoshiKoi: I'm not really sure how that would look to be honest; I'm not familiar enough with idiomatic Ruby patterns.

test/fixtures/category_filter_defaults.yml
```diff
@@ -0,0 +1,11 @@
...
+
+one:
+  user: one
+  filter: one
+  category: one
+
+two:
+  user: two
+  filter: two
+  category: two
```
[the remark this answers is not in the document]
MoshiKoi: The test fixtures were autogenerated by rails so that's why they haven't been properly set up (and also I don't know how to set up ruby tests).

app/views/search/_filters.html.erb
```diff
@@ -0,0 +1,75 @@
+<% allow_delete ||= false %>
+<% allow_apply = true if allow_apply.nil? %>
+<% @active_filter ||= {} %>
+
+<div class="form-group">
+  <div class="form-group-horizontal">
+    <div class="form-group">
+      <%= label_tag :predefined_filter, 'Predefined Filters', class: "form-element" %>
+      <%= select_tag :predefined_filter, options_for_select([@active_filter[:name]],
```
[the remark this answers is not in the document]
MoshiKoi: It doesn't; this is just to set the active filter as the active selection in the select element. Select2 fills in the actual options.

app/views/search/_filters.html.erb
cellio: What is the current behavior?  (She says, hoping to avoid having to pull and set it up locally to find out. :-) )

db/migrate/20220916075849_add_tags_to_filters.rb
```diff
@@ -0,0 +1,6 @@
+class AddTagsToFilters < ActiveRecord::Migration[7.0]
+  def change
+    add_column :filters, :include_tags, :string
```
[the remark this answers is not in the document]
MoshiKoi: This is the same way that tags are stored on posts and it seems to be doing okay - I don't imagine someone is going to be filtering by more tags than can fit on a post, but I could be wrong. 

https://github.com/codidact/qpixel/blob/ad595c3b2d186d77a9070894af12be050bf8f853/db/schema.rb#L411

app/views/search/_filters.html.erb
MoshiKoi: It's just Not There™️

test/fixtures/filters.yml
```diff
@@ -0,0 +1,17 @@
+# Read about fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
+
+one:
+  name: MyString
+  min_score: 1.5
```
Taeir, line 5:
```suggestion
  user: standard_user
  min_score: 1.5
```

app/models/user.rb
```diff
@@ -24,6 +24,7 @@ class User < ApplicationRecord
   has_many :comments, dependent: :nullify
   has_many :comment_threads_deleted, class_name: 'CommentThread', foreign_key: :deleted_by_id, dependent: :nullify
   has_many :comment_threads_locked, class_name: 'CommentThread', foreign_key: :locked_by_id, dependent: :nullify
+  has_many :filters, dependent: :destroy
```
Taeir, line 27: Tests fail because this doesnt reference the category_filter_defaults, needs:

```suggestion
  has_many :filters, dependent: :destroy
  has_many :category_filter_defaults, dependent: :destroy
```

app/models/user.rb
```diff
@@ -24,6 +24,7 @@ class User < ApplicationRecord
...
   belongs_to :deleted_by, required: false, class_name: 'User'
```
Taeir, line 28: On another note, the question is why category_filter_defaults is linked to user, since it is also linked to filter and filter is linked to a user.

MoshiKoi: The problem is the System filters. We could have, for instance

Filter(Positive, belongs to System)
CategoryFilterDefault(Moshi, Meta, Positive)

To be honest, attaching "universal" filters to System might not have been the brightest idea, since it leads to weird edge cases like this.

MoshiKoi:
This test no longer succeeds due to [a change in how search works](https://meta.codidact.com/posts/287031):

https://github.com/codidact/qpixel/blob/8f36e90346ce1413cab8e6973620381a184c0153/test/controllers/search_controller_test.rb#L6-L10

Now, an empty search will return all posts. Should this method be changed to instead assert not nil?

MoshiKoi:
Since it was easy, I just added in #905 as well.

app/helpers/search_helper.rb
Taeir: Since the treatment of each property is different, I see no good way to do what you propose.

MoshiKoi:
Work is continuing on #976

PR #892, opened 2022-09-15
Do not display closed posts in hot posts

Taeir:
We had some of our very old but closed posts (no longer relevant, relating to a previous academic year) show up in hot posts due to an edit, comment or something else. We believe that closed posts are never a good fit for hot posts, so that is what this PR changes.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -268,7 +268,8 @@ class ApplicationController < ActionController::Base
 
     @hot_questions = Rails.cache.fetch('hot_questions', expires_in: 4.hours) do
       Rack::MiniProfiler.step 'hot_questions: cache miss' do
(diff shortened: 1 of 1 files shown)


cellio:
Good idea.  How does this change affect posts that can't be closed (like articles)?  I don't know what `closed: false` does in this case.

Taeir:
Good question. I checked with the schema and `closed` is a field on all posts which cannot be null, so it must be either true or false, even for articles.

ArtOfCode-:
Excluding closed posts is fine, but as @cellio says it won't exclude things like articles from last year's course. You could exclude locked posts as well, if they're not already, and lock posts from previous years?

Taeir:
@ArtOfCode- updated to also exclude locked posts. We don't have any articles that are "year-bound" so that is not a problem for us.

PR #893, opened 2022-09-15
276 Tag synonyms

### LINKED ISSUE #276: Tag synonyms
https://judaism.codidact.com/questions/276821

Sometimes there's more than one way to describe a concept and the words are very different -- even more so when a community uses more than one language, like Judaism where there are both Hebrew and English terms for things.  It would be helpful to be able to create tag synonyms so all those posts get tagged the same way regardless of how the author thinks of the concept.

Taeir:
Adds support for tag synonyms. Best explained with some images.

### Tags overview + search
image
image

Combined with hierarchy
image

I chose teal to have a different color from the hierarchy.

### Viewing a single tag with synonyms
image showcasing the overview of a single tag

### Tag edit/create
image

### Tag search
example of tag search for terrier, showing how synonyms are displayed
example of tag search for dog, showing how synonyms are displayed

### TODO
* [x] Testing
* [x] Performance considerations for many tags
* [x] CSS Styling alignment of delete button for tags.

### Feedback
The javascript is perhaps not the prettiest (I don't particularly like javascript). There are libraries which do the logic shown here (i.e. the cocoon library for rails allows dynamic form generation) but better and cleaner. We could consider using the library instead.

Closes #276


ArtOfCode-:
Worth considering: what happens when a tag has lots and lots of synonyms? It's not implausible that a widely used tag could end up with a few dozen synonyms. Should the system only show the most popular synonyms, or just the first few? Is showing most popular performance friendly? How many does it show - and is that configurable?

cellio:
For the case Art raised, consider showing the first N (whether "first" is popular or first-defined or alpha or...?) and a control to pull up the full list.  I'm thinking of something like the reactions interface in Discord, where you get "thumbup by A, B, C, and 14 others" and you have to click something to see all 17.

Taeir:
@ArtOfCode- Good considerations.

On tag show page, no issue to display all:
image

On tag overview page, issues:
image

On tag search, issues:
image

With regards to popularity, that can with its current implementation not be tracked unless if users vote themselves somewhere. This is because the synonym is only used for the search and then returns the true tag, so there is no link between post and synonym. To track popularity of the synonym we would also have to store the synonym used on the post, which would complicate the required logic and code changes quite a lot. In that case, synonyms should work more like hierarchical tags do.

Limiting to first-n synonyms (configurable or not) shown is possible, but I would need to create some more complex javascript on the front end to ensure that the searched-for-synonym is contained in the displayed list (That or significantly change the search such that the backend is aware of the synonym that matched). In that case I would still be sending all synonyms of a tag to the frontend and filter the display there, essentially equivalent to a "broader" tag search and having the performance of one.

On another note, is alias perhaps a better name than synonym?

Taeir:
@cellio I had a look at select2 and it seems that it is doable to create such a component. However, I'm not really a great front-end designer so then I would perhaps like some help with designs of this before implementing something (perhaps good to have the wanted design clear before moving on).

not-a-ethan:
Who would be able to make synonyms? Anybody? People with x amount of positively reserved posts in a tag? Only mods?

Taeir:
@not-a-ethan At the moment anyone who is able to edit tags (I guess only mods?). But we should consider whether voting to add a synonym to a tag should be another feature like suggesting edits (though I would keep that separate from this PR to keep the scope reasonably small and add that as an additional feature later).

cellio:
I agree about popularity; given how we're doing this, that doesn't make sense, and tracking popularity is not important and we shouldn't rework how we're doing this in order to support it.

> @cellio I had a look at select2 and it seems that it is doable to create such a component. However, I'm not really a great front-end designer so then I would perhaps like some help with designs of this before implementing something (perhaps good to have the wanted design clear before moving on).

Another approach would be to just say "N aliases" and not try to list them where they don't fit.  That would mean (and I should clarify I'm thinking out loud here, not dictating):

- individual tag page: as you have it now
- tags list: tag name + "(N aliases)" if there are any; either show the list on hover somehow or just make people click through if they want to know
- tags editor on posts: this is a little more complicated because if you start typing "terrier" and see "dog" that could be a little confusing, but is it practical to show just the primary tag and the alias you're typing, and not all of them?  In this example, "dog (terrier)", or if we want to be a little more informative, "dog (terrier, 13 more aliases)"?

What do you think?  Is that practical for both users and implementation?

MoshiKoi:
Suggestion: "Merge as Synonym"

I think this would be a good addition (in addition to, rather than replacing) our current merging logic - If someone added the tag, chances are that more people will also think of adding the same tag, and it'd be nice to redirect them to the canonical version of the tag.

MoshiKoi:
> @cellio I had a look at select2 and it seems that it is doable to create such a component. However, I'm not really a great front-end designer so then I would perhaps like some help with designs of this before implementing something (perhaps good to have the wanted design clear before moving on).

What about something like this?

image
image

Taeir:
@cellio

With latest changes, I have limited it to 3 synonyms max
image
image
image
image

cellio:
Nice!  The part about only showing potential matches is especially nice.

app/assets/javascripts/tags.js
```diff
@@ -73,6 +73,22 @@ $(() => {
...
   });
 
+  function processSynonyms($search, synonyms) {
+    if (!synonyms) return synonyms;
+
+    if (synonyms.length > 3) {
+      const searchValue = $search.data('select2').selection.$search.val().toLowerCase();
+      displayedSynonyms = synonyms.filter(ts => ts.name.includes(searchValue)).slice(0, 3);
+    } else {
+      displayedSynonyms = synonyms;
```
ArtOfCode-, line 83: `displayedSynonyms` is missing a full declaration. You probably want to add `let displayedSynonyms;` before the if-else block, so that you can set it in each branch.

ArtOfCode- (approved):

PR #896, opened 2022-09-21
Fixes #889

### LINKED ISSUE #889: Time difference between last activity and post timestamps on post creation
When creating a post, in the category feed it shows a "last activity" timestamp different from the "posted" timestamp.

Example from [the dev server](https://a.dev.codidact.org/categories/1)

posted 9m ago by Test User‭  ·  last activity 4h ago by Test User

Example from my own dev instance:

posted <1m ago by Moshi‭  ·  last activity 7h ago by Moshi‭

As shown, the time difference is not the same across instances, but seems consistent within the same instance. Therefore, it is likely that it is relying on the local time for the instance somewhere when calculating how long ago the last activity was.

MoshiKoi:


app/controllers/posts_controller.rb
```diff
@@ -48,6 +48,8 @@ def create
                 end || nil
     @post = Post.new(post_params.merge(user: current_user, body: helpers.post_markdown(:post, :body_markdown),
                                        category: @category, post_type: @post_type, parent: @parent))
+    @post.save
```
cellio, line 51: How does this save interact with the subsequent checks that can reject the post (e.g. title too short)?  Is there another step beyond save where the changes get committed?

PR #897, opened 2022-09-21
Fix seeding and caching

Taeir:
Added logic for seeding to translate all _id mappings in the seeds to their actual rails objects to please rails and its validations.

Additionally fixes redis by reverting it to version 4 and updating the settings. Version 5 currently has a bug which causes issues for the development setup, and in some cases also the production setup. See https://github.com/redis/redis-rb/issues/1157


Taeir:
Better solution, downgrade redis-rb to version 4 to prevent a whole bunch of problems...

Issue #899, opened 2022-09-30
Adjust wording on vote-summary page


cellio:
https://meta.codidact.com/posts/287101

> In the votes summary page of my profile it says: "A daily summary of votes you have received for your posts."  This is an accurate description of my vote summary page. However, if I look at someone else's vote summary page, it still says "votes you have received for your posts".

PR #900, opened 2022-09-30
Generalized wording on vote-summary page

cellio:
Fixes https://github.com/codidact/qpixel/issues/899 .  Not a critical change or anything, but easy.  New wording is: "A daily summary of votes received for posts."  The user name is already shown immediately above, so no need to repeat it here.

screenshot showing change


cellio:
Err, I changed a couple words in one paragraph.  I don't understand the CI failure?

PR #918, opened 2022-12-09
Add setup for creating closed communities

Taeir:
This adds configuration options for creating a closed community (which requires users to be signed in before they can see any content).

The restriction is added in the application controller such that _all endpoints_, except for error pages, javascript and css, assets and the help and policy pages, are protected. The original location a user was visiting is also stored such that they get redirected to the right place after signing in.

NOTE: images are protected as well, that is what the ActiveStorage base controller is for. I'm note sure this is entirely necessary, what are your thoughts?
Since it does not descend from ApplicationController it does not actually have most helper methods, and rendering our default error pages is not possible there. Instead, I have it redirect to the home page with a forbidden status code.


Taeir:
Rubocop issues:
* Application controller has become too long (322 lines), what should I do to fix that?
* Other issues would be fixed with #936

cellio:
> Rubocop issues:
> 
> * Application controller has become too long (322 lines), what should I do to fix that?

@ArtOfCode-  any suggestions?

> * Other issues would be fixed with [Fix test/rubocop errors #936](https://github.com/codidact/qpixel/pull/936)

That PR has now been merged.  Does merging it into this branch fix those issues now?

ArtOfCode-:
This has been here for a while... I hadn't even noticed it! My immediate response is that if we're providing a way to set a closed community, it should be _fully_ closed - folks using it may not realise it has pitfalls like not protecting help/policy (or images, if we were to change that). Pure assets (JS, CSS) are one thing, but user-uploaded images should definitely be protected, and I'd suggest we protect help and policy pages too.

Taeir:
@ArtOfCode- 

> user-uploaded images should definitely be protected

Okay, then let's do so in the way this PR does.

> and I'd suggest we protect help and policy pages too.

My main reasoning for not doing so is that the privacy policy, which may factor in a decision to join a community, would be blocked until signed in. I was trying to account for multiple different types of closed communities. For example, one that allows registrations normally and allows anyone to join, but you only see details once you have joined. You would want the policy pages to be visible in that case.

ArtOfCode-:
That makes sense now you mention it. I can see two options then - either we could protect _help_ pages but not _policy_ pages (makes the privacy policy/ToS etc available before joining, but potentially has the same downfall if there's useful info in the help pages), or we could add a site setting to give admins the option to protect or not.

Taeir:
@ArtOfCode- I have updated this PR and added the suggested settings.

ArtOfCode- (approved):

PR #919, opened 2022-12-13
Migrate CI workflow from CircleCI

sau226:
This PR uses a modified output from the GitHub [Actions Importer](https://docs.github.com/en/actions/migrating-to-github-actions/automating-migration-with-github-actions-importer) to migrate all workflows from CircleCI to GitHub Actions itself.

By using GitHub Actions, we can ensure that CI can be managed adjacent to code without needing to depend on a third party service (CircleCI), and integrates it natively into GitHub (on PRs etc).

You'll need to create 3 [actions secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository) to make this config work. These secrets are stored as `MASTER_KEY`, `COVERALLS_REPO_TOKEN` and `DEV_SSH_KEY` in CircleCI env vars. Add the actions secrets with same name and value as the CircleCI env vars.

As this is a test of the migration tool, the migrated workflow will still be using CircleCI docker images, at least for now. We can migrate to alternatives once we feel comfortable with having integrated actions in GitHub.


ArtOfCode-:
Info: is there a particular reason we should move out of Circle?

ArtOfCode-:
This has diverged from the Circle workflow now - we also have system tests, and both tests and system tests are run against ruby 2.7 and 3.1.

Taeir:
@ArtOfCode- I think it may still be valuable to switch over to GitHub actions. Updating this PR to match should not be too difficult as it contains sort of the same information (and still uses CircleCI's docker images).

ArtOfCode-:
I'm closing this one at the moment. We don't seem to have had any major reasons to move away from Circle in the last year. This would also need review for correctness against current process.

Oaphi:
Permission to reopen, @ArtOfCode- under my responsibility? We've recently had trouble where no one with access has been available to simply rerun a jobs on CircleCI ([reference discussion](https://discord.com/channels/634104110131445811/676843271762214932/1314425713654239252)), our system tests there are also flaky, which contributes to the previous problem significantly. Additionally, it's a third-party dependency that is not mission-critical (speaking from experience that CI done via GH Actions fully covers our use-case, is at least second-party, and will allow us to setup up our own runners going forward).

ArtOfCode-:
@Oaphi no objection

Oaphi:
> @Oaphi no objection

Much obliged!

Oaphi:
@sau226 apologies in advance for converting the PR to draft for now - need to look through it, possibly bring it up to date and devise a path forward if everything checks out given that this is a breaking change to the CI pipeline.

Oaphi:
My apologies, @sau226, I completely forgot that we'd already had work on dropping CircleCI started (and basically completed) here - losing access to the pipeline really got on my nerves, it seems. Closing this as the switch from CircleCI to GitHub Actions has happened in #1513

PR #926, opened 2022-12-29
Fix double bracketed footnotes

### LINKED ISSUE #259: Footnotes are now displayed with an extra pair of [] - usually
https://meta.codidact.com/questions/278943

Footnotes in post bodies now (usually) show up with two square brackets, as in the linked question, and https://languages.codidact.com/questions/278240#answer-278909, and https://meta.codidact.com/questions/278644.  However, it doesn't happen in https://meta.codidact.com/questions/277504.  I looked for differences in the Markdown source on those posts but didn't spot anything unusual.

MoshiKoi:
Fixes #259

Also re: https://meta.codidact.com/posts/287424

The footnote plugin automatically generates closing brackets, which means that adding it via. CSS is superfluous. See the [example output here](https://github.com/markdown-it/markdown-it-footnote).

Note: This will result in footnotes that currently only have one pair of brackets having none. Unfortunately, this is an inconsistency in the HTML itself which cannot be corrected except by editing. I believe that this is because in the past, we had inconsistency in whether the server-side or client-side code would render post's to markdown, as has been noted before. Since we have moved to client-side rendering, inconsistencies should no longer be an issue going forward.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -145,22 +145,6 @@ hr {
   }
 }
 
-.footnote-ref a::before {
-  content: '[';
-}
-
-div.post-preview .footnote-ref a::before {
-  content: '';
-}
-
-.footnote-ref a::after {
-  content: ']';
-}
-
-div.post-preview .footnote-ref a::after {
-  content: '';
-}
-
 .footnotes-sep + .footnotes {
   border-top: 0;
 }


cellio:
See also [this comment thread](https://meta.codidact.com/comments/thread/6966).  Are we saving the brackets as HTML in the posts table, and that's the root of the problem?

MoshiKoi:
The brackets are in the HTML, yes.

cellio:
Ok, brackets in the HTML, so the CSS doesn't need to add them.  For posts saved in some time window I don't know, there are no brackets in the HTML so with this change they'll go away.  I agree there's nothing reasonable we can do about that.  If editing the post restores them, then there's an obvious fix for any cases that are bothering people.

PR #929, opened 2023-01-02
Clarify fields on password-change form

cellio:
Re: https://meta.codidact.com/posts/287649

Changes the order to: current pwd, new pwd, confirm new pwd, and also clarifies some labels.

screenshot

While looking for the right source file, I also came across app/views/devise/passwords/edit.html.erb, which doesn't match any page I can find from the user profile on a live site.  Is that file used?  Where?


cellio:
Rubocop is unhappy with files I didn't change.  Am I supposed to try to correct that?

PR #931, opened 2023-01-03
Better Notifications

### LINKED ISSUE #930: Make the inbox page more visible
Someone asked on meta how to see the full inbox (not just the items in the drop-down).  There's a link at the very bottom of that modal, which was pointed out in  [this comment thread](https://meta.codidact.com/comments/thread/7086), but, demonstrably, most people don't notice it because you have to scroll all the way to the end.

But this means that the page exists: /users/me/notifications on any community shows your full inbox, paginated.  While the URL is on a particular community, it's the *global* inbox, containing everything that showed up in the drop-down modal. 

Proposal: let's expose this global inbox from the profile page somehow.  Options I can think of:

1. Add a "global inbox" tab (include "global" so people don't think it's filtered).  We have network preferences, so network-wide stuff here isn't necessarily wrong.  Like "account", "edit", and "preferences", this is shown only on your own profile when you are logged in.
2. Add a link (not a tab) on your own profile to your global inbox, perhaps alongside the website/twitter/discord links at the top or the "subscribe" and "mobile sign-in" buttons under the "about" section, or in the right column somewhere (though that's less accessible on a phone).  By making it a link not a tab, we don't have to change anything about the inbox page itself.

2 seems simple: decide where to put the link and edit the profile page's HTML.  1 is a little more work: add the tab to all the profile sections and add all the tabs to the inbox page.  That doesn't sound that complex either, though I don't know how the tabs are managed (I hope there's a single item that all the pages include but I don't know that).

MoshiKoi:
Fixes #930 

 - Moves the "See all notifications" button to the top for visibility
   image
 - Adds a "Notifications" tab to the profile page
   image

### DIFF
--- a/app/assets/javascripts/notifications.js
+++ b/app/assets/javascripts/notifications.js
@@ -67,8 +67,6 @@ $(() => {
         const item = $(makeNotification(notification));
         $inboxContainer.append(item);
       });
-  
-      $inboxContainer.append(`<a href="/users/me/notifications" class="button is-muted is-small">See all your notifications &raquo;</a>`);
     }
   });
 
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -167,6 +167,7 @@
   <div class="h-p-2 h-bg-tertiary-050 h-fw-bold">Notifications</div>
   <div class="h-p-2 h-bg-tertiary-050 h-fw-bold">
     <a class="no-unread js-read-all-notifs" href="#"><i class="fas fa-envelope-open"></i> Mark all as read</a>
+    <a href="/users/me/notifications" class="button is-muted is-small">See all your notifications &raquo;</a>
   </div>
   <div class="inbox--container h-p-2"></div>
 </div>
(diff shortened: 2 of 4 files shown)


trichoplax:
I like the "See all your notifications" at the top of the notifications drop down panel, but does this change also remove it from the bottom? I'm aware that the top section doesn't scroll out of view, so even after scrolling through all the available messages the user can still see the "See all your notifications" link at the top, so it isn't essential to also show it at the bottom.

I just wondered if it might help discoverability to also show the link at the bottom (where it currently is before this change), as that's exactly where the user will be looking after scrolling down and not finding the message they were looking for.

MoshiKoi:
@trichoplax I deleted the code that added it to the bottom, so yes, it's gone from there. I can just add it back if it'd help with discoverability, I just thought it wasn't really necessary to keep around.

trichoplax:
That was just my personal opinion on discoverability - not my call. Thought I'd mention it for discussion since it won't take up any extra space on the page (being the very bottom of a scrollable panel).

Taeir (approved):
👍 Code changes look good. Since the "See all your notifications" is stuck to the top and stays on top I think only having it there is fine. It will always be on screen.

Issue #906, opened 2022-10-18
Add structured Q&A schema data


ullallulloo:
Currently it doesn't appear that Codidact has any schema data on the pages. Adding such should help Codidact in both ranking and presentation in search engines. An overview of what is needed for such can be found here: https://developers.google.com/search/docs/appearance/structured-data/qapage

JSON-LD seems like the more popular way of going about such and could produce cleaner resultant code, but Microdata would be much easier to implement. It might be hard to get all the question information into the header, but I'm not really familiar with how the templates work here.

PR #932, opened 2023-01-03
Add structured data for question pages

MoshiKoi:
Working on #906

This should work, as shown [here](https://search.google.com/test/rich-results/result?id=QXJfszzFDEbJYqjtePRFDA), but I'm not sure why the preview doesn't. If you have any insights, please let me know; I'm not familiar with SEO things.

### DIFF
--- a/app/views/posts/_schema.html.erb
+++ b/app/views/posts/_schema.html.erb
@@ -1,0 +1,27 @@
+<% if post && post.post_type.name == 'Question' %>
+  <% suggested_answers = @children.filter { |answer| answer.score > 0.5 } %>
+  <script type="application/ld+json">
+    {
+      "@context": "https://schema.org",
+        "@type": "QAPage",
+        "mainEntity": {
+        "@type": "Question",
+        "name": "<%= @post.title %>",
+        "text": "<%= @post.body.gsub(/\r?\n/, '') %>",
+        "answerCount": <%= @children&.length || 0 %>,
+        "upvoteCount": <%= @post.upvote_count %>,
+        "suggestedAnswer": [
+        <% suggested_answers.each_with_index do |answer, i| %>
+          <% if i > 0 then %>,<% end %>
+          {
+          "@type": "Answer",
+          "text": "<%= answer.body.gsub(/\r?\n/, '') %>",
+          "upvoteCount": <%= answer.upvote_count %>,
+          "url": "<%= post_url(answer) %>"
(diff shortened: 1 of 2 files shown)


app/views/posts/_schema.html.erb
```diff
@@ -0,0 +1,27 @@
...
+        "@type": "QAPage",
+        "mainEntity": {
+        "@type": "Question",
+        "name": "<%= @post.title %>",
+        "text": "<%= @post.body.gsub(/\r?\n/, '') %>",
+        "answerCount": <%= @children&.length || 0 %>,
+        "upvoteCount": <%= @post.upvote_count %>,
+        "suggestedAnswer": [
+        <% suggested_answers.each_with_index do |answer, i| %>
+          <% if i > 0 then %>,<% end %>
```
ArtOfCode-, line 15: `<% suggested_answers.first %>` should work here?

MoshiKoi: Not quite sure how this would work? This is just to get a comma-separated array for the JSON, since the Google analyzer complained when I had a trailing comma.

MoshiKoi: Unfortunately, that doesn't work, and it looks like rails doesn't have a cleaner way to do it other than manually checking

ullallulloo:
Google is very unpredictable about what previews it shows. I don't know why the preview doesn't show up in Google's tool, but I don't see how it could be anything wrong with the schema data.

cellio:
This has been in draft state for a while.  What additional work do we need to do?

MoshiKoi:
Honestly forgot I made this whoops

The main thing is that I have no way of actually knowing whether it works or not until it gets put on the server and I see if Google does the preview when I search. As mentioned above, Google's test page doesn't preview it correctly, so I'm not sure the generated schema is correct.

Maybe it could be put on the dev server to test...? I'm not sure if it's indexed.

cellio:
If it doesn't work, will the change do any harm?  (I gather not?)  @ArtOfCode- , do you have an opinion on this?

ullallulloo:
Note that I made a PR on @MoshiKoi 's fork fixing a couple of bugs, but besides that I would say it should be merged. Providing the data can only help.

cellio:
Can we get ullallulloo's changes and these ones together and proceed with this?  Sounds like it can't hurt and might help.

PR #933, opened 2023-01-03
Make voting tool sticky

MoshiKoi:
Somewhat old thing that I just came across: https://meta.codidact.com/posts/279092/279096#answer-279096

This makes the voting tools sticky, so it will remain visible even after scrolling.

image

A minor issue: Due to the way the page is laid out, it will in fact go past the article and into the comments section. (That being said, why is the comment section part of the post body?)

image

### DIFF
--- a/app/assets/stylesheets/posts.scss
+++ b/app/assets/stylesheets/posts.scss
@@ -12,6 +12,11 @@
   }
 }
 
+.vote-widget {
+  position: sticky;
+  top: 0;
+}
+
 .post--container {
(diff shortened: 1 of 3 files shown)


ArtOfCode-:
Question and a request: does this _only_ sticky the widget within the post, or does it continue down the whole page? How does it work with multiple posts on the page?

And a request: I know some folks will love this, but personally I like my voting buttons in the same place every time - can we make this preference-controlled?

PR #934, opened 2023-01-03
Pending suggestions

### LINKED ISSUE #452: Make pending edits more obvious
https://meta.codidact.com/posts/280872

The linked post seeks input on how to make the existence of pending edit suggestions more visible.  Right now edit suggestions can sit for weeks or months because no one noticed them.  This issue is to explore the options and figure out what's feasible.

One possibility is to unify pending edits from all categories, instead of having to go to each category to see them, and light up that community-wide indicator if something is waiting.  Another possibility is to leave the "edits" link where it is now but, similarly, light it up.

If we light up an indicator, we need to consider whether it can be dismissed.  One approach would be show the number of pending edits (rather than a boolean "are/aren't" indicator), and leave it to users to notice if the number is changing.  Another approach would be light it up when there's something new since the last time this user looked, but I don't know if per-user tracking is performant.  (I guess we do that for the "new posts" category indicators, so maybe?)  Maybe there are other approaches.

This issue is for an exploration task, to think about how this should work and what would be practical before diving into the code.

MoshiKoi:
Adds an indicator for when there are pending suggested edits in the category. Resolves #452 (Baseline)

image

### DIFF
--- a/app/models/suggested_edit.rb
+++ b/app/models/suggested_edit.rb
@@ -10,6 +10,12 @@ class SuggestedEdit < ApplicationRecord
   has_and_belongs_to_many :tags
   has_and_belongs_to_many :before_tags, class_name: 'Tag', join_table: 'suggested_edits_before_tags'
 
+  after_save :clear_pending_cache
+
+  def clear_pending_cache
+    Rails.cache.delete "pending_suggestions/#{post.category_id}"
+  end
+
   def pending?
     active
   end
--- a/app/models/suggested_edit.rb
+++ b/app/models/suggested_edit.rb
@@ -10,7 +10,7 @@ class SuggestedEdit < ApplicationRecord
   has_and_belongs_to_many :tags
   has_and_belongs_to_many :before_tags, class_name: 'Tag', join_table: 'suggested_edits_before_tags'
 
-  after_save :clear_pending_cache
+  after_save :clear_pending_cache, if: Proc.new { saved_change_to_attribute?(:active) }
 
   def clear_pending_cache
     Rails.cache.delete "pending_suggestions/#{post.category_id}"
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -244,8 +244,13 @@
                       class: "category-header--nav-item #{active?(current_cat) && !['tags', 'suggested_edit'].include?(controller_name) ? 'is-active' : ''}" %>
           <%= link_to 'Tags', category_tags_path(current_cat),
                       class: "category-header--nav-item #{active?(current_cat) && controller_name == 'tags' ? 'is-active' : ''}" %>
-          <%= link_to 'Edits', suggested_edits_queue_path(current_cat),
-                      class: "category-header--nav-item #{active?(current_cat) && controller_name == 'suggested_edit' ? 'is-active' : ''}" %>
+          <%= link_to suggested_edits_queue_path(current_cat),
+                      class: "category-header--nav-item #{active?(current_cat) && controller_name == 'suggested_edit' ? 'is-active' : ''}" do %>
+            Edits
+            <% if SuggestedEdit.where(post: Post.undeleted.where(category: current_cat), active: true).any? %>
+              <span class="badge is-status" title="Suggested edits pending"></span>
+            <% end %>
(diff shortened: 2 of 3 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -244,8 +244,16 @@
...
           <%= link_to 'Tags', category_tags_path(current_cat),
                       class: "category-header--nav-item #{active?(current_cat) && controller_name == 'tags' ? 'is-active' : ''}" %>
-          <%= link_to 'Edits', suggested_edits_queue_path(current_cat),
-                      class: "category-header--nav-item #{active?(current_cat) && controller_name == 'suggested_edit' ? 'is-active' : ''}" %>
+          <%= link_to suggested_edits_queue_path(current_cat),
+                      class: "category-header--nav-item #{active?(current_cat) && controller_name == 'suggested_edit' ? 'is-active' : ''}" do %>
+            Edits
+            <% has_pending = Rails.cache.fetch "pending_suggestions/#{current_cat.id}" do %>
+              <% SuggestedEdit.where(post: Post.undeleted.where(category: current_cat), active: true).any? %>
+            <% end %>
```
ArtOfCode-, line 252: suggestion: This is controller-type code in a view - should probably move to a helper that can be called here instead.

app/models/suggested_edit.rb
```diff
@@ -10,6 +10,12 @@ class SuggestedEdit < ApplicationRecord
   has_and_belongs_to_many :tags
   has_and_belongs_to_many :before_tags, class_name: 'Tag', join_table: 'suggested_edits_before_tags'
 
+  after_save :clear_pending_cache
+
+  def clear_pending_cache
+    Rails.cache.delete "pending_suggestions/#{post.category_id}"
```
ArtOfCode-, line 16: query: At the moment, this clears the indicator after _any_ save of _any_ suggested edit in the category. Is that intended, or do we need to add some logic here to only clear it if it's actually clear?

ArtOfCode- (changes requested):

app/models/suggested_edit.rb
MoshiKoi: It was just the simplest way I could think of to trigger a recalculation of whether there are any pending suggestions; it probably could be made a bit more efficient

app/models/suggested_edit.rb
```diff
@@ -10,6 +10,12 @@ class SuggestedEdit < ApplicationRecord
   has_and_belongs_to_many :tags
   has_and_belongs_to_many :before_tags, class_name: 'Tag', join_table: 'suggested_edits_before_tags'
 
+  after_save :clear_pending_cache
```
Taeir, line 13: This should improve the efficiency (note I did not check this directly). Is there a better property to check whether it is valid?

```suggestion
  after_save :clear_pending_cache, if: :active
```

MoshiKoi: I'm not sure that this works? Presumably this checks after the save, so if a suggested edit is accepted, then `:active` is false and so it doesn't clear the cache. At any rate, I don't think non-active suggested edits ever even get updated anyway.

Taeir: Ah, then I misunderstood, sorry.

MoshiKoi: Cache should be cleared when a suggested edit changes `:active` from true to false (theoretically also false to true, but un-rejecting or un-accepting isn't a thing so this case doesn't matter)

Taeir: Then it should be an if statement for
`saved_change_to_attribute?(:active, from: true, to: false)`
That can either be in the method or as a proc to the if parameter

```suggestion
  after_save :clear_pending_cache, if: Proc.new { saved_change_to_attribute?(:active, from: true, to: false) }
```

(I wrote this on my phone, please double check for mistakes 🙈 )

Taeir: (See https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-saved_change_to_attribute-3F )

MoshiKoi: This works, thanks 👍

PR #935, opened 2023-01-03
Fix tour test for the 3rd time

### LINKED ISSUE #192: Can't add tags in tour
https://meta.codidact.com/questions/277092

A user taking the tour got stuck at the tags, getting the message "The results could not be loaded" when trying to add them.  The user eventually found out how to continue through the tour, but this error was confusing.

### LINKED ISSUE #838: Tour tagset not created/initialized

Taeir:
Also add tests to ensure that the tagset is correctly created by the seeds (because I'm tired of this problem happening)...

Fixes #192 
Fixes #838

### DIFF
--- a/app/controllers/tour_controller.rb
+++ b/app/controllers/tour_controller.rb
@@ -5,7 +5,9 @@ class TourController < ApplicationController
 
   def question1; end
 
-  def question2; end
+  def question2
+    @tagset_id = TagSet.find_by(name: 'Tour')&.id || -1
+  end
 
   def question3; end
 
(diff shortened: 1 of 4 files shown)


Taeir:
I'm currently unable to run the tests locally due to some other issues, but assuming it passes on the CI then this should (after another seed run) really truly fix the problem.

ArtOfCode- (approved):

PR #936, opened 2023-01-03
Fix CI issues

Taeir:
- Fix rubocop errors accidentally merged
- Fix failing test


Taeir:
The tests seem to fail on a missing encryption key in circleci.

@ArtOfCode- that may be why the CI was not running tests on the merge request, since the master key is private info and it should only be accessible on protected branches. I think that for the test run it probably should not need the key and the stage that adds it should be removed.

PR #968, opened 2023-01-19
Decouple codidact from code

Taeir:
This pull request decouples codidact from the codebase in various places. It does so by defining a helper function which checks the domain of the community for `.codidact.[com|org]`. It then disables or changes components on pages based on this setting.

#### TODO
* [ ] Email addresses as settings

### DIFF
--- a/app/helpers/codidact_helper.rb
+++ b/app/helpers/codidact_helper.rb
@@ -1,0 +1,11 @@
+# Provides helper methods to determine whether this site is part of the codidact network
+module CodidactHelper
+  # Whether this server is part of the codidact network.
+  def codidact?
+    Rails.cache.fetch 'is_codidact' do
+      Rails.env.development? ||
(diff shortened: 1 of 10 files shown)


cellio:
Thanks for working on this!  I knew we only got some of them (the easiest ones, naturally) in the previous sweep, and then I lost track of it.

app/helpers/codidact_helper.rb
```diff
@@ -0,0 +1,11 @@
+# Provides helper methods to determine whether this site is part of the codidact network
+module CodidactHelper
+  # Whether this server is part of the codidact network.
+  def codidact?
+    Rails.cache.fetch 'is_codidact' do
```
Taeir, line 5: I wonder whether the Redis cache is going to be faster than checking the string twice. It might be premature optimization.

Issue #956, opened 2023-01-16
Show number of search matches on the results page


cellio:
https://meta.codidact.com/posts/287246

Can we add the number of matches found to the top of the page of search results?  The results are paginated, so the code knows how many results there are (in order to produce the page links at the bottom), so I'm assuming it isn't too hard to expose that information?  With this change, it's more obvious right away if you need to consider refining your search, as opposed to scrolling to the bottom and seeing that there are many many more pages.  (Yes we try to show the most relevant first, but if your search is very specific and you didn't formulate it quite right, this would help you see that.)

duncanscode:
Working on this, should have it ready shortly.

PR #972, opened 2023-01-20
Add post count to Search results page

duncanscode:
Adds a post count to the Search page, based on [issue 956](https://github.com/codidact/qpixel/issues/956).
image

One thing I was uncertain about is whether the text should be Post(s) or Result(s). I figured that because the search returns only posts at this time, it makes sense to label the count "Posts" - which is consistent the count labelling on other pages.

### DIFF
--- a/app/views/search/search.html.erb
+++ b/app/views/search/search.html.erb
@@ -25,11 +25,17 @@
 </div>
 
 <% unless @posts.nil? %>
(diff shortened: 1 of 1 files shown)


cellio:
"Posts" looks like a good choice to me -- as you said, it's consistent.  And I don't think we return a post more than once no matter how many search results it might contain.

Taeir (approved):
Looks good!

PR #980, opened 2023-01-22
Copyable code

### LINKED ISSUE #942: Can we support copying code blocks from posts somehow?
https://meta.codidact.com/posts/287061

In a post with a long code block, it can be hard to copy the whole block, for example so you can paste it into an editor to try to reproduce the problem a question asks about.  You have to click and drag, including through scrolling.

Some sites (I know I've seen this in documentation somewhere) have a "copy" button next to each code block that copies the entire thing to the clipboard.  Is there something we could do, that wouldn't get in the way of reading the code, to support that use case?  A post can have more than one code block, so whatever it is has to be block-level not post-level.  Maybe a right-justified text link, "copy code", above the block and in a smaller size than the body font?  It has to be discoverable and it ought to work on mobile (where we use the full width and thus don't have a margin available like on desktop), and it can't impede reading either the text or the code block in the post.

MoshiKoi:
image

Resolves #942 

It's just a small little js script that inserts a button to copy the code block text, nothing fancy.

### DIFF
--- a/app/assets/javascripts/codeblocks.js
+++ b/app/assets/javascripts/codeblocks.js
@@ -1,5 +1,6 @@
 $(() => {
-  $(".post--content pre").each((_, block) => {
+  $(".post--content pre > code").each((_, block) => {
+    const wrapper = block.parentElement;
     const copyButton = document.createElement('button');
     copyButton.classList.add('button', 'is-muted', 'is-outlined', 'has-float-right');
     copyButton.innerText = 'Copy';
@@ -11,6 +12,6 @@ $(() => {
       }, 2000);
     });
 
-    block.insertAdjacentElement('afterbegin', copyButton);
+    wrapper.insertAdjacentElement('afterbegin', copyButton);
   });
 });
--- a/app/assets/javascripts/codeblocks.js
+++ b/app/assets/javascripts/codeblocks.js
@@ -1,24 +1,16 @@
 $(() => {
-  $(".post--content pre > code").each((_, block) => {
-    const wrapper = block.parentElement;
-
-    const wrapperDiv = document.createElement('div');
-    wrapperDiv.style.position = 'relative';
-
-    const copyButton = document.createElement('button');
-    copyButton.style.position = 'absolute';
-    copyButton.style.right = 0;
-    copyButton.classList.add('copy-button', 'button', 'is-muted', 'is-outlined', 'has-margin-2');
-    copyButton.innerText = 'Copy';
-    copyButton.addEventListener('click', _ => {
-      navigator.clipboard.writeText(block.innerText);
-      copyButton.innerText = 'Copied!';
-      setTimeout(() => {
-        copyButton.innerText = 'Copy';
-      }, 2000);
-    });
-
-    wrapper.insertAdjacentElement('beforebegin', wrapperDiv);
-    wrapperDiv.append(copyButton, wrapper);
+  $(".post--content pre > code")
+    .parent()
+    .each(function() {
+      const content = $(this).text()
+      $(this)
+        .wrap('<div style="position:relative;"></div>')
+        .parent()
+        .prepend($('<button class="copy-button button is-muted is-outlined has-margin-2">Copy</button>')
+          .click(function () {
+            navigator.clipboard.writeText(content);
+            $(this).text('Copied!');
+            setTimeout(() => { $(this).text('Copy'); }, 2000);
+          }))
   });
 });
--- a/app/assets/javascripts/codeblocks.js
+++ b/app/assets/javascripts/codeblocks.js
@@ -1,0 +1,11 @@
+$(() => {
+  $(".post--content pre").each((_, block) => {
+    const copyButton = document.createElement('button');
+    copyButton.innerText = 'Copy';
+    copyButton.addEventListener('click', _ => {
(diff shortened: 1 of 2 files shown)


cellio:
What happens if the first couple lines of code are long?  Does the button overlay the code or move out of the way somehow?

trichoplax:
Looks like the button does overlay the code if the early lines are long enough. What would be the desired solution to this?
- Partially transparent button (my very least favourite solution).
- Button only half overlapping the code block, high enough that it does not obscure the code below, and low enough that it does not obscure the prose above:
  Image alt text
(works but looks horrible).
- Button that is hidden until the user hovers over the code block, so that code is not obscured until then (useless on mobile).
- Button in the unused space to the left of the post that is set aside for the voting buttons:
  Image alt text
(problematic if the code block is at the top of the post in line with the voting buttons).
- Place the button intelligently based on the layout of the code (so if the first few lines are too long, the button appears a few lines lower). This sounds challenging to implement and only delays the problem until someone posts code where all the lines are long.
- Pretend it isn't a problem (which seems to be common on other sites).
- Add a blank line at the top of the code block:
  Image alt text
(nothing is obscured, but it still looks horribly cramped).
- Add 2 blank lines at the top of the code block:
  Image alt text
(less cramped, more wasteful).

I don't have a recommendation, just a lot of options I dislike for different reasons.

trichoplax:
Also worth testing what happens if the code block is tall enough to scroll. Does the button scroll out of sight or remain fixed in place as the code scrolls under it?

trichoplax:
In practice I'd expect the problem of long lines going behind the button to be much more common on mobile, so worth taking more seriously for that reason.

cellio:
Hmm, yeah.  As you say, those aren't ideal options.  And whatever we need to do needs to work and be discoverable on mobile.

What if we put the copy button at the *bottom* of the code block, below not within so it can't overlap code, and add a little vertical space between the end of the code block and the next thing that follows (presumably more text, but could also be tags).  Would that work?  I'd still keep it flush right rather than flush left, so it doesn't interfere with the usual flow of reading a post.

trichoplax:
Something like this?
Image alt text

This does seem less intrusive than a button outside and above the code block. What can we do to make clear that the thing being copied is the code block above? I'm imagining potential confusion when there are two consecutive code blocks, or a code block followed directly by an image.

cellio:
I would attach the button to the bottom of the code block, which should make it clear what's being copied:

mockup

MoshiKoi:
> * Button that is hidden until the user hovers over the code block, so that code is not obscured until then (useless on mobile).

Mobile hover generally works by tap, so this works fine. (See for instance GitHub's code-copy feature on mobile)

trichoplax:
> > * Button that is hidden until the user hovers over the code block, so that code is not obscured until then (useless on mobile).
> 
> Mobile hover generally works by tap, so this works fine. (See for instance GitHub's code-copy feature on mobile)

I understand that this would make the feature usable for people who know what to do, but it would not be discoverable. People would have to be taught that there's an invisible button that only appears if you tap on the code block, rather than having a visible button that lets them work this out without reading a guide.

trichoplax:
Testing out GitHub's code block copy feature for comparison (try it on mobile):

```text
This is a standard code block with no additional markdown for the copy button - it's just added automatically by GitHub.
```

trichoplax:
Looking at the previous comment on mobile, the copy button only appears in the code block after tapping on the code block, which isn't an action a user would generally know to try. On a site without a copy button, long press or double tap would be the ways a mobile user would usually try to highlight text. Neither of these reveal the copy button.

We could make long press and double tap trigger the unhiding of our copy button, but that means it will appear in front of the code and cause problems for someone trying to copy only some of the code (as the button may be in front of the section of the code they are trying to drag the selection markers over).

trichoplax:
> (Extra whitespace around the button is an artifact of my poor graphics skills, not a design suggestion.)

Disclaimer: I don't have graphics skills, I just went to a Codidact page and used the Firefox developer window to look at the HTML, then right click and "Create New Node". It adds a `<div></div>` by default but that can be edited to be a `<button></button>`. Then I added in the classes from Moshi's code and the word "Copy" to give `<button class="button is-muted is-outlined has-float-right">Copy</button>` and the browser does the rest, displaying the button exactly as in Moshi's screenshot. Then that `<button>` tag can be dragged around the HTML and wherever it's dropped, the browser updates the rendered page to show how it would look. (I imagine the developer window is very similar in Chrome too.)

No graphics editor required - just copy and paste, drag and drop.

app/assets/javascripts/codeblocks.js
```diff
@@ -0,0 +1,24 @@
...
+      copyButton.innerText = 'Copied!';
+      setTimeout(() => {
+        copyButton.innerText = 'Copy';
+      }, 2000);
+    });
+
+    wrapper.insertAdjacentElement('beforebegin', wrapperDiv);
+    wrapperDiv.append(copyButton, wrapper);
+  });
+});
```
ArtOfCode-, line 24: suggestion: I'd pick one - either use jQuery, or go vanilla - mixing both feels weird. This feels like it could be simpler with jQuery.

MoshiKoi:
I somehow feel like my jQuery is highly questionable here

PR #985, opened 2023-01-24
site -> community

superplane39:


trichoplax:
I know it would make it a bigger change, but would it be worth also changing the variable names (in addition to the user facing text already changed)? Developers can misinterpret ambiguous words too.

It looks like the usage of the variable name `site` extends outside this file so it wouldn't be a trivial change but the resulting reduced confusion might reduce bugs and speed up development later.

Issue #978, opened 2023-01-22
Add hint about details/summary to formatting help


cellio:
https://meta.codidact.com/posts/287346 asks about adding support for spoilers, but CommonMark doesn't have spoiler syntax.  That's ok because HTML has the details/summary tags, but a lot of people don't know about them.  Let's add a small bit to the formatting help to cover this.  There is a seeded help topic for formatting in this repo, which is why I'm filing this instead of just fixing it on our network.  We'll need to apply the change in both places.

trichoplax:
It might be worth including in the same help section that a blank line is required inside the `details` tag immediately after the closing `summary` tag before the content begins if the content contains any markdown.

There are examples in [Details tag requires an internal blank line to render markdown](https://meta.codidact.com/posts/287279). Those are separated out into individual examples to demonstrate that they are all distinct causes, but for a help section example a few of them could be combined into a single example with and without a blank line.

Without a mention of this it can be non-obvious what is causing the markdown to not render.

PR #986, opened 2023-01-25
added info about using HTML details/summary to formatting help (CommonMark doesn't support)

cellio:
Fixes https://github.com/codidact/qpixel/issues/978.  Sample corresponding on-site fix: https://meta.codidact.com/help/formatting  (I don't know why the pre block at the bottom has those extra lines at the end.)

I pulled the info from https://meta.codidact.com/posts/287346 and https://meta.codidact.com/posts/287279 -- thanks!

### DIFF
--- a/db/seeds/posts/formatting.html
+++ b/db/seeds/posts/formatting.html
@@ -64,3 +64,26 @@
 </blockquote>
 <h3 id="-footnotes-"><strong>Footnotes</strong></h3>
 <p>  To include a footnote in your post, you can use the syntax <code>[^1]</code>. In your main text, include <code>Text[^1] and more text</code>, and at the bottom (where you want to include your footnote), place a line resembling <code>[^1]: footnote text</code>.</p>
+
+<h3 id="-details-"><strong>Hidden Sections</strong></h3>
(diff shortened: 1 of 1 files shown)


trichoplax (commented):
The new code looks good apart from the extra blank lines which you spotted. They are due to the HTML and Markdown renderers confusing each other. There are workarounds that can make it render correctly, but they would be needless extra complication.

Instead I recommend rewriting the section in Markdown instead of HTML. The only part that needs to be HTML is the part that we wish to render as a spoiler section. The part describing how to write it can be all Markdown, which will give no surplus blank lines and will be easier to work with for future editors.

Since the whole file is written in HTML, it gives editors the impression that they must also write in HTML. I've made a pull request #988 to fix this, along with various other pre-existing problems in the file. Your pull request can then use Markdown instead of HTML without looking out of place in a file full of HTML.

For now I have just made the change for this one file, to facilitate your addition of this new section, but also I recommend that all the help files (both database seeds and live help files that may have been edited) be checked to ensure there are no unnecessary HTML versions. It's worth mentioning this to other instances too.

PR #988, opened 2023-01-25
Rewrite formatting help section db seed in Markdown

trichoplax:
Fixes several places where the rendered output is incorrect in the Formatting section of the help, and converts the whole database seed file to Markdown to reduce the risk of future errors and make the help section editable by non-technical site admins.

Running `UPDATE_POSTS=true rails db:seed` in my local dev environment successfully pulled through the Markdown into the Formatting section of the help, and the output was correctly rendered.

Prompted by the subtle bug highlighted by #978 - surplus blank lines are appearing due to the HTML and Markdown renderers having a difference of opinion. I'll omit the details of why, unless requested, because they become irrelevant once the file is Markdown only.

Since converting all the HTML to Markdown makes the diff useless, I've described below those changes that go beyond simply representing the same thing in Markdown.

---

## Links section
- I've manually text wrapped the code block so that it is readable on mobile without side scrolling. Unlike paragraphs, code blocks do not wrap automatically. Previously it required side scrolling even on desktop. Since this is user editable content I don't see an option for making it wider on desktop only, so I've made it narrow enough to be readable on either type of device.
- Old line 15 ended in an apparently unintentional backtick, and the whole line was omitted from rendering. Removing the backtick has fixed this.
- Old lines 18-19 used Markdown syntax inside HTML, even though they are intended to show as rendered, so the help page showed them as raw Markdown. They now show as rendered output as intended.

## Headings section
- Old lines 57 and 59 were preceded by blank lines, which meant the Markdown renderer converted them even though they are inside a code block and not meant to be rendered. They now show as raw Markdown as intended.

## Changes that do not affect appearance
- I've removed the `id` attributes from the headings. Markdown will create these automatically, but with slightly different names. This will only be a problem if they have been linked to from elsewhere (there are no internal links to break).
- I've removed the class attributes. These had values such as "hljs-string" and "hljs-symbol". They did not appear to be having any effect on the rendered output, and appear to have been added by an automated process that did not realise it was dealing with prose rather than code. Removing them allowed finishing the conversion to Markdown.

## Changes that do affect appearance, but that I anticipate being acceptable
- I've replaced all of the `<h3><strong>` headings with `##` headings, which render as `<h2>`. This avoids the semantic problem of skipping `<h2>` (the post title will be `<h1>`), and avoids the use of the semantic `<strong>` tag in a context it was not intended for. This should make the page easier to parse for screen readers. The visual result is slightly larger section headings.

### DIFF
--- a/db/seeds/posts/formatting.html
+++ b/db/seeds/posts/formatting.html
@@ -1,66 +1,94 @@
-<p>Codidact uses the <a href="https://commonmark.org">CommonMark implementation</a> of Markdown for formatting posts.</p>
-<h3 id="-italics-"><strong>Italics</strong></h3>
-<p>  You can <em>italicize</em> a word by placing a single asterisk (<code>*</code>) or single underscore (<code>_</code>) around a word. Either <code>*Codidact*</code> or <code>_Codidact_</code> will render as <em>Codidact</em>.</p>
-<h3 id="-bold-"><strong>Bold</strong></h3>
-<p>  You can <strong>bold</strong> a word by placing a double asterisk or double underscore around that word. Either <code>**Codidact**</code> or <code>__Codidact__</code> will render as <strong>Codidact</strong>.</p>
-<h3 id="-monospace-code-formatting-"><strong>Monospace / code formatting</strong></h3>
-<p>  To make a word monospace, you can surround it with a single backtick (<code>&#96;</code>). This is generally used for code formatting. (Some people use this for emphasis. Don&#39;t do that! Screen readers will read out words formatted as code character by character - use <em>italics</em> or <strong>bold</strong> for emphasis instead.)</p>
-<p>  To mark several lines as code, you can use a code fence. This is three backticks on a newline both on top and at the bottom of your lines of code. Don&#39;t use this for quotes - see below for guidance on quote formatting.</p>
-<h3 id="-links-"><strong>Links</strong></h3>
-<p>  You can generate a link using one of two methods. One way is to encase your link text in square brackets (<code>[ ]</code>) and then the target URL in parentheses (<code>( )</code>). <code>[Codidact](https://codidact.org)</code> will render as <a href="https://codidact.org">Codidact</a>.</p>
-<p>  You can also format a link with the inline text being <code>[Codidact][1]</code>, and then at the bottom of the post include the URL with the format <code>[1]: https://codidact.org</code>. This will look something like this:</p>
-<pre><code>
-  You can check out [<span class="hljs-string">Codidact</span>][<span class="hljs-symbol">1</span>] for an open-source, community-oriented Q&amp;A platform!
+Codidact uses the [CommonMark implementation](https://commonmark.org) of Markdown for formatting posts.
 
-  [1]: https://codidact.org`
-</code></pre><p>  Which will render like this:</p>
-<blockquote>
-    <p>You can check out [Codidact][1] for an open-source, community-oriented Q&amp;A platform!</p>
-    <p>[1]: <a href="https://codidact.org">https://codidact.org</a></p>
-</blockquote>
-<p>  If you&#39;d like to include rollover text for your link, you can put your text, surrounded by quote marks (<code>&quot;</code>), into the URL of your link. For instance, <code>[Codidact](https://codidact.org &quot;The open-source Q&amp;A platform&quot;)</code> will create a link with <code>The open-source Q&amp;A platform</code> as the rollover text.</p>
-<h3 id="-blockquotes-"><strong>Blockquotes</strong></h3>
-<p>  When quoting text from somewhere else, you can use a blockquote to signify that this is not your original text, but a quote from somewhere else. Put a greater than symbol (<code>&gt;</code>), followed by a space, before a paragraph to mark that quote as a blockquote. For instance, <code>&gt; This is a blockquote!</code> will render as a quote:</p>
-<blockquote>
-    <p>This is a blockquote!</p>
-</blockquote>
-<p>  If you want to include a second paragraph as part of the same quote, you have to include another <code>&gt;</code> before each line, like this:</p>
-<pre><code>  &gt; This <span class="hljs-keyword">is</span> a blockquote!
-  &gt;
-  &gt; This <span class="hljs-keyword">is</span> part <span class="hljs-keyword">of</span> <span class="hljs-keyword">the</span> same <span class="hljs-literal">quote</span>!
-</code></pre><p>  This will render as a single blockquote:</p>
-<blockquote>
-    <p>This is a blockquote!</p>
-    <p>This is part of the same quote!</p>
-</blockquote>
-<p>  See also the <a href="/help/referencing">citation and referencing help</a> for guidelines on how to properly reference different quotes from various places.</p>
-<h3 id="-images-"><strong>Images</strong></h3>
-<p>  Embedding images in a post is very similar to the formatting for links (detailed above). The format <code>![alt text](image URL)</code> will embed an image in your post. Please remember to <a href="/help/alt-text">write alt text for your image</a>! Codidact allows you to upload an image directly, or you can embed an external image from image hosting sites such as Imgur.</p>
-<h3 id="-lists-"><strong>Lists</strong></h3>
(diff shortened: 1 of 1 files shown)


trichoplax:
I can't find a reason not to make this change, but to double check I've posted on Collab: [Do seed posts need to be HTML rather than Markdown?](https://collab.codidact.org/posts/287786)

cellio:
These sound like good changes.  I haven't pulled it yet to take a closer look; just a couple things based on your summary.

I've asked in Discord about the reasons for help being HTML, just in case there's some barrier to converting at all.

I'd rather keep the `id` attributes.  We don't know that there aren't links to them, either in other customized help topics or in posts on our communities.  Is there a way to avoid breaking any links that might exist and still switch to Markdown?  (Worst case is we use `<h2>` for these instead of `##`, right?  Not ideal but not fatal.)

Proper heading semantics (1, 2, 3, no skips) sounds like a win to me.  I don't care if the text is a little bigger; it might make the page easier to scan.

trichoplax:
> I'd rather keep the `id` attributes. We don't know that there aren't links to them, either in other customized help topics or in posts on our communities. Is there a way to avoid breaking any links that might exist and still switch to Markdown? (Worst case is we use `<h2>` for these instead of `##`, right? Not ideal but not fatal.)

I was wrong. I thought that ids would be automatically inserted by the Markdown renderer, but it seems not. On GitHub, Markdown files have automatic ids for every heading level, but apparently that's a GitHub thing rather than a Markdown thing.

Options I can see:
1. Go back to using `<h2>` as you suggest, with the old ids
1. Check all the other documents (including all posts in all communities...?) to see if any are using internal links or links to sections of other documents. If none are, we can use `##`. If some are, we can insert ids just for those cases
1. Keep all as `##` without checking. Any existing links will *half* break. They will still take you to the correct page, just not scroll to the relevant section (more of a problem the longer the target page is)

I'm not fond of the first 2 options (for the inconsistency), but they would work. I'd possibly be comfortable with option 3 if we could be reasonably sure there were very few links or that they were only to short documents (maybe if we could run a regex over the database to find links with a `#` in and manually check through the results).

Are you solely concerned about existing links, or would you still want the potential even if no links had been made yet?

cellio:
I'm concerned about existing links, not so much future ones, but you're right that the page would still load so they're only half-broken.

(It would be nice to have in-page anchors, something that's been requested for headings within regular posts too, but that's a larger issue.  If that happens then I hope whatever the solution is would work for help too, but we don't need to block on "somebody might want to direct-link this section".)

ArtOfCode-:
AFAIK, seeds still need to be written in HTML so that they seed and render correctly. @trichoplax sounds like you've found different here, can you confirm?

trichoplax:
@ArtOfCode- At the time of writing I had tested this in development for the Formatting help page only. I've now tested for a few others, and discovered a subtlety I had missed previously (in ***bold italics*** below). The following have been true for each help page I have tested so far:

- Editing the seed file to contain Markdown instead of HTML and then running `UPDATE_POSTS=true rails db:seed` results in the help page rendering correctly.
- Editing that help page through the user interface shows that it is now stored internally as Markdown, and can be edited as such.
- Saving that edit results in the help page rendering correctly.
- ***Once a help page has been edited through the user interface, it is skipped on subsequent runs of `UPDATE_POSTS=true rails db:seed` if the edited version (not the seed) contains text outside of HTML tags (even plain text with no other Markdown).***
  - Adding Markdown in the seed file does not cause it to be skipped on the initial run of `UPDATE_POSTS=true rails db:seed`. It is also not skipped on the next run, even though the site version of the help file now contains Markdown. It is only skipped after an edit that results in the file containing Markdown (even if that's an empty edit to a file that already contains Markdown from the seed file - it won't be skipped until the empty edit is performed).
  - Adding an HTML tag through the user interface does not result in the help page being skipped on the next run of `UPDATE_POSTS=true rails db:seed`.
  - Once skipping of a help page has been triggered once, editing it through the user interface to no longer contain anything outside of HTML tags does not stop it continuing to be skipped.

Note that the filename still ends `.html` - it is only the content which I've been changing to Markdown.

I don't know why the presence of non-HTML content blocks the seeds from updating the help pages, but only after a user edit. If there isn't a good reason for this behaviour, should it be considered a bug? If fixed it looks like it would allow all the help pages to be in Markdown and easily editable.

ArtOfCode-:
@trichoplax IIRC that's intended behaviour - if a help file has been edited by a community, we deliberately don't apply any seeded updates to it to avoid overwriting community-specific changes. Sounds like that's working as intended. Also sounds like replacing everything with Markdown does actually work.

trichoplax:
In the meantime, does this need to block this pull request?
- The problem (user edits containing Markdown block future seeding) already applies to the existing HTML-only seed files.
- I'm expecting that the seed files be used when setting up a new instance, rather than being used to overwrite existing user edited help files.

trichoplax:
> @trichoplax IIRC that's intended behaviour - if a help file has been edited by a community, we deliberately don't apply any seeded updates to it to avoid overwriting community-specific changes. Sounds like that's working as intended.

Ah I see. Then in that case the bug is that the prevention of overwriting community-specific changes is broken when the community-specific changes are solely HTML.

ArtOfCode-:
Agree. No need to block this PR, then - has conflicts but once those are resolved this can be merged.

cellio:
I got a little lost in the thread, but it sounds like this change will not cause overwrites of locally-modified help, right?  I know at least one of our communities has heavily revised the help and others have made smaller changes, and we don't know what other deployments might be doing.

trichoplax:
@ArtOfCode- I've fixed the merge conflict but please hold off merging pending @cellio's concern.

If a community has made edits through the user interface, and if they have kept strictly to only using HTML, then it seems that their changes would be overwritten by `UPDATE_POSTS=true rails db:seed`.

I guess we can either:
- back up their customised wording before deploying
- edit their customised wording to contain at least one Markdown paragraph to prevent overwriting
- hold off merging this change until the bug that allows overwriting customised wording is fixed

I'd lean towards holding off rather than have to think about other instances we might not know about (and therefore can't warn) getting unexpected data loss.

cellio:
I'd lean toward holding off, too.  Is there an issue for the other bug?  If not, could you create one @trichoplax?

trichoplax:
I've raised #992 and changed this pull request to a draft until the bug is fixed.

trichoplax:
This pull request was converted to a draft pending #992 which has now been fixed. However, I have a question before converting this back from draft:

The problem for which #992 was raised was as follows:
- Editing a help post to add text ***outside*** of HTML tags led to that post not being overwritten by future database seeding (as intended).
- Editing a help post to add text only ***inside*** an HTML tag led to that post being overwritten by the next database seeding (which is the problem).

My current understanding is that #1201 (that fixes #992) causes initial revisions to be added during the next database seeding for any help posts that have zero revisions. This means that after the next run of `UPDATE_POSTS=true rails db:seed` every help post will have at least 1 revision (either an added initial revision if it previously had zero, or its existing revisions if already present). This should ensure that any modifications of help posts through the user interface ***after that point*** are detected and not overwritten by future runs of `UPDATE_POSTS=true rails db:seed`.

What would happen to a post that had previously been modified (before this fix) through the user interface Do we need to warn other known instances of potential overwriting of past modifications?
in the way that previously was not protected from being overwritten (that is, a help post that only had extra text added inside HTML tags)?

It seems clear to me that future edits will be protected from being overwritten, but I don't understand the mechanism by which past edits were being overwritten when HTML only, so I don't know whether they will continue to be overwritten after the fix. Would the next database seeding (the one that retroactively adds initial revisions) also overwrite such edits? 

I know my understanding must be lacking somewhere, because I don't know the answer to the following question: If posts were overwritten due to having no revisions, what was stopping posts with edits outside of HTML tags from also being overwritten?

To discuss before converting this back from draft:
- Is there a risk of overwriting help posts that were modified before this fix?
- Do we need to take a copy of any modified help posts from our communities before merging this pull request (since it may involve seeding the database)?
- Do we need to warn other known instances of potential overwriting of past modifications?
- Do we need to suggest that other known instances run the database seeding in order to apply the fix now, so that any future modifications are protected?
- Is there a risk of effect on QPixel instances that we are not aware of and therefore cannot warn?

trichoplax:
If my previous comment leads to needing to warn other instances, would it be best to convert all other help posts to Markdown at the same time, rather than have to warn again for future conversions?

PR #1016, opened 2023-03-17
Update rmagick to version 5

Taeir:
Rmagick was not locked to version 4 in the Gemfile, and it broke when version 5 was released. This explicitly locks the version to version 5 and updates the code to support a minor syntax change.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,7 +70,7 @@ gem 'net-smtp', '~> 0.3'
 gem 'ruby-progressbar', '~> 1.11'
 
 # Image generation
-gem 'rmagick'
+gem 'rmagick', '~> 5.0'
 
 # Payments. Kinda important, y'know.
 gem 'stripe', '~> 5.55'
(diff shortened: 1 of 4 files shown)


ArtOfCode- (approved):

ArtOfCode-:
Merged as e36dcdf2701185e648b25ac94ccda8df80457f15

cellio:
After pulling `develop` I'm getting errors (on my Mac) from `bundle install` here.  I'm trying to get help sorting that out, but I'll also ask: do we _need_ version 5, or do we need to have locked version 4?

Taeir:
The changes from this pr which were not included in the commit are about the specific mac versions to use of the libraries. However, bundler should find the mac specific binaries automatically. When i pull from develop and bundler install I don’t get any issues, what error do you get?

To answer your question, no we don’t need version 5. In the process of figuring out why things were broken when it automatically updated, I already found how to fix and upgrade to the new version. It is generally good to stay up to date with libraries to ensure that we get bug fixes and security updates. Usually only the last major version receives such updates, which means that we would eventually need to upgrade to the new version and apply these changes.

PR #1019, opened 2023-03-17
Fix circle-ci for MRs

Taeir:
- Fixes rubocop reported errors currently in develop
- Fixes tests on merge requests by fixing the CircleCI setup
- Test with both ruby 2.7 and 3.1
- Cache based on the lockfile rather than the Gemfile to ensure that version changes trigger cache changes to ensure consistent builds (perhaps you can configure an expiration for caches if not already set?)


Oaphi (approved):
The changes here all LGTM (the fact that the failing check passes here is evidence enough that it should be, unless there are any objections, merged as soon as possible). @Taeir do you have any specifics you'd like me to comment on other than what we've discussed (apparently, I've missed the request)?

Taeir:
@Oaphi nothing in specific. I just thought that since we discussed it and you gave some good suggestions, you would also be a good fit to review :)

cellio:
I did not notice that this had been approved, sorry.  I wonder if we can get GitHub to generate notifications for approvals.

PR #1030, opened 2023-04-10
Post history links

### LINKED ISSUE #1024: Add links to specific revisions in post histories
https://meta.codidact.com/posts/287547

Items in post histories are already numbered, e.g.:

history screenshot

Can we add the ability to link to specific entries?  One way to do it would be to make the number a link that people can copy.  Another would be to add "copy link" somewhere, like we have on posts.

The behavior when the link is followed would be to go to the history and expand the linked entry, so you'd see the diff without having to click the expander.  That would make it instantly obvious which entry was linked to.

MoshiKoi:
image

Resolves #1024

Adds a button that copies a link to the post history with an anchor to the specific entry. Since I didn't really know what to use for it, it's just `#1`, `#2`, etc. if there are any specific ideas for it, let me know and I'll change it.

### DIFF
--- a/app/assets/javascripts/post_histories.js
+++ b/app/assets/javascripts/post_histories.js
@@ -1,0 +1,3 @@
+$(() => {
+  $(location.hash).parent('details').prop('open', true);
+});
--- a/app/views/post_history/_diff.html.erb
+++ b/app/views/post_history/_diff.html.erb
@@ -1,4 +1,4 @@
-<div class="diff">
+<div class="diff" id="<%= @history.size - index %>">
   <% if before.present? && after.present? %>
     <% if before.is_a?(String) && after.is_a?(String) %>
       <% diff = Diffy::SplitDiff.new(before, after, format: :html) %>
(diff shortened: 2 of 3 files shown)


cellio:
Looks good to me.  We could also make "#5" (e.g.) a link, but we use explicit "copy link" elsewhere, so this is more consistent.

When you follow the link, does it expand the target version for you (so it would show as in your screenshot)?

MoshiKoi:
Yes, It does. Chromium browsers seem to do it automatically, but Firefox doesn't so I had to add in a small script for it.  https://github.com/codidact/qpixel/commit/a844d97d01c3ca9de908904124ef3069ac8982c5

trichoplax (approved):
This works when following a link in a new tab, but there is a subtle problem when pasting a copied link into the same tab.

To reproduce:
- Copy a link to #1 in the edit history of a post
- Paste this into a new tab
- Copy a link to #2 in the edit history of the same post
- Paste this into the same tab as previously
- The page still shows edit #1 expanded, rather than edit #2 (at least in Firefox)

Since it's a minor problem I'm happy to raise a separate bug for it, and approve this pull request since it's a definite improvement.

Issue #1034, opened 2023-04-18
Use `@latest` instead of a version number for Co-Design


trichoplax:
**Is your feature request related to a problem? Please describe.**
Currently the QPixel codebase accesses Co-Design through CDN links which reference a specific version number of Co-Design. This means that every time a new version of Co-Design is released, a manual change must be made to QPixel to update the version number in 4 different places. This is unnecessary manual work, and also introduces a risk of forgetting to update one of the places, or making a typo and bringing in the wrong version number.

**Describe the solution you'd like**
Implement the suggestion in the [Co-Design installation guide](https://design.codidact.org/setup/installation/) and replace the explicit version numbers with `@latest` in the 4 places they appear in the QPixel codebase.

trichoplax:
This was fixed by #1035 which is now live.

PR #1035, opened 2023-04-18
Use @latest in CDN links instead of version number

trichoplax:
Addresses #1034.

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -19,7 +19,7 @@
 <link rel="preconnect" href="https://cdnjs.cloudflare.com" />
 
 <%= stylesheet_link_tag "https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@5.11.2/css/all.min.css" %>
-<%= stylesheet_link_tag "https://cdn.jsdelivr.net/npm/@codidact/co-design@0.12.5/dist/codidact.css" %>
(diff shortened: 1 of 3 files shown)


Taeir (commented):
While this will ensure that QPixel is always up to date with the latest co-design, it does mean that if there are fundamental changes in co-design, those will automatically (and untestedly?) apply to the production instance of QPixel. If changes in both co-design and QPixel are needed for a style change, then this could cause problems.

So the decision is mainly about whether we think that a good idea.

trichoplax:
@Taeir I didn't get around to responding to your comment before this pull request was merged, but I'll add my thoughts now:

I don't have the experience to guess whether the type of change you describe (affecting both Co-Design and QPixel simultaneously) will be the majority or minority of future changes, but since the merged change is easily reversible I'm happy with just trying this out and reassessing after a few future changes.

For any particular change where `@latest` would prevent us merging, or be too great a risk, we can still temporarily fix the version. So if we do get such a change it should hopefully be a minor inconvenience rather than an obstacle.

cellio:
> I don't have the experience to guess whether the type of change you describe (affecting both Co-Design and QPixel simultaneously) will be the majority or minority of future changes, but since the merged change is easily reversible I'm happy with just trying this out and reassessing after a few future changes.
> 
> For any particular change where `@latest` would prevent us merging, or be too great a risk, we can still temporarily fix the version. So if we do get such a change it should hopefully be a minor inconvenience rather than an obstacle.

I escalated the question to Art and he merged it.  As you say, we can reverse it if at some point in the future we need to separate them.

PR #1036, opened 2023-04-21
Close notifications when an item is clicked

### LINKED ISSUE #683: Notification pane doesn't close when the target is on the current page
https://meta.codidact.com/posts/283323

If you click on a notification that's on a different page, you navigate there and the notification pane is closed.  However, if the target is on the same page (such as a new answer to the question you're currently on), following the link takes you to the right place on the page but leaves the notification pane open.  I'm guessing that we never explicitly close it but that going to a new page resets it, in which case this might be solved with a judiciously-placed onClick handler, maybe?  (This is the part where I wave my hands a bit.)  The linked post mentions page redraw, FWIW.

If the notification is near the top of the page, like a new comment under the question, then the pane can cover part of the new content, forcing you to go to the top, close the pane, and then go back down.

MoshiKoi:
Resolves #683 

Automatically closes the notifications panel when an item in it is clicked on.

### DIFF
--- a/app/assets/javascripts/notifications.js
+++ b/app/assets/javascripts/notifications.js
@@ -8,7 +8,7 @@ $(() => {
                 <span data-livestamp="${notification.created_at}">${notification.created_at}</span>
             </div>
             <p><a href="${notification.link}" data-id="${notification.id}"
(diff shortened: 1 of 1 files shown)


Taeir (approved):
Lgtm

PR #1041, opened 2023-05-12
change alt-text placeholder to single token

cellio:
I'm attempting to make the small change requested in https://meta.codidact.com/posts/287653, to make the placeholder for image alt text after upload a single token so you can double-click to select it.  (We want people to replace that string; let's make it a little easier.  As someone who's sometimes picked up a bracket while selecting by click and drag, I sympathize.)

I tried to test this change locally but the upload fails so I don't get as far as generating this text.  I think the upload fails because it assumes S3.  I remember that someone else ran into this problem and added `storage.sample.yml`, which implies I need a `storage.yml`, but I wasn't clear on what to put in it for "test this on my local machine".  (Should we add comments to that file?  Does YAML support comments?)

### DIFF
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -117,7 +117,7 @@ $(() => {
 
   QPixel.addPrePostValidation(text => {
     // This regex catches Markdown images with no or default alt text.
-    const altRegex = /!\[(?:Image alt text)?\](?:\(.+(?!\\\))\)|\[.+(?!\\\])\])/gi;
+    const altRegex = /!\[(?:Image_alt_text)?\](?:\(.+(?!\\\))\)|\[.+(?!\\\])\])/gi;
     if (text.match(altRegex)) {
       const message = `It looks like you're posting an image with no alt text. Alt text is important for ` +
                       `accessibility. Consider adding alt text to the images in your post - ` +
--- a/app/assets/javascripts/posts.js
+++ b/app/assets/javascripts/posts.js
@@ -63,7 +63,7 @@ $(() => {
 
     const $postField = $('.js-post-field');
     const postText = $postField.val();
-    $postField.val(postText.replace(placeholder, `![Image alt text](${data.link})`));
+    $postField.val(postText.replace(placeholder, `![Image_alt_text](${data.link})`));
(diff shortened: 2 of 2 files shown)


cellio:
There is also a [very minor related bug report](https://meta.codidact.com/posts/287656), but I'm not good enough at regex to solve that while I'm in there.

trichoplax:
Until you get your local uploads working, I've tested your changes on my machine and they work perfectly. I can add an image and save it with no problems.

trichoplax:
As the [related bug report](https://meta.codidact.com/posts/287656) is likely to be high cost to fix (I've added a comment there) but very very low benefit (the observed behaviour might not be triggered again for a very long time, and doesn't cause a huge problem beyond temporary confusion when it is triggered), I'd recommend we leave it as a separate bug and not attempt to fix it in this pull request.

trichoplax:
For the `storage.yml` file, mine is an exact copy of the example version. I haven't changed anything and it works fine. Image uploads are saved to the `uploads` folder under a long meaningless file name and are successfully retrieved when viewing the post - the markdown shows, for example: `http://localhost:3000/uploads/p3tzpjkhynmyqs84bnsrbwx60eig`.

Hopefully you should see the same behaviour if you copy `example.storage.yml` and rename the copy to `storgage.yml`.

trichoplax:
Although no changes are required to the `storage.sample.yml` file for development purposes, this is not obvious. In answer to your question, comments can be added to a YAML file using a `#`, either to add a comment at the end of an existing line, or on a separate line of its own. It might be worth adding a comment to explain that for development the file can simply be copied and named `storage.yml`.

cellio:
@trichoplax thank you for all the information and the testing!  I'll spin off a separate low-priority issue for the other bug and add a YAML comment separately.

Taeir (approved):

Taeir:
With regards to the other bug, its not something one can feasibly solve with regex. To fix something like that i would do something in the markdown processing and report the error through that (it knows whether its actually rendering an image), but even that is a lot of work.

cellio:
@trichoplax I couldn't add you directly as a reviewer, but https://github.com/codidact/qpixel/pull/1048 makes the doc change.  If you say it's ok I'll consider that good enough for that particular review. :-)

PR #1048, opened 2023-05-18
Add doc to storage.sample.yml

cellio:
Per https://github.com/codidact/qpixel/pull/1041#issuecomment-1545849997 -- added some comments explaining how to use this in a non-S3 environment.

I tested these instructions in my dev environment.

### DIFF
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -117,7 +117,7 @@ $(() => {
 
   QPixel.addPrePostValidation(text => {
     // This regex catches Markdown images with no or default alt text.
-    const altRegex = /!\[(?:Image alt text)?\](?:\(.+(?!\\\))\)|\[.+(?!\\\])\])/gi;
+    const altRegex = /!\[(?:Image_alt_text)?\](?:\(.+(?!\\\))\)|\[.+(?!\\\])\])/gi;
     if (text.match(altRegex)) {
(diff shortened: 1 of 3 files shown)


cellio:
I guess I need to rebase this on `develop` but I'm not quite sure how to do that without making a mess.  The first commit her was already merged; didn't realize my local environment wasn't proper.

trichoplax:
The documenting comment looks good to me.

trichoplax:
As for the rebasing, the [git rebase documentation](https://git-scm.com/docs/git-rebase) suggests that the redundant commit (that is present in both `develop` and your new branch) should automatically be discarded if you rebase, so you shouldn't need a complicated command that tells git which commits to keep.

You should be able to just do the following:

```text
git pull
git checkout "cellio/storage-yaml-doc"
git rebase develop
```

You can then use `git log` to check that you have the commits you expect before you push. After that, this pull request should only have the documentation commit, and not the other commit that is already in develop.

Worst case, you end up with an extra commit and we'll resort to the longer form of `git rebase`, but it sounds like that won't be necessary.

cellio:
I rebased and now there are _more_ diffs (that I did not make).  Y'know what?  This is a trivial change; I'm not going to try to fix this branch when I can just reimplement it.

cellio:
New clean PR: https://github.com/codidact/qpixel/pull/1050 .

PR #1055, opened 2023-05-30
collapse filter on search results page

cellio:
https://meta.codidact.com/posts/288181

On search results, collapse the filter widget by default and also move it to right above the list of posts (for consistency).  This puts the guidance about search grammar right next to the textbox where you can type it, and puts the filters right next to the posts like on the category page.

Thanks to @MoshiKoi for helping me find the right place to update.

This is what that page looks like now:

screenshot

### DIFF
--- a/app/views/search/_widget.html.erb
+++ b/app/views/search/_widget.html.erb
@@ -8,6 +8,4 @@
       <%= submit_tag 'Search', class: 'button is-medium is-outlined', name: nil %>
     </div>
   </div>
-  <h2>Filters</h2>
-  <%= render 'filters' %>
 <% end %>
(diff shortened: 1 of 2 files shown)


MoshiKoi:
it appears to break the filters.

image

Selecting a filter doesn't fill in anything, and none of the buttons work when clicked on either

cellio:
Uh oh.  That's not good.

cellio:
I can't reproduce what MoshiKoi is seeing (help, please!), and I'm also mystified by how relocating some static text and an invocation like this would break it.  I'd like to learn from this; I'm stuck.

MoshiKoi:
Woops, misclicked (was going to comment that the most recent commit fixed the issue I mentioned)

MoshiKoi (approved):

PR #1057, opened 2023-06-04
Move edit history ids up a level and close all but the relevant edit

### LINKED ISSUE #1056: Edit history does not expand if URL is updated in Firefox
**Describe the bug**
Following #1030 there is a subtle problem when pasting a copied edit history link into the same tab.

Since it's a minor problem I'm raising it here as a separate issue, to avoid holding up approval of the mentioned pull request since it's a definite improvement.

**To Reproduce**
Steps to reproduce the behavior ~(will only work after #1030 is released, or in development)~:
1. Copy a link to edit number 1 in the edit history of a post
1. Paste this into a new tab
1. Copy a link to edit number 2 in the edit history of the same post
1. Paste this into the same tab as previously
1. The page still shows edit number 1 expanded, rather than edit number 2 (at least in Firefox)

In Chrome the correct edit number is expanded, but the others are not collapsed if they were already expanded before updating the URL.

**Expected behavior**
Requirement: The edit corresponding to the number in the URL should be automatically expanded when the URL is changed.

Optional behaviour (to be discussed): Any other open edits should be automatically collapsed, when the URL is changed.

**Desktop**
 - OS: Fedora 38 (no reason to think this will be different for other operating systems)
 - Browser: Firefox 113.0.1 (64 bit)

trichoplax:
Fixes #1056.

- Moves the edit history ids up to the `<details>` tags
- Amends the JavaScript to expand the relevant edit in the edit history every time the URL hash changes, rather than only on page load
- Amends the JavaScript to collapse all the other edits in the edit history every time the URL hash changes

Behaviour is now consistent between Firefox and Chromium.

### DIFF
--- a/app/assets/javascripts/post_histories.js
+++ b/app/assets/javascripts/post_histories.js
@@ -1,7 +1,7 @@
 $(() => {
   const openRelevantEditOnly = () => {
-    $("details.history-event").attr('open', false);
-    $(location.hash).attr('open', true);
+    $("details.history-event").prop('open', false);
+    $(location.hash).prop('open', true);
   }
 
   window.addEventListener("hashchange", openRelevantEditOnly);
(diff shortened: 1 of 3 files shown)


MoshiKoi (approved):
Looks good to me

Issue #1000, opened 2023-02-02
Inform user that 15 characters are required for some flag types


trichoplax:
**Is your feature request related to a problem? Please describe.**
When flagging a post, some of the radio buttons result in the "Details" field being required to have at least 15 characters, but the user is only informed of this after trying to submit the flag.

![Flag modal for a question showing 5 radio buttons](https://meta.codidact.com/uploads/r1at8ztfdlzqtk3b9h9zoxthviai)

**Describe the solution you'd like**
When the details are optional, the field can remain as currently:

> Details?
> You might want to add some details to your flag.

When the details are required, drop the question mark and replace the second line with the character requirement:

> Details
> (minimum 15 characters)

**Describe alternatives you've considered**
We considered reducing or dropping the character requirement but that could cause problems for some flag types (specifically "needs author's attention" as that leads to a comment being created) so for now we can simply keep the user better informed.

**Additional context**
[Discussion on Discord](https://discord.com/channels/634104110131445811/668890340199235613/1070488182552002591)

cellio:
Some reactions require comments which also must be at least 15 characters, and right now we fail with a 422 that the user might not see.  Is there a common path here?  That is, thing that requires comment -> enforce comment length -> proceed ?  Or do we have to check each case separately?

cellio:
Now that minimum flag length is 1, reaction comments are handled, and the "needs author's attention" flag is gone, have we covered everything?

PR #1066, opened 2023-06-07
Show errors to user when adding a reaction fails

### LINKED ISSUE #1004: Reaction with comment with too few characters gives no feedback to user
**Describe the bug**
When a user tries to add a reaction, some reaction types require a comment and some do not.

- For a reaction that requires a comment, trying to enter zero characters in the comment field leads to a red message "This reaction type requires a comment with an explanation" (as intended).
- For a reaction that does not require a comment, trying to enter zero characters in the comment field leads to a successful reaction (as intended).

However, trying to enter more than zero but fewer than 15 characters results in no action when the user presses "Add reaction". The dialogue box remains open and there is no feedback to tell the user why nothing is happening. This happens for both types of reaction (those that require a comment and those that do not).

**To Reproduce**
1. Go to a question page that contains an answer
2. Click on "react" on an answer
3. Select a reaction type
4. Enter a comment from 1 to 14 characters
5. Click on "Add reaction"
6. Observe that the dialogue box remains open (the developer window network tab shows a 422 HTTP response, but no feedback is shown to the user)

**Expected behavior**
Where a comment is required, there should be a clear message to the user indicating that 15 characters are required. Ideally this should be clear *before* pressing "Add reaction", but at the very least there should be a feedback message when the user tries to submit a comment from 1 to 14 characters.

Where no comment is required, does it make sense to impose a 15 character minimum if zero characters is also acceptable? There should either be a clear message indicating that a comment, if included, must include at least 15 characters, or otherwise there should be no character requirement.

MoshiKoi:
Resolves #1004 

image

I also added placeholder text for the textarea

image

### DIFF
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -33,13 +33,19 @@ class ReactionsController < ApplicationController
 
     reaction = Reaction.new(user: current_user, post: @post, reaction_type: reaction_type, comment: comment)
 
-    ActiveRecord::Base.transaction do
-      thread&.save!
-      comment&.save!
-      reaction.save!
-    end
+    begin 
+      ActiveRecord::Base.transaction do
+        thread&.save!
+        comment&.save!
+        reaction.save!
+      end
 
-    render json: { status: 'success' }
+      render json: { status: 'success' }
+    rescue
+      render json: { status: 'failed', message: "Could not create comment thread: #{(thread&.errors&.full_messages \
+                                                        + comment&.errors&.full_messages \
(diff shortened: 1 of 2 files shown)


cellio:
Does this also resolve https://github.com/codidact/qpixel/issues/1000?

trichoplax:
The code changes don't affect the flag dialogue, so this doesn't resolve #1000 but that was a lesser problem.

With #1004 there was no feedback to the user even *after* attempting to submit a too-short comment for reactions. With #1000 it is only about warning the user up front - they already get a red message *after* trying to submit a too-short comment when flagging, just no indication of how many characters are required beforehand.

PR #1070, opened 2023-06-13
Allow reactions without votes

luap42:
Show the sidebar with a react button even if no votes are not allowed on this post type


cellio:
(Sorry reviewers -- didn't notice it's a draft.  Now removed.)

PR #1079, opened 2023-06-20
Fix filters getting desynchronized

MoshiKoi:
Meta issue https://meta.codidact.com/posts/288530

Fixes filters getting desynchronized by removing the broken browser cache logic.

### DIFF
--- a/app/assets/javascripts/qpixel_api.js
+++ b/app/assets/javascripts/qpixel_api.js
@@ -253,10 +253,7 @@ window.QPixel = {
   },
 
   filters: async () => {
-    if (this._filters == null && localStorage['qpixel.user_filters']) {
-      this._filters = JSON.parse(localStorage['qpixel.user_filters']);
-    }
-    else if (this._filters == null) {
+    if (this._filters == null) {
       // If they're still null (or undefined) after loading from localStorage, we're probably on a site we haven't
       // loaded them for yet. Load via AJAX.
       const resp = await fetch('/users/me/filters', {


trichoplax (approved):
## Summary
I recommend we merge this pull request as is and raise separate issues for the other problems observed, to avoid delaying this fix.

## Detail
### New problems introduced by this pull request
- User filters are always fetched from the server, making page loads slightly slower. I recommend proceeding anyway, as this quick fix is still a big improvement over showing incorrect user filters.
- This also makes the code that stores user filters in localStorage redundant. I recommend leaving this redundant code in because other fixes will make it relevant again soon (see below).

### Existing problems with filters spotted in this review
- We make a call to attempt to fetch user filters when there is no user signed in (line 208). See the solution used for preferences (line 161).
- Filters needs a user specific localStorage key, similar to that used by preferences (line 345).

### Similar problem affecting preferences spotted in this review
- Preferences has a similar caching problem - preferences on one device will not be reflected in another device because both will always trust their own localStorage. I recommend storing a hash of the preferences in the preferences object, which will mean it is automatically stored both serverside and clientside. Then we can put the hash in the HTML and reload the preferences from the server when the hashes in the HTML and localStorage do not match.

This same approach can be applied for filters.

### Recommendation
I recommend raising all of these as separate issues and proceeding with this pull request without changes, as it already fixes one problem.

Let's not delay one fix for the sake of other fixes - they can each be deployed separately so the users get fixes as soon as possible.

cellio:
Thanks to both of you.  @trichoplax can you file follow-on issues as needed?

trichoplax:
I'm not at a computer at present but I have reminders set for raising the issues mentioned.

PR #1094, opened 2023-07-04
Make follow/unfollow icons consistent

### LINKED ISSUE #1067: Inconsistency in comment thread following icons
**Describe the bug**
Pull request #1021 much improved the clarity of the "Follow new" button under posts, but the equivalent button on an individual comment thread page is now inconsistent as it still uses the old "+" and "-" icons.

**To Reproduce**
1. From a post with a comment thread, expand the thread and click "show thread" to open it in a new page
1. See "+follow" or "-unfollow" at the top right of the comment thread

**Expected behavior**
The new bell / slashed bell icons introduced by #1021 along with the matching text should be used for consistency.

superplane39:
Resolves #1067

### DIFF
--- a/app/views/comments/thread.html.erb
+++ b/app/views/comments/thread.html.erb
@@ -47,13 +47,13 @@
           <a href="#" class="widget--header-link js--unrestrict-thread" data-action="follow"
             data-thread="<%= @comment_thread.id %>" title="You are following this thread and will be notified of every response. You can unfollow at any time."
             role="button" aria-label="Unfollow this thread">
-            <i class="fa fa-minus fa-fw"></i>unfollow
+            <i class="fas fa-fw fa-bell-slash"></i>unfollow
           </a>
         <% else %>
           <a href="#" class="widget--header-link js--restrict-thread" data-action="follow"
             data-thread="<%= @comment_thread.id %>" title="Follow this thread to be notified of every response."
             role="button" aria-label="Follow this thread">
-            <i class="fa fa-plus fa-fw"></i>follow
+            <i class="fas fa-fw fa-bell"></i>follow
           </a>
         <% end %>
       <% end %>


app/views/comments/thread.html.erb
```diff
@@ -47,13 +47,13 @@
           <a href="#" class="widget--header-link js--unrestrict-thread" data-action="follow"
             data-thread="<%= @comment_thread.id %>" title="You are following this thread and will be notified of every response. You can unfollow at any time."
             role="button" aria-label="Unfollow this thread">
-            <i class="fa fa-minus fa-fw"></i>unfollow
+            <i class="fas fa-fw fa-bell-slash"></i>unfollow
```
trichoplax, line 50:
```suggestion
            <i class="fas fa-fw fa-bell-slash"></i> unfollow
```

app/views/comments/thread.html.erb
```diff
@@ -47,13 +47,13 @@
...
           </a>
         <% else %>
           <a href="#" class="widget--header-link js--restrict-thread" data-action="follow"
             data-thread="<%= @comment_thread.id %>" title="Follow this thread to be notified of every response."
             role="button" aria-label="Follow this thread">
-            <i class="fa fa-plus fa-fw"></i>follow
+            <i class="fas fa-fw fa-bell"></i>follow
```
trichoplax, line 56:
```suggestion
            <i class="fas fa-fw fa-bell"></i> follow
```

trichoplax (changes requested):
Looks good. I've suggested a minor change - to add spaces between the icon and the wording in each case for consistency with the icons on the main post page.

There are other inconsistencies between the follow button on this page and the follow button on the main post page, such as starting with upper case or lower case, and being styled as a button or a link, but each of those is consistent with the surrounding buttons/links, so I think that's best left for a bigger change in future. The spaces are the only thing I'd change in this issue.

trichoplax (approved):

Issue #1109, opened 2023-07-10
Title text shows on unrelated parts of the posts list pages


trichoplax:
**Describe the bug**
see [Title text (tooltip/hovertext) shows on unrelated parts of the page](https://meta.codidact.com/posts/287245).

That post raises the problem that hovering anywhere on the same row as the post count shows the post count as the title text, even over the subscribe button or the sorting buttons.

It has also been highlighted during review of #1108 that this will need to be fixed for the user posts list too.

**To Reproduce**
1. Go to either 
    - the category level posts list: https://meta.codidact.com/
    - the user posts list: https://meta.codidact.com/users/53890/posts (once #1108 is merged)
1. Hover over the post count at the top left, and correctly see the number of posts.
1. Hover over other places on the same row (the other buttons, or even the wide blank space between them), and incorrectly see the number of posts.

**Expected behavior**
The post count should only appear when hovering over the post count (the number or the word "posts"), not anywhere else on that row.

It may seem odd to have the number of posts appear when hovering over the number of posts, but for larger numbers the reason becomes clear: the number of posts is shown abbreviated, "1.2k", while the text on hover is the full number, "1234".

PR #1111, opened 2023-07-11
fixed tooltips at tops of post lists

cellio:
Fixes https://github.com/codidact/qpixel/issues/1109 on:

- category page
- user posts page
- search-results page

There is a separate (pre-existing) placement problem on the search page, which I'll record separately.

### DIFF
--- a/app/views/categories/show.html.erb
+++ b/app/views/categories/show.html.erb
@@ -12,19 +12,20 @@
 
 <% post_count = @posts.count %>
 <div class="has-color-tertiary-500 category-meta">
-  <span title="<%= post_count %> posts">
-    <%= short_number_to_human post_count, precision: 1, significant: false %>
-    <%= 'post'.pluralize(post_count) %> &nbsp;
+  <span>
+    <span title="<%= post_count %> posts">
+      <%= short_number_to_human post_count, precision: 1, significant: false %>
+      <%= 'post'.pluralize(post_count) %> &nbsp;
+    </span>
 
-  <% if current_user&.is_admin %>
-  <%= link_to 'Edit Category', edit_category_path(@category) %>
-  <% end %>
+    <% if current_user&.is_admin %>
+    <%= link_to 'Edit Category', edit_category_path(@category) %> &nbsp;
+    <% end %>
 
-
-  <%= link_to 'Subscribe',
-      new_subscription_path(type: 'category', qualifier: @category.id, return_to: request.path),
-      class: 'button is-outlined',
-      title: 'Get email notifications for new top-level posts' %>
+    <%= link_to 'Subscribe',
+	new_subscription_path(type: 'category', qualifier: @category.id, return_to: request.path),
+	class: 'button is-outlined',
+	title: 'Get email notifications for new top-level posts' %>
   </span>
       
   <div class="button-list is-gutterless has-margin-2">
--- a/app/views/categories/show.html.erb
+++ b/app/views/categories/show.html.erb
@@ -12,22 +12,21 @@
 
 <% post_count = @posts.count %>
 <div class="has-color-tertiary-500 category-meta">
-  <div title="<%= post_count %> posts">
+  <span title="<%= post_count %> posts">
     <%= short_number_to_human post_count, precision: 1, significant: false %>
-    <%= 'post'.pluralize(post_count) %> 
-  </div>
-  <div>
-    <% if current_user&.is_admin %>
-      <%= link_to 'Edit Category', edit_category_path(@category) %>
-      &middot;
-      <% end %>
-  </div>
-  <div title="Get email notifications for new top-level posts">
-    <%= link_to 'Subscribe',
-                new_subscription_path(type: 'category', qualifier: @category.id, return_to: request.path),
-                class: 'button is-outlined' %>
-  </div>
+    <%= 'post'.pluralize(post_count) %> &nbsp;
 
+  <% if current_user&.is_admin %>
+  <%= link_to 'Edit Category', edit_category_path(@category) %>
+  <% end %>
+
+
+  <%= link_to 'Subscribe',
+      new_subscription_path(type: 'category', qualifier: @category.id, return_to: request.path),
+      class: 'button is-outlined',
+      title: 'Get email notifications for new top-level posts' %>
+  </span>
+      
   <div class="button-list is-gutterless has-margin-2">
     <%= link_to 'Activity', query_url(sort: 'activity'),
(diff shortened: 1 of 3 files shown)


app/views/categories/show.html.erb
```diff
@@ -11,16 +11,18 @@
 <% end %>
 
 <% post_count = @posts.count %>
-<div class="has-color-tertiary-500 category-meta" title="<%= post_count %>">
-  <div>
+<div class="has-color-tertiary-500 category-meta">
+  <div title="<%= post_count %> posts">
```
trichoplax, line 15: This div and the 2 that follow it (the post count, the Edit Category link, and the Subscribe button) need to be all contained within a single div. This time an extra layer of containment really is the answer. Without it, the styling of the next div up causes them to spread out to fill the available horizontal space (so the further you zoom out, the further they move apart from each other, rather than staying all on the left). I hadn't anticipated this when I suggested this would be an easy issue, but this will fix the spreading out problem.

app/views/categories/show.html.erb
```diff
@@ -11,16 +11,18 @@
...
     <%= short_number_to_human post_count, precision: 1, significant: false %>
-    <%= 'post'.pluralize(post_count) %> &middot;
-
+    <%= 'post'.pluralize(post_count) %> 
+  </div>
```
trichoplax, line 18: Once the extra layer of containment has been added, this inner div around just the post count will cause it to be on a new line, rather than the same line as the Edit Category link and the Subscribe button.

This is because a div is a block element that defaults to being on its own line. If you change it to a span (which is an inline element) it will default to being on the same line as its neighbours.

app/views/categories/show.html.erb
```diff
@@ -11,16 +11,18 @@
...
+  <div>
     <% if current_user&.is_admin %>
       <%= link_to 'Edit Category', edit_category_path(@category) %>
       &middot;
-    <% end %>
-
+      <% end %>
+  </div>
```
trichoplax, line 24: The same problem happens with this div - it will default to being on its own line. It turns out it isn't needed so it can simply be removed which will solve the problem. Also a slight indentation problem - the `<% end %>` should be aligned with the `<% if ... %>` (and this whole Edit Category section will need one level of indentation less once its div is gone).

app/views/categories/show.html.erb
```diff
@@ -11,16 +11,18 @@
...
+  <div title="Get email notifications for new top-level posts">
```
trichoplax, line 25: I like the addition of title text here, but using a div is going to cause the same problem of defaulting to being on its own line. Instead the title can be added directly to the `<% link_to` element, and the div can be removed.

As this bit is Ruby rather than HTML, it will need to be `title:` rather than `title=`, and the style seems to be to use single quotes rather than double quotes for the Ruby sections. So it will end up looking something like this:

```ruby
    <%= link_to 'Subscribe',
                new_subscription_path(type: 'category', qualifier: @category.id, return_to: request.path),
                class: 'button is-outlined',
                title: 'Get email notifications for new top-level posts' %>
```

trichoplax (changes requested):
I like the addition of "posts" to the title text so it's no longer a bare number with no context.

There's a subtle problem I hadn't anticipated which only shows up when zoomed out or on a large screen. I've added some comments explaining how to fix it by adding a div and removing a couple of others.

I've only added comments to the Categories file for now - I wanted to get the feedback to you early. The other files may need similar changes if zooming out makes the buttons spread out. I'll check the other 2 files when I come back to this tomorrow.

app/views/categories/show.html.erb
cellio: When I made these changes, the buttons all moved onto a new line.  I'm going to work through this again and see if I can figure out what's going on, but this didn't work on a first pass.

cellio:
I've made some changes to the category page only (so far -- best to fully understand before doing the others).  This is what it currently looks like for an admin:

screenshot

and an anonymous user:

screenshot

This seems to do the right thing when zooming way out.  

I haven't seen the buttons flow to a second line, though some comments I haven't addressed yet suggest I should.  Can you review these changes and help me understand the unaddressed comments?  Thanks.

app/views/categories/show.html.erb
```diff
@@ -11,21 +11,23 @@
 <% end %>
 
 <% post_count = @posts.count %>
-<div class="has-color-tertiary-500 category-meta" title="<%= post_count %>">
-  <div>
-    <%= short_number_to_human post_count, precision: 1, significant: false %>
-    <%= 'post'.pluralize(post_count) %> &middot;
+<div class="has-color-tertiary-500 category-meta">
+  <span>
```
trichoplax, line 15: This containing span is the main difference from my suggested approach. It doesn't seem to cause any problems so I see no reason not to stick with the span. (No change required here.)

app/views/categories/show.html.erb
```diff
@@ -11,21 +11,23 @@
...
+    <span title="<%= post_count %> posts">
+      <%= short_number_to_human post_count, precision: 1, significant: false %>
+      <%= 'post'.pluralize(post_count) %> &nbsp;
```
trichoplax, line 18: The non breaking space here is a change from the middot. I don't have a preference between them myself, but just to clarify: The middot I suggested removing in the previous pull request was because there was no other button to separate the post count from, so it was serving no purpose. (No change required here - just explaining my reasoning.)

trichoplax (approved):
Looks good - all 3 files are working well now.

I was only seeing the spreading out problem before an extra layer of containment was added. Your use of a span has solved this.

I was only seeing the items moving onto the next line when the inner divs were present. I can still recreate this in your version of the code by adding a div around the Edit Category link - it causes the 3 items to be arranged vertically rather than horizontally. Since you don't have that div anymore, there is no more problem.

I can't see any layout problems at different widths (such as mobile phone width) that weren't there already (the Random button doesn't quite fit on the same line as the other sorting buttons on a narrow mobile screen). This isn't relevant to this pull request so I'm happy to approve.

app/views/categories/show.html.erb
cellio: Oh, I meant to comment on that.  I removed the middot because it was unnecessary.  Then things were too close together to my eye (the "edit" link was too close to the "N posts" text, in particular), so I added the spaces.

I never knew before this PR about the middot or that those stray specs I thought were monitor cruft or vision problems were intentional.  They seem too small to do anything other than make this reader wonder what's wrong. :-)   In this case we have "text - link - button", three different kinds of things; putting a tiny dot between the first two but not the second two would be inconsistent but putting one between the link and the button looks especially weird.  So I simplified.

PR #1124, opened 2023-07-16
added some of the pre-search 'hints' to the results page 

cellio:
Before you start a search, the pop-up gives you a summary of some common search operations.  When you get your search results and want to refine your search, though, it's not also on the results page (though you could go up to the "search" button again to get it, but that would mean starting over).

Adding the entire block would add too much stuff to get through before you get to the results, particularly on a phone.  I considered making it collapsible, but that's easier to miss.  I opted for a terse summary in a paragraph (rather than the mult-line layout) -- enough to jog memory, and the link to the full help is right there.

screenshot

Requested in https://meta.codidact.com/posts/288175.


cellio:
[I learned via ](https://meta.codidact.com/posts/288178) points out that `m` is minutes, not months.  I've updated the PR, but I didn't make a new screenshot.

ArtOfCode- (approved):

PR #1127, opened 2023-07-17
SSO Multi Community - Add signing in through other community

Taeir:
For the way SSO integration was previously set up, all SSO requests will end up back at the same community. However, if there are multiple communities and they are not hosted on the same domain (common), this will only allow one community to actually use SSO sign in (due to how devise_saml is configured and designed / due to common restrictions on SAML requests).

The changes in this PR allows a user on one community to sign in through an SSO linked to a different community. The community to which the SSO is linked is defined as the "base community", and is determined based on the assertion consumer service url (`base_community` in `SamlSessionsController`). If this is unsuccessful, the first community is considered as the base.

The authentication flow is as follows:
1. (Requesting) User clicks on a SSO Sign in button -> goes to (GET) `/users/saml/new` (`SamlSessionsController#new`)
2. (Requesting) If user at base community, go to step 6.
3. (Requesting) Otherwise -> redirect to (GET) `{base_community}/users/saml/sign_in_request_from_other/{requesting_community_id}` (`SamlSessionsController#sign_in_request_from_other`)
4. (Base) Add cookie `signing_in_for` which holds (encrypted) the community id where the user came from and is trying to sign into.
5. (Base) If already signed in, go to step 9
6. (Base) (Otherwise) redirect to SAML IDP (actual SSO login)
7. (Base) User is redirected from SSO Login to POST `{base_community}/users/saml/new` (`SamlSessionsController#create`)
8. (Base) User is redirected to `{base_community}/users/saml/after_sign_in_check` (`SamlSessionsController#after_sign_in_check`)
9. (Base) If no `signing_in_for` cookie is present, go to step 12
10. (Base) Otherwise (`signing_in_for` cookie is present) redirect to `{requesting_community}/users/saml/sign_in_return_from_base` (`SamlSessionsController#sign_in_return_from_base`) with encrypted info about the user who signed in
11. (Requesting) Decrypt user info and sign in the user
12. (Requesting) redirect user to whatever location they should go to after a normal sign in.
  
Some additional notes:
- The `after_sign_in_check` endpoint exists only because clients don't like sending cookies whenever they are doing a cross-origin request. So after logging in on the SSO, they are redirected back to the application. At this moment, the client does not send cookies. According to the specification it is possible by setting some attributes, but only when using https (i.e. won't work in development). Additionally, some browsers just don't do this (to prevent tracking across sites?), so it is not reliable. The second endpoint thus exists to do a redirection within the same origin to get the client cookies.
- While this seems convoluted, the user should notice very little of the extra redirects. With low latency it is not noticeable. With higher latency, the sign-in procedure may just take a little longer.
- Previous flow was steps 1, 6, 7 and 12.
- When only using one community, we skip steps 3, 4, 5, 10 and 11 (So only 2, 8 and 9 are new)
- When a user is already signed in on the base community, devise_saml normally just redirects the user to the root rather than bringing them to the SSO. To work around this, step 5 exists. This means that a user can sign in on another community without authentication as long as they are signed in at the base, which is potentially unwanted. An alternative is to sign the user out at the base community, then let devise redirect them to the SSO, and then they will be signed in at the base community and the target community. Issue with that is that if the user does not sign in, they will be "suddenly" signed out at the base community.

### DIFF
--- a/app/controllers/users/saml_sessions_controller.rb
+++ b/app/controllers/users/saml_sessions_controller.rb
@@ -171,7 +171,8 @@ class Users::SamlSessionsController < Devise::SamlSessionsController
   # @param user [User]
   def encrypt_user_info(user)
     len = ActiveSupport::MessageEncryptor.key_len - 1
-    crypt = ActiveSupport::MessageEncryptor.new(Rails.application.secrets.secret_key_base[0..len])
+    key = Rails.application.secrets.secret_key_base || Rails.application.credentials.secret_key_base
+    crypt = ActiveSupport::MessageEncryptor.new(key[0..len])
     crypt.encrypt_and_sign(user.id, expires_in: 1.minute)
   end
 
@@ -179,7 +180,8 @@ class Users::SamlSessionsController < Devise::SamlSessionsController
   # @param data
   def decrypt_user_info(data)
     len = ActiveSupport::MessageEncryptor.key_len - 1
-    crypt = ActiveSupport::MessageEncryptor.new(Rails.application.secrets.secret_key_base[0..len])
+    key = Rails.application.secrets.secret_key_base || Rails.application.credentials.secret_key_base
+    crypt = ActiveSupport::MessageEncryptor.new(key[0..len])
     crypt.decrypt_and_verify(data)
   end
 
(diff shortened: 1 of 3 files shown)


cellio:
@ArtOfCode-  I added the usual reviewers, but I'm also requesting from you directly because of your particular knowledge here.

ArtOfCode-:
I can't say I'm particularly knowledgeable about SAML and associated stuff, but the explanation @Taeir has here seems sane to me and the code looks like it does exactly that, so I'm happy.

ArtOfCode- (approved):

PR #1134, opened 2023-07-21
Add close reason as comment to history

### LINKED ISSUE #215: Record close reason in history
https://meta.codidact.com/questions/278719

When a question is closed, the close reason is not recorded in the history.  Knowing what the problem was, particularly when the question is subsequently reopened, is helpful and I hope not hard.  We add the reason to the post itself; can we also add it to the history?

Taeir:
Comments are now rendered as markdown, to allow inclusion of e.g. links.

This MR was inspired by @MoshiKoi 's version (#863), with the main change relative to that version being that I added markdown support to be able to create links. This also means that markdown can be used when the user provides a comment themselves. However, most markdown is again filtered by the PostHistoryScrubber, which only accepts the absolute basics.

image

Fixes #215

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -271,13 +271,15 @@ class PostsController < ApplicationController
       end
 
       duplicate_of = Question.find(params[:other_post])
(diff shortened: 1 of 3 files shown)


Taeir:
This approach will require changes to work together with #1141, will probably rework it into there.

PR #1141, opened 2023-07-21
Rollback history (#340)

### LINKED ISSUE #215: Record close reason in history

### LINKED ISSUE #340: Support rollbacks in edit history
https://meta.codidact.com/questions/279795

If you need to undo an edit, you have to do it by hand.  Let's add rollback support to make that easier.

Suggested presentation: in the edit history for a post, if and only if you have the edit ability, show something with each non-current revision that says "roll back to this version".  If the user clicks that, prompt for an edit reason (like for other edits) to be appended to the standard notice: "rolled back to version N (version timestamp)".

Notes:

1. I'm suggesting we gate it by the edit ability to avoid having rollbacks as suggested edits, which seems like it could be confusing to review.  If you can't edit you can't roll back, sorry.  You can always leave a comment or raise a flag if there's something that needs attention.

2. The edit history should clearly show that an edit was a rollback no matter what the person does or doesn't type, hence the standard text there (which can be supplemented).  I'm suggesting that we include the version number of the edit being rolled back to, for easier history viewing, and the timestamp of that revision, so it'll be more obvious if somebody just rolled a post back by a year or something.

3. Yeah, that would require revision numbers in the edit history, which we don't have.  Can we?

Taeir:
# Rollback History
Adds support for history item rollbacks for post edits, closing/reopening and deleting/restoring. **Please note: this does not restore state to a previous version, but undoes certain actions in the history.**

Implementation was quite tricky, so please thoroughly review for mistakes or missing things (e.g. permissions).

## Usage details
Something can only be rolled back (undone) if its changes are still present. E.g. you cannot rollback (undo) something that adds a tag that already is no longer present. You also cannot do a partial rollback. If one edit changes both a tag and the title, but the title was again changed after that, you won't be able to roll it back (though you can first rollback the title change and then rollback the larger edit).

Any previous state can be reached by just rolling back the changes from most recent to least recent.

## Implementation Details
* Post history now has an "extra" field in JSON format (in the DB, in Rails its just a Hash), which holds extra information about the Post History. Close reasons and rollback information are stored in here. (**Advice on whether this is the best way appreciated**)
* Post History has `can_rollback?` which checks whether the post is currently matching the after state of the change. If that is the case, it is rollbackable.
* The Post History Controller performs a rollback with the logic in `rollback_post_history`. For question_reopened it needs to find the event preceding it which closed the question to figure out the close reason.
* The Post History Controller records a rollback as a Post History by inverting the name of the original post history type (`PostHistoryType.name_inverted`) and storing that as new history item together with a full before and after of the post.
* The rolled back Post History gets `rolled_back_with` added to its `extra` field which stores the id of the Post History which rolled it back. This is used to determine whether a post was rolled back and the precise id may in the future be useful if we want to expand the feature.
* Close reasons and rollback information is stored in the comment of the post history. Comments also render in markdown now to allow links in them (from #1134), but most elements are disallowed.
* A migration adds close reasons into the history for all posts where it is still recoverable.

## Images
The rolled back history item receives a strikethrough. The rolling back item has a comment about the rollback.
image

Sometimes multiple rollback items can be rolled back. This can happen even if e.g. the title was changed multiple times, as long as the end title is the same as the one applied in the event.
image

## WIP
* [ ] Design of the undo button

Fixes #215
Fixes #340


Taeir:
With post history rollbacks, mods can now finally duke it out in the history in a nice way:
image

Taeir:
Tested with various changes and multiple users, works as intended:
image

PR #1143, opened 2023-07-23
History redaction

Taeir:
I had actually implemented this already 10 months ago for redacting full-blown answers to mandatory homework, but I did not feel like the feature would be something Codidact  could benefit from. While chatting with @superplane39 the topic came up and they indicated that the feature would actually be useful, so here it is.

UPDATE: I did completely rewrite it now.

### Details
When making an edit, you can tick a box to make that edit a redaction. If you do, it follows your edit up by a history hidden (replaces initial revision) and sets all previous history as hidden. When viewing history, details of hidden items are not shown, except if the user is the owner of the post, the person who made that history item or is an admin.

As such, all attribution is kept (as required by common licenses), but hides the diffs which could contain the redacted information.

Currently, all users who can make edits can make those edits redacting edits.
- This allows for quick responses to prevent impact of whatever info needed to be redacted.
- The potential for misuse is limited, as admins/post owners can see the history still and can then address the situation.

### Screenshots
Edit page (redaction at the bottom):
Edit page

History for post owner/admin:
History for post owner/admin

History for other users:
History for other users

### DIFF
--- a/app/controllers/post_history_controller.rb
+++ b/app/controllers/post_history_controller.rb
@@ -6,7 +6,7 @@ class PostHistoryController < ApplicationController
       return not_found
     end
 
-    @history = PostHistory.where(post_id: params[:id]).visible(current_user)
+    @history = PostHistory.where(post_id: params[:id])
                           .includes(:post_history_type, :user, post_history_tags: [:tag])
                           .order(created_at: :desc, id: :desc).paginate(per_page: 20, page: params[:page])
     render layout: 'without_sidebar'
--- a/app/controllers/post_history_controller.rb
+++ b/app/controllers/post_history_controller.rb
@@ -8,7 +8,7 @@ class PostHistoryController < ApplicationController
 
     @history = PostHistory.where(post_id: params[:id]).visible(current_user)
                           .includes(:post_history_type, :user, post_history_tags: [:tag])
-                          .order(created_at: :desc).paginate(per_page: 20, page: params[:page])
+                          .order(created_at: :desc, id: :desc).paginate(per_page: 20, page: params[:page])
     render layout: 'without_sidebar'
   end
 end
--- a/app/controllers/post_history_controller.rb
+++ b/app/controllers/post_history_controller.rb
@@ -6,7 +6,8 @@ class PostHistoryController < ApplicationController
       return not_found
     end
 
-    @history = PostHistory.where(post_id: params[:id]).includes(:post_history_type, :user, post_history_tags: [:tag])
+    @history = PostHistory.where(post_id: params[:id]).visible(current_user)
+                          .includes(:post_history_type, :user, post_history_tags: [:tag])
                           .order(created_at: :desc).paginate(per_page: 20, page: params[:page])
     render layout: 'without_sidebar'
   end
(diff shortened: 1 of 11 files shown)


cellio:
I agree that limiting the scope of a particular problem by allowing any editor to redact is good, but this also has the potential to remove others' attribution, which is a violation of the CC-BY-SA licenses that many communities use.  Example: user A posts, user B edits, user A edits and redacts -- now user B's work is attributed to user A.  Or user C edits and redacts, and now has attribution for both A and B.

I can think of two ways to mitigate this and I don't know which is better:

1. Only allow redaction if no one else has edited. You post, realize you included PII, and edit/redact to fix that -- that's fine.
2. Allow any editor to redact, but show all other users who had edits (without the details), to preserve some measure of attribution.  This might not be license-compliant but it's a good-faith effort.

Redaction should be a rare event.  But if it's not, moderators will want to know -- that's a signal that at least one community member probably doesn't understand some of the guidelines (on either posting or redacting).  Can we automatically raise a moderator flag on the post when someone redacts a post?  The flag can be attributed to the System user and either be an "other" flag with text of "redacted post" or a new "redaction" flag type that is not exposed to users -- whatever's easier.

cellio:
Question: in the screenshot, the text under the Redact heading looks like a link.  Is it? What does it link to?

trichoplax:
For option 1, would that mean anyone can redact provided there is only 1 author in the edit history? That is, if I see someone else's PII, can I press redact immediately rather than inform them and it have to wait until they are next online?

cellio:
> For option 1, would that mean anyone can redact provided there is only 1 author in the edit history? That is, if I see someone else's PII, can I press redact immediately rather than inform them and it have to wait until they are next online?

Yes.  I hadn't thought about this case (I was fixated on the attribution issue), but what you said makes sense to me.  The author can see the full history and can thus restore content that was wrongly redacted.

trichoplax:
Sounds good. In such a case, should we inform the author in addition to informing mods?

cellio:
> Sounds good. In such a case, should we inform the author in addition to informing mods?

Since a redaction is an edit (in the sense that it changes the edit history), I assumed the author already got notified.  Is that the case @Taeir ?  Or do we need to add this?

Taeir:
In light of the mentioned issues with regards to licensing, and thinking about the problem again, I think we should change the approach to a different and simpler one.

Rather than have a history hidden event, store on the set of events at the moment of redaction that their contents should be hidden, then don't display their contents to non-moderator users. The attributions will remain and the query to determine what should be visible becomes much less complex. This also avoids the need for some of the other changes.

Taeir:
@cellio @trichoplax I now made changes to the approach where it only hides the details of the events, rather than the full events themselves. 

Currently the logic is set to only allow admins to see the redaction. I think there may be cases where a user also wouldn't want moderators to see. Not sure if that is the way to go though.

> Question: in the screenshot, the text under the Redact heading looks like a link. Is it? What does it link to?

It's not a link.

> Since a redaction is an edit (in the sense that it changes the edit history), I assumed the author already got notified. Is that the case @Taeir ? Or do we need to add this?

I don't think any edit triggers a notification to the original author, but if it does, the redaction will also (it is now the same event/logic as a normal edit).

> Redaction should be a rare event. But if it's not, moderators will want to know -- that's a signal that at least one community member probably doesn't understand some of the guidelines (on either posting or redacting). Can we automatically raise a moderator flag on the post when someone redacts a post? The flag can be attributed to the System user and either be an "other" flag with text of "redacted post" or a new "redaction" flag type that is not exposed to users -- whatever's easier.

Hmm, I think it would be possible to add a flag for it if you want that. Is it necessary to pull the attention to the post though?

ArtOfCode- (approved):

PR #1150, opened 2023-08-01
check setting for minimum title length instead of hard-coding a value

cellio:
We changed the server to allow changing the minimum title and body lengths, but the UI code still hard-wires those values.  This change fixes the title.  I haven't found the right invocation to fix the body yet, but either I'll figure it out soon or we should at least fix titles, which directly affect proposals (and which we marked status-completed before realizing there was this additional piece).

### DIFF
--- a/app/views/posts/_form.html.erb
+++ b/app/views/posts/_form.html.erb
@@ -65,7 +65,9 @@
     </div>
     <div>
       <span class="has-float-right has-font-size-caption js-character-count-post-title hide"
-            data-max="255" data-min="15" data-display-at="0.75">
+	    data-min="<%= category&.min_title_length %>"
+	    data-max="255"
+	    data-display-at="0.75">
         <i class="fas fa-ellipsis-h js-character-count__icon"></i>
         <span class="js-character-count__count">0 / 255</span>
       </span>


app/views/posts/_mdhint.html.erb
```diff
@@ -1,8 +1,15 @@
+<%
+  # defaults
+  category ||= defined?(category) ? category : nil
+  %>
```
ArtOfCode-, line 4: This doesn't work how you think it does - this block means you need to pass a `category` value to the `_mdhint` view when it's rendered, it doesn't automatically pull the category from elsewhere.

ArtOfCode- (changes requested):

Taeir:
Ah, i actually also implemented this 🙈  I guess that is the problem of no associated issue

app/views/posts/_mdhint.html.erb
ArtOfCode-: Not too difficult, though: `mdhint` is rendered as part of `shared/body_field` on line 57 of `_form`. You should just need to copy this block to both `body_field` and `mdhint`, then add a `category: category` parameter to the `render` calls for each of them - line 57 in `_form`, and line 25 in `body_field`.

cellio: I've made these changes and tested both creating and editing.

ArtOfCode- (approved):

PR #1151, opened 2023-08-02
Post Templates

MoshiKoi:
Proof of concept that I made in an afternoon lol

Addresses https://meta.codidact.com/posts/276900

Templates are tied to a post type, so you can have templates for questions, answers, articles, help docs, and even templates for templates (probably not useful, but hey, why not).

### Database changes:

 - Adds a new field to `posts`, `template_post_type_id`
 - Adds `before_template_post_type_id` and `after_template_post_type_id` to `post_histories`

### Seed changes

 - Creates a new `PostType`, `PostTemplate`
 - Seeds a new default Category called Templates to create and edit templates. The default permission level is that everyone can view, but only moderators can edit
 - Due to behavior changes (below), the sequence field of Q&A and Meta are now seeded as 1 and 2, respectively

### Behavior changes

 - Settings the "sequence" of a category to -1 will now hide the category.

### Other

Trust Level only affects *creating* posts afaik, so users with the edit ability will still be able to edit these posts since by default, I let everyone be able to view the category. This is easily configured since it's just a normal category, so if you like, you can set the permissions however you want, and even lower it to let normal users create templates.


app/assets/javascripts/posts.js
```diff
@@ -6,6 +6,13 @@ const ALLOWED_ATTR = ['id', 'class', 'href', 'title', 'src', 'height', 'width',
   'start', 'dir'];
 
 $(() => {
+  $('.js-template').on('click', async ev => {
+    const $postField = $('.js-post-field');
+    $tgt = $(ev.target);
+    const content = $('#template-md-' + $tgt.attr('data-template-id')).val();
```
ArtOfCode-, line 12:
```suggestion
    const content = $(`#template-md-${$tgt.attr('data-template-id')}`).val();
```

app/models/post_type.rb
```diff
@@ -19,6 +19,10 @@ def reactions
     end
   end
 
+  def templates
+    Post.undeleted.where(post_type_id: PostTemplate.post_type_id).where(template_post_type_id: self).to_a
```
ArtOfCode-, line 23: Won't work how you think it does - you need _either_:

```ruby
.where(template_post_type: self)
```

or

```ruby
.where(template_post_type_id: self.id)
```

db/migrate/20230801014134_add_template_post_type_to_posts.rb
```diff
@@ -0,0 +1,5 @@
+class AddTemplatePostTypeToPosts < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :posts, :template_post_type, foreign_key: { to_table: :posts}
```
ArtOfCode-, line 3: nitpick: missing space before closing `}`

ArtOfCode- (changes requested):

app/models/post_type.rb
MoshiKoi: Somehow it does work (I've tested this after all) but yeah it's better not to rely on it

MoshiKoi: The first one doesn't work (`Mysql2::Error: Unknown column 'posts.template_post_type' in 'where clause'`) but the second one does.

ArtOfCode-: That's because you're missing an association - you need to add the `belongs_to :template_post_type` line to the Post model, then the first variant will work.

ArtOfCode-:
More generally: I'm not sure that having a Templates category at the top of _every_ page on the site is right - it almost feels _too_ visible for something that won't be used _that_ frequently. I like having it as a category, because it makes template admin familiar, but can we find a way to hide the Templates category from the top bar? Might need to add a field to the Category model. It'll still be linked from /categories, just not on every page.

MoshiKoi:
You can now remove the category from the header by setting the `sequence` setting to null. Seeds have also been updated.

cellio:
> In terms of functionality, everything should be there, but the UI is really ugly; if anyone has any ideas how to improve it, please let me know.

Can you position the modal over the post editor (buttons + textbox), read-only with the "apply" action being more obvious?

What happens to text that's already in the editor when you apply a template?  Does the template get inserted after (or before), are templates not available if there's text, or does existing text get clobbered?  (Any but the last is ok with me, preferring an additive approach over disabling.)

MoshiKoi:
Template content is inserted, not replaced (though apparently I forgot to push that change before going on my break from QPixel, whoops)

cellio:
> You can now remove the category from the header by setting the `sequence` setting to null. Seeds have also been updated.

Some existing categories don't set a value.  Consider `-1` or even `0` as opposed to null?

MoshiKoi:
> > You can now remove the category from the header by setting the `sequence` setting to null. Seeds have also been updated.
> 
> Some existing categories don't set a value. Consider `-1` or even `0` as opposed to null?

Was allowing existing categories to not have a value intentional behavior before this PR? If so, what was the expected behavior? I had assumed that categories not being seeded with values was a bug.

cellio:
I believe the default is to show them in order of creation, and until a community needs something beyond Q&A and Meta, there was no reason to set explicit values.  (Even then, IIRC correctly, some just set Meta to 99 and otherwise leave them blank.)  For intent we'll have to ask @ArtOfCode- ; I'm just reporting what's in production.   I don't have an objection to requiring categories to specify where they are in the sequence, but we'll need to add migrations for ones that don't have it set today.

ArtOfCode-:
It was intentional to allow categories to not set the `sequence` value, if that helps - empty has so far been a valid value.

MoshiKoi:
Ah, OK. If having it blank had well defined behavior, I'm fine with using -1 instead

Taeir:
It seems impossible to select the templates post type as allowed for a category since the old value remains in the redis cache forever, perhaps we need to do something with seeds or after_create handlers to clear specific cache values and make things more robust to changes. Of course, post_types don't really change often, but having behavior in the code to automatically deal with necessary cache invalidation for it should a new one be created would be useful. Not really something for this PR though.

Taeir:
Since the title name of the template is important for the list, it could be common to have duplicate names, e.g. "Default" both for Question and for Answer. However, after creation you cannot see in the list what post type it was for.

I suggest to add to `/app/views/posts/_list.html.erb`, line 36:

```diff
        <%= post.post_type.is_top_level ? post.title : post.parent.title %>
        <%= post.post_type.is_closeable && post.closed && !post.duplicate_post  ? "[closed]" : "" %>
        <%= post.duplicate_post && post.post_type.is_closeable && post.closed ? "[duplicate]" : "" %>
+       <%= post.post_type.name == 'PostTemplate' ? "[#{post.template_post_type.name}]" : "" %>
      <% end %>
    </div>
    <% if (SiteSetting['PostBodyListTruncateLength'] || 0) > 0 %>
```

Which will show up as follows in the list:

image

db/seeds/categories.yml
```diff
@@ -1,4 +1,5 @@
 - name: Q&A
+  sequence: 1
```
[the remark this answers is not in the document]
MoshiKoi: I don't think it's necessary, no, but I don't think it hurts to be explicit here

app/assets/javascripts/markdown.js
```diff
@@ -15,6 +15,14 @@ $(() => {
     $field.val(prev.substring(0, $field[0].selectionStart) + text + prev.substring($field[0].selectionEnd));
   };
 
+
+  $('.js-template').on('click', async ev => {
+    const $postField = $('.js-post-field');
+    $tgt = $(ev.target);
+    const content = $(`#template-md-${$tgt.attr('data-template-id')}`).val();
+    replaceSelection($postField, content);
```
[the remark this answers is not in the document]
MoshiKoi: See Monica's comment here: https://github.com/codidact/qpixel/pull/1151#issuecomment-1710485872, but also a confirmation would also be nice to have if they still have selected text that might be overridden (undo functionality is broken on all Markdown tools (#1152), so there's no going back).

app/views/shared/_body_field.html.erb
```diff
@@ -35,7 +35,7 @@
         end[1]
       %>
     <% end %>
-    <%= render 'shared/markdown_tools' %>
+    <%= render 'shared/markdown_tools', post_type: post&.post_type if post.is_a? Post %>
```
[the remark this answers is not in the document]
MoshiKoi: Ah, thanks for noticing. I think I had a scoping issue here and this was meant to be `post_type: (post&.post_type if post.is_a? Post)`. Yours is definitely cleaner anyway though 👍

app/assets/javascripts/markdown.js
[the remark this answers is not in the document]
Taeir: Yeah in my tests if you select text it does override (which is luckily consistent with what the javascript method is named after). The question is how complex we would want to make it.

MoshiKoi:
As a proof of concept, I think this worked out well in order to see what sort of functionality we want, but due to lack of further activity (entirely my fault) and the divergence of the main branch, I'll close this PR for now to give space for any new implementation if there is more interest.

PR #1154, opened 2023-08-02
Fix errors on non-signed-in user viewing redacted history

Taeir:


trichoplax (approved):
Looks good. Redacted revisions now show as such when no user is signed in.

PR #1155, opened 2023-08-02
Add (more) system tests relating to sign in / sign out

Taeir:
I updated the passwords of all users to be 'test123' to be able to sign in with arbitrary users.

Fixed the log_out helper function. The button does not appear within the header element, and `click_on` did not seem to wait correctly for the button to have appeared. Find does do this, so it's what I used now.


Taeir:
GitHub seems to be having some issues. CircleCI cannot seem to checkout the branch and new commits aren't showing....
Since it involves tests, this should only be merged if it passes the CI.

ArtOfCode- (approved):

PR #1158, opened 2023-08-02
Add system tests for posts

Taeir:
This adds various tests for dealing with posts:

- [x] Question creation
  - [x] when not signed in (disallowed)
  - [x] when signed in, valid information
  - [x] when signed in, with invalid info (fails validations)
- [x] Answer creation
- [x] Selection screen for selecting post type to create
- [x] View post + answers
  - [x] Sort answers
- [x] Edit post directly

Of course, more tests can be imagined, but this PR should provide a decent coverage of the features related to creating, editing and viewing the different types of posts that we support. 

NOTE: I had wanted to properly test that the submit post button gets disabled if the title/body is too short, but this only happens if it was the last field filled in. We are also not making use of default HTML form validations. Something to look into for the future perhaps.

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -62,4 +62,42 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
       users(user_or_fixture)
     end
   end
+
+  # In the post form, this method will select the given tag.
+  #
+  # @param tag_name [String] the name of the tag
+  # @param create_new [Boolean] whether creating a new tag is allowed (default false)
+  def post_form_select_tag(tag_name, create_new = false)
+    # First enter the tag name into the select2 search field for the tag
+    within find_field('Tags (at least one):').find(:xpath, '..') do
+      find('.select2-search__field').fill_in(with: tag_name)
+    end
+
+    # Get the first item listed that is not the "Searching..." item
+    first_option = find('#select2-post_tags_cache-results li:first-child') { |el| el.text != 'Searching…' }
+
+    if first_option.first('span').text == tag_name
+      # If the text matches the tag name, first check whether we are creating a new tag.
+      # If so, confirm that we are allowed to. If all is good, actually click on the item.
+      if create_new || !first_option.text.include?('Create new tag')
+        first_option.click
+      else
+        raise "Expected to find tag with the name #{tag_name}, " \
+              'but could not select it from options without creating a new tag.'
+      end
+    elsif create_new
+      # The first item returned is not the tag we were looking for (another tag partial match + not existing)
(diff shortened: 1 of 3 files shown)


cellio:
Do we need to test articles separately or is that covered?  That is, does "question" mean top-level posts, or specifically _questions_?

Taeir:
> Do we need to test articles separately or is that covered? That is, does "question" mean top-level posts, or specifically _questions_?

Well, there is a page for when a category has multiple types of posts, then first you need to select one (or at least be presented this option). So I think it would be valuable to cover at least that page in our unit tests. The article creation itself happens on the exact same page with the same UI components on it, so I don't think we need to system test that.

I think the power for system tests lies in collaboration with unit tests. We don't need a system test for every single validation that could go wrong. We should have at least one system test that checks whether validations are applied at all in the UI (e.g. by just having an empty title), and then some unit tests that check the specific validation cases that should be rejected/accepted. Together they give reasonable confidence that validations are applied and that the validations check all parts.

So in this case, a system test that checks post creation, and then unit tests that also try whether article creation works. Perhaps this also applies to creating answers, but I think there is a different way to get there and it may load different UI components so there may be merit to test that separately.

test/application_system_test_case.rb
```diff
@@ -62,4 +62,42 @@ def user(user_or_fixture)
...
+  # @param tag_name [String] the name of the tag
+  # @param create_new [Boolean] whether creating a new tag is allowed (default false)
+  def post_form_select_tag(tag_name, create_new = false)
+    # First enter the tag name into the select2 search field for the tag
+    within find_field('Tags (at least one):').find(:xpath, '..') do
+      find('.select2-search__field').fill_in(with: tag_name)
+    end
+
+    # Get the first item listed that is not the "Searching..." item
+    first_option = find('#select2-post_tags_cache-results li:first-child') { |el| el.text != 'Searching…' }
```
ArtOfCode-, line 77: Careful: the ellipsis after "Searching" is a single Unicode character here, but I believe it's three dots in the actual UI - worth checking this actually picks the right tag.

ArtOfCode- (commented):
Doesn't need changes, but does need checking - see comment

test/application_system_test_case.rb
Taeir: I actually fought for quite a while with getting this to work correctly because I was using just 3 dots (and then the test fails). When getting the exact text of this element, it turned out that it was a single character.

That said, i will add an assertion that verify that the right tag is selected

ArtOfCode-: Huh. I thought this was something we set - having just looked, the "Searching" text actually comes from select2, so that makes more sense.

ArtOfCode- (approved):

PR #1162, opened 2023-08-03
Fix Post.locked?

Taeir:
While writing comments and adding rbs specs, I noticed that the `Post#locked?` method is not returning the correct thing. It should return whether the post is locked, to which the answer after the first 2 lines should be false.


Taeir:
Looks like either the tests are expecting a different behavior, or we are as well. Needs investigation.

ArtOfCode- (approved):

ArtOfCode-:
Code looks good, but apparently breaks things (tests report some 200 responses that should be 403s).

PR #1164, opened 2023-08-03
Include recovery codes

### LINKED ISSUE #266: Enabling 2FA should include some recovery codes
https://software.codidact.com/questions/277158

> After enabling two factor authentication (2FA) for my account I'm only notified It is successfully turned On.
>
> I think it should include some recovery codes as well (to be used in case I can't access my authenticator app, etc.).

@ArtOfCode- wrote in a comment:

> Just realised I never actually responded to this. It's not one we thought of (the 2FA system was copied over from another, older, implementation of mine from before it was a recommended practice), but it's definitely one we should do. — ‮edoCfOtrA‭ 3 months ago

ArtOfCode-:
Fixes #266 

Adds recovery codes for 2FA accounts. This instructs the user to save their recovery code before enabling 2FA.

Signing in with a recovery code disables 2FA on the account so the user can reconfigure it with a new app.

Also includes an email and script to set recovery codes on existing 2FA accounts, which can be manually run.

### DIFF
--- a/app/views/two_factor_mailer/backup_code.html.erb
+++ b/app/views/two_factor_mailer/backup_code.html.erb
@@ -1,0 +1,24 @@
+<p>Hi <%= @user&.username %>,</p>
+
+<h3>Your two-factor authentication backup code</h3>
+
+<p>
+  You're receiving this email because you have two-factor authentication enabled on your Codidact network account. We've
+  recently added backup codes for every account using 2FA, so that you can still log in if you lose access to your
+  authenticator app.
+</p>
+
+<p>
+  Your backup code is shown below. Please save it somewhere safe. Please note that logging in using this backup code
+  will disable 2FA on your account so that you can reconfigure it using a new app or device &mdash; this will also
+  reset your backup code.
+</p>
+
+<p>
+  Your backup code: <strong><code><%= @user&.backup_2fa_code %></code></strong>
+</p>
+
+<p>
+  Thanks,<br/>
+  The Codidact Team
+</p>
--- a/app/views/two_factor_mailer/backup_code.text.erb
+++ b/app/views/two_factor_mailer/backup_code.text.erb
@@ -1,0 +1,17 @@
+Hi <%= @user&.username %>,
+
+Re: Your two-factor authentication backup code
+
+
+You're receiving this email because you have two-factor authentication enabled on your Codidact network account. We've
+recently added backup codes for every account using 2FA, so that you can still log in if you lose access to your
+authenticator app.
+
+Your backup code is shown below. Please save it somewhere safe. Please note that logging in using this backup code
+will disable 2FA on your account so that you can reconfigure it using a new app or device - this will also
+reset your backup code.
+
+Your backup code: <%= @user&.backup_2fa_code %>
+
+Thanks,
+The Codidact Team
--- a/scripts/create_backup_2fa_codes.rb
+++ b/scripts/create_backup_2fa_codes.rb
@@ -1,0 +1,4 @@
+User.where(enabled_2fa: true).each do |user|
+  user.update(backup_2fa_code: SecureRandom.alphanumeric(24))
+  TwoFactorMailer.with(user: user, host: 'meta.codidact.com').backup_code.deliver_later
+end
--- a/app/controllers/two_factor_controller.rb
+++ b/app/controllers/two_factor_controller.rb
@@ -14,7 +14,8 @@ class TwoFactorController < ApplicationController
     case params[:method]
     when 'app'
       @secret = ROTP::Base32.random
-      current_user.update(two_factor_token: @secret, two_factor_method: 'app')
+      current_user.update(two_factor_token: @secret, two_factor_method: 'app',
+                          backup_2fa_code: SecureRandom.alphanumeric(24))
       totp = ROTP::TOTP.new(@secret, issuer: 'codidact.com')
       uri = totp.provisioning_uri("#{current_user.id}@users-2fa.codidact.com")
       qr_svg = RQRCode::QRCode.new(uri).as_svg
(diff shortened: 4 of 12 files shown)


cellio:
If I understand correctly, the "we've added codes" email is effectively sending a password in the clear in email not expected by the recipient.  (As opposed to when you make a request and you're monitoring what's coming in.)  Could we, instead, give the user a way to look up the code while logged in?

trichoplax:
> Signing in with a recovery code disables 2FA on the account so the user can reconfigure it with a new app.

If a user considers the security of their account sufficiently important to switch on 2FA, could we avoid disabling it until they explicitly request it? Is it possible to make the reconfiguration with a new app the step that disables the previous 2FA, rather than having a period during which there is no 2FA?

trichoplax:
> Could we, instead, give the user a way to look up the code while logged in?

I like the idea of this, perhaps combined with the ability to send an email saying "Here is where you can find your recovery codes". The page displaying the recovery codes could perhaps also have a `<details>` section to avoid displaying them immediately, similar to the recent change to the mobile sign in QR code page.

ArtOfCode-:
> Is it possible to make the reconfiguration with a new app the step that disables the previous 2FA, rather than having a period during which there is no 2FA?

Not as easily. It'd be doable, but it's a bit of a pain to carry that state through and ensure the user actually goes through with resetting 2FA - easier to just disable it with a reminder to reconfigure it.

ArtOfCode-:
> Could we, instead, give the user a way to look up the code while logged in?

Done - I've changed the email to direct the user to where to find it, and added the code to the 2FA settings behind a `<details>` section.

app/views/two_factor_mailer/backup_code.text.erb
```diff
@@ -0,0 +1,17 @@
+Hi <%= @user&.username %>,
```
cellio, line 1: Does backup_code.html.erb replace this?

ArtOfCode-: They supplement each other - we send the email multipart so that we can send both content types - if someone can't read or doesn't want HTML email, they can use the plain text form.

cellio: Ah.  The text version includes the code, so I thought it was a leftover.

trichoplax: Is the text version intended to have the same wording as the HTML version, and direct the user to look up their backup code rather than including it in the email?

ArtOfCode-: Whoops, missed that.

app/views/two_factor/tf_status.html.erb
```diff
@@ -45,4 +45,17 @@
...
+<h3>Recovery</h3>
+<p>
+  Your <strong>backup code</strong> for two-factor authentication can be found below. You should save this somewhere
+  safe &mdash; if you lose access to your authenticator app, you can sign in using this code to get back into your
+  account and reconfigure 2FA.
+</p>
+
+<details>
+  <summary>Show code</summary>
+  <label for="backup-code" class="form-element">2FA backup code</label>
```
Taeir, line 59: Normally, sites give me backup codes only once (when configuring) and then don't let me see them any more. There is a potential security risk if someone can access their 2FA backup code from their account, as this allows you to effectively go from having a session cookie to full access to the account.

As a suggestion, how about adding a checkbox + button that says "I have saved this backup code." and after confirmation the backup code will no longer be displayed on this page?

app/views/two_factor_mailer/backup_code.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<p>
+  You're receiving this email because you have two-factor authentication enabled on your Codidact network account. We've
+  recently added backup codes for every account using 2FA, so that you can still log in if you lose access to your
+  authenticator app.
+</p>
+
+<p>
+  Your backup code can be found in your user account settings, under Two-Factor Authentication settings. Please save it
+  somewhere safe. Please note that logging in using this backup code will disable 2FA on your account so that you can
+  reconfigure it using a new app or device &mdash; this will also reset your backup code.
```
Taeir, line 14: Can we add a link here such that it is easier for the user to get to it:

```suggestion
  Your backup code can be found in your user account settings, under
  <a href="<%= two_factor_status_path %>">Two-Factor Authentication settings</a>. Please save it somewhere safe.
  Please note that logging in using this backup code will disable 2FA on your account so that you can reconfigure it
  using a new app or device &mdash; this will also reset your backup code.
```

app/views/two_factor/tf_status.html.erb
trichoplax: An alternative approach for consideration:

- Require a 2FA code before viewing the backup code.

This means if I leave my account signed in and unattended, no one can view my backup code unless they also have my 2FA device (such as my mobile phone).

ArtOfCode-: Done - I've required a current 2FA code before the backup code is shown.

Taeir (changes requested):
When you set the two factor method to be email, then you cannot access the backup code because it asks to enter a code, which one cannot do via email.

Taeir:
It also seems that I cannot enter a backup code because it contains letters and the UI wants me to input a number only. We need to remove that restriction (the feature works when I change the 2FA code field to be of type text rather than number).

app/views/two_factor_mailer/backup_code.html.erb
```diff
@@ -0,0 +1,21 @@
...
+
+<p>
+  You're receiving this email because you have two-factor authentication enabled on your Codidact network account. We've
+  recently added backup codes for every account using 2FA, so that you can still log in if you lose access to your
+  authenticator app.
+</p>
+
+<p>
+  Your backup code can be found in your user account settings, under
+  <a href="<%= two_factor_status_path %>">Two-Factor Authentication settings</a>. Please save it somewhere safe.
```
Taeir, line 13: After testing this a bit more thoroughly it turns out that one has to use url and not path to get a valid link in the mailers. Apologies for the mistake.

```suggestion
  <a href="<%= two_factor_status_url %>">Two-Factor Authentication settings</a>. Please save it somewhere safe.

```

scripts/create_backup_2fa_codes.rb
```diff
@@ -0,0 +1,4 @@
+User.where(enabled_2fa: true).each do |user|
+  user.update(backup_2fa_code: SecureRandom.alphanumeric(24))
+  TwoFactorMailer.with(user: user, host: 'meta.codidact.com').backup_code.deliver_later
```
Taeir, line 3: Testing in development, with `deliver_later` it doesn't seem to deliver the emails (probably because there is no active job configuration). Is production configured in such a way that this does work?

ArtOfCode-:
Should all be fixed.

Taeir (approved):
Tested locally, now correctly is disabled with email sign ins. LGTM

PR #1165, opened 2023-08-04
Add warning to hidden revisions

ArtOfCode-:
Simple one. Revisions hidden by redaction aren't very visible - there's only a small italic note at the top to tell you they're not visible to the public.

This adds a warning notice to hidden revisions to make it clear that the details of that revision are not public and should not be shared.


Taeir (approved):
Good improvement :) LGTM

PR #1173, opened 2023-08-11
Fix DOMPurify

### LINKED ISSUE #724: Preview renders unsupported HTML tags, but final post (correctly) doesn't
https://meta.codidact.com/posts/284503  
https://meta.codidact.com/posts/284505

When composing a post, an unsupported HTML tag (`div` and `kbd` in these reports) was accepted and rendered.  However, after submission the post doesn't render it, which is correct because we don't support the tag.  This is confusing to users who haven't memorized (or don't look up) which HTML tags we do/don't support.

I know there's at least one other issue about differences between preview and final rendering, though I couldn't find it.  We use different libraries in the two cases so differences aren't surprising.  If we can't use the same library (or logic) to render the Markdown in both cases, is there anything we can do to provide some feedback when editing?  Can we "lint" the post body and indicate if we found something?  Maybe, as with missing alt text, that could be something we do when the user clicks the "post" button, so it doesn't have to be a performance drain.

MoshiKoi:
Resolves #724 

According to the documentation, USE_PROFILES overrides ALLOWED_TAGS and ALLOWED_ATTR and should not be used together. Basically, we weren't actually using the client sanitizer at all. This should fix basically all the preview issues.

### DIFF
--- a/app/assets/javascripts/posts.js
+++ b/app/assets/javascripts/posts.js
@@ -146,7 +146,6 @@ $(() => {
         const converter = window.converter;
         const unsafe_html = converter.render(text);
         const html = DOMPurify.sanitize(unsafe_html, {
-          USE_PROFILES: { html: true },
           ALLOWED_TAGS,
           ALLOWED_ATTR
         });


trichoplax (approved):
Looks good. I've checked in my local development environment that this continues to strip out SVG and continues to allow MathJax (where relevant).

PR #1175, opened 2023-08-15
Fix seeds overriding edited help articles

Taeir:
It seems that help articles don't (always) have an initial revision. Because the seeds were expecting them to have this, they could actually consider help articles that were edited as not-edited, and update them when reseeding...


Oaphi (approved):
As for the change itself - LGTM to me, checked locally. I took the liberty to bring the branch up-to-date with `develop`, hopefully this is fine with you, @Taeir. I have one "elephant in the room" question, though - is there a reason why we _don't_ have an initial revision for seeded policy posts?

Taeir:
@Oaphi If I knew why, I would have fixed that instead :)

Regardless, I think it is good to do this type of "defensive programming" and not make too many assumptions on how data looks, such that it works in all cases.

Oaphi:
> @Oaphi If I knew why, I would have fixed that instead :)
> 
> Regardless, I think it is good to do this type of "defensive programming" and not make too many assumptions on how data looks, such that it works in all cases.

Yup, no issue with the update, was jist wondering if that's intended. I have a suspicion that the revision creation fails validation, but haven't got the time to look into it properly yet. Will let you know if that's the source of the problem!

Oaphi:
@Taeir ah, yes, the `PostHistory` creation fails due to the "User must exist" error and is silently swallowed. I will take a look when I am able to.

Taeir:
@Oaphi may be a seed order issue. I expect users to not have been created perhaps?

Oaphi:
> @Oaphi may be a seed order issue. I expect users to not have been created perhaps?

@Taeir nah, I've been running seeds on an already fully initialized database - it's actually due to the user not being passed as one of the parameters during post creation. I think if we just make the system user the owner of those posts, it should be fine. Will make a separate PR improving the seeds.

PR #1181, opened 2023-08-17
Configurable email senders

Taeir:
Previously, all email senders were hardcoded in the source to be `@codidact.com`/`@codidact.org` email addresses. This changes those to be configurable.

- Creates configuration for all the emails QPixel can send
- Require both developer and global admin to send email to all users, rather than just developer.
- Avoid accidental send all users email by adding more warnings and confirmation
- Move buttons for email sending together in admin tools.


ArtOfCode-:
LGTM - will wait for ready-to-merge status :)

Taeir:
@cellio I added you as reviewer not for the code, but because I think it would be good for you to review whether the email addresses configured actually make sense.
For reference, here are the settings:
image

cellio:
@Taeir I think SupportReceiverEmail should probably be `support@codidact.com` rather than `info`, but @ArtOfCode- please confirm.

Another question for Art: do we need a separate support address for donations, or can everything involving donations come down to DonationSenderName and DonationSenderEmail?

ArtOfCode-:
Email addresses can basically be whatever we like - support should be support@, yes, and donations are fine to all come from one address

PR #1182, opened 2023-08-18
Fix name of the sample .irbrc file in Docker's local-setup script

Oaphi:
This is a minor PR fixing incorrect naming of the sample `.irbrc` file in Docker's installation `local-setup.sh` script.


Taeir (approved):
Oops, thanks for fixing :)

PR #1183, opened 2023-08-18
Fix not being able to edit privilege thresholds to 0 from the admin panel

Oaphi:
This PR fixes a bug with privilege thresholds not being editable to `0`. The script submitting the form used to parse the input value and short-circuit to `null` on any falsy value, resulting in `0` being interpreted as `null`.

### DIFF
--- a/app/assets/javascripts/privileges.js
+++ b/app/assets/javascripts/privileges.js
@@ -30,7 +30,7 @@ $(() => {
 
     const rawValue = parseFloat($input.val() || '')
 
-    const value = rawValue || rawValue === 0 ? rawValue : null;
+    const value = Number.isNaN(rawValue) ? null : rawValue;
 
     const resp = await fetch(`/admin/privileges/${name}`, {
       method: 'POST',
--- a/app/assets/javascripts/privileges.js
+++ b/app/assets/javascripts/privileges.js
@@ -28,6 +28,7 @@ $(() => {
     const name = $input.data('name');
     const type = $input.data('type');
 
+    // incorrect input values will cause rawValue to be NaN
     const rawValue = parseFloat($input.val() || '')
 
     const value = Number.isNaN(rawValue) ? null : rawValue;
--- a/app/assets/javascripts/privileges.js
+++ b/app/assets/javascripts/privileges.js
@@ -29,7 +29,7 @@ $(() => {
     const type = $input.data('type');
 
     // incorrect input values will cause rawValue to be NaN
-    const rawValue = parseFloat($input.val() || '')
+    const rawValue = parseFloat($input.val())
 
     const value = Number.isNaN(rawValue) ? null : rawValue;
 
--- a/app/assets/javascripts/privileges.js
+++ b/app/assets/javascripts/privileges.js
@@ -27,7 +27,10 @@ $(() => {
     const $input = $td.find('.js-privilege-edit');
     const name = $input.data('name');
     const type = $input.data('type');
-    const value = parseFloat($input.val() || '') || null;
+
+    const rawValue = parseFloat($input.val() || '')
+
+    const value = rawValue || rawValue === 0 ? rawValue : null;
 
     const resp = await fetch(`/admin/privileges/${name}`, {
       method: 'POST',


cellio:
Merge target isn't `develop`?  What is the branch you want to merge to?

app/assets/javascripts/privileges.js
```diff
@@ -27,7 +27,10 @@ $(() => {
     const $input = $td.find('.js-privilege-edit');
     const name = $input.data('name');
     const type = $input.data('type');
-    const value = parseFloat($input.val() || '') || null;
+
+    const rawValue = parseFloat($input.val() || '')
```
trichoplax, line 31: If `$input.val()` is `null` or `undefined` then the result will be `parseFloat('')` which is `NaN`. This works perfectly with the next line, but may not be obvious to all readers. Is it worth adding a hint that the result will be `NaN`?

I would expect the following change to still give identical results in all cases:

```suggestion
    const rawValue = parseFloat($input.val() || 'NaN')
```

Do you think this is more or less readable?

app/assets/javascripts/privileges.js
```diff
@@ -27,7 +27,10 @@ $(() => {
...
+
+    const value = rawValue || rawValue === 0 ? rawValue : null;
```
trichoplax, line 33: Are the only possible falsy values for rawValue:
- 0
- NaN

If so, we could simplify this to:

```suggestion
    const value = Number.isNaN(rawValue) ? null : rawValue;
```

trichoplax (changes requested):
Works well. I can now change to zero, and can still change to "-".

I've suggested a couple of changes that might be more readable, and which I don't expect to change the behaviour. Let me know if I've misinterpreted anything or if it's not more readable.

app/assets/javascripts/privileges.js
Oaphi: Not a big fan of `Number.isNaN` (or the old `isNaN` function for that matter), but sure, I tried to come up with a scenario where it could fail, but every corner case seems to be correctly handled (and one has to deliberately attempt to work around the `Number.isNaN` check at that).

app/assets/javascripts/privileges.js
Oaphi: I think it is more confusing :) A couple of years from now no one will understand why we short-circuit to a string `"NaN"`. A comment should suffice for that, methinks, a moment.

Oaphi: One thing I just realized is that I am not even sure we need a short-circuit there as `parseFloat` will do the job either way.

trichoplax: Good point about not needing the short circuit `||`. I should have checked what parseFloat does with `null` and `undefined` before commenting.

In that case, could we remove the `|| ''`?

trichoplax (changes requested):
Looks like you are correct about parseFloat handling `null` and `undefined` so we don't need the `|| ''`. I've added a comment suggesting we could remove it, but mentioning it again here as it was a comment in a resolved conversation so I'm not sure if it will notify you.

app/assets/javascripts/privileges.js
trichoplax: If you wanted to avoid any type of `isNaN` the following is equivalent:

```js
    const value = rawValue === rawValue ? rawValue : null
```

I personally find this a little cryptic though - my preference is for being explicit about what's being checked for.

ArtOfCode- (approved):

Oaphi:
> Looks like you are correct about parseFloat handling `null` and `undefined` so we don't need the `|| ''`. I've added a comment suggesting we could remove it, but mentioning it again here as it was a comment in a resolved conversation so I'm not sure if it will notify you.

@trichoplax Thanks for an extra mention! For the future, don't worry about me not getting the notification, though, I track resolved conversations until the PR is merged at least. Marked it as resolved as I've added the clarifying comment and want to double-check with the spec if I haven't forgotten anything [I think I haven't, but it never hurts to be sure] that will affect the removal of the short-circuit (thus making it a separate concern from what we've discussed).

trichoplax (approved):

Issue #1169, opened 2023-08-07
Invalid URL in user profile website during editing leads to 500 error


trichoplax:
**Describe the bug**
If a user enters text into the "Website" input box during editing their user profile, and that text is not recognised as a valid URL, the user sees a 500 error page rather than meaningful feedback.

**To Reproduce**
1. Go to your user profile page.
2. Click on the "Edit" tab.
3. Scroll down to the "Website" input box.
4. Enter the text `a"` into the "Website" input box.
5. Press "Save".
6. See 500 error page.

**Expected behavior**
The user should see feedback so that they know that it was the website that was rejected. They could have changed many things at once in that edit, and the 500 error page gives no indication of which of those things was a problem.

cellio:
Also, letting 500s get to the user is not good.  We should either catch it and use one of those in-page error-notification popups or accept the bad text but not linkify it if it's not actually a URL.  (Presumably the former, since it sounds like the server is expecting a valid URL.)

PR #1188, opened 2023-08-23
Fix invalid profile website URI causing a 500 error page on invalid URIs

Oaphi:
This is a minor PR fixing #1169 by adding a `validate_profile_website` method to the users controller. The core of the issue is an unguarded attempt to parse the provided URI with `URI.parse`, which throws a `URI::InvalidURIError` on invalid URIs. Here is a preview that everything seems to be working as expected now:

[Screencast from 23.08.2023 13:31:38.webm](https://github.com/codidact/qpixel/assets/34833340/1c413f11-1af2-46d9-b167-27ace4aae153)

I opted to remove the URI from the parameters and notify the user as I believe this is not severe enough to cause other valid changes to fail.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -342,14 +342,25 @@ class UsersController < ApplicationController
     render layout: 'without_sidebar'
   end
 
+  def validate_profile_website(profile_params)
+    uri = profile_params[:website]
+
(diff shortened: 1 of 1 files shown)


Taeir (approved):
LGTM :)

PR #1189, opened 2023-08-23
Rework of the character count requirements display

### LINKED ISSUE #1145: Minimum length requirements are not shown to the user
Re: https://proposals.codidact.com/posts/289061

Minimum length requirements are not shown to the user for post titles (and post content, comments, and probably everywhere else that has a requirement, though I haven't checked); it just makes the text box turn red.

Could we display this information somewhere? We already have a display for when the user has *more* characters than the limit, so there shouldn't be a UI problem in triggering a similar display for when the user has less.

Oaphi:
This PR reworks the display of character count requirements, including:
- adds a shared char_count view for reusable & configurable character counters;
- adds proper abstractions (`setCounterState`, `setInputState`, `setSubmitButtonState`) for the character_count JS handler;
- ensures character count limits are always displayed on initial load so as the user knows how many they can enter (or left to enter);
- ensures current character count is properly displayed in forms with content already set;

The following is an example of how the character count will work after the change:

[Screencast from 23.08.2023 17:19:13.webm](https://github.com/codidact/qpixel/assets/34833340/d1740a6b-de32-497d-960a-8ae8216f099b)

Closes #1145

### DIFF
--- a/app/assets/javascripts/character_count.js
+++ b/app/assets/javascripts/character_count.js
@@ -1,5 +1,9 @@
 $(() => {
-  const setIcon = (el, icon) => {
+  /**
+   * Sets the icon to show before the counter, if any
+   * @param {'fa-times'|'fa-exclamation-circle'|'fa-check'} icon name of the icon to show
+   */
+  const setCounterIcon = (el, icon) => {
     const icons = ['fa-ellipsis-h', 'fa-check', 'fa-exclamation-circle', 'fa-times'];
     el.removeClass(icons.join(' ')).addClass(icon);
   };
@@ -61,16 +65,16 @@ $(() => {
     
     if(gtnMax || ltnMin) {
       setCounterState($counter, 'error');
-      setIcon($icon, 'fa-times');
+      setCounterIcon($icon, 'fa-times');
       setSubmitButtonDisabledState($button, 'disabled');
       setInputValidationState($tgt, 'invalid');
     } else if (gteThreshold) {
       setCounterState($counter, 'warning');
-      setIcon($icon, 'fa-exclamation-circle');
+      setCounterIcon($icon, 'fa-exclamation-circle');
       setSubmitButtonDisabledState($button, 'enabled');
     } else {
       setCounterState($counter, 'default');
-      setIcon($icon, 'fa-check');
+      setCounterIcon($icon, 'fa-check');
       setSubmitButtonDisabledState($button, 'enabled');
       setInputValidationState($tgt, 'valid');
     }
--- a/app/assets/javascripts/character_count.js
+++ b/app/assets/javascripts/character_count.js
@@ -20,10 +20,13 @@ $(() => {
     el.toggleClass('failed-validation', state);
   };
 
-  const setSubmitButtonState = (el, state) => {
-    if (el) {
-      el.attr('disabled', state).toggleClass('is-muted', state);
-    }
+  /**
+   * Sets the submit button's disabled state
+   * @param {'disabled'|'enabled'} state the state to set
+   */
+  const setSubmitButtonDisabledState = (el, state) => {
+    const isDisabled = state === 'disabled';
+    el.attr('disabled', isDisabled).toggleClass('is-muted', isDisabled);
   };
 
   $(document).on('keyup change paste', '[data-character-count]', (ev) => {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/character_count.js
```diff
@@ -4,57 +4,60 @@ $(() => {
...
+  const setCounterState = (el, state) => {
+    if(state == 'info') {
+      el.removeClass('has-color-yellow-700 has-color-red-500').addClass('has-color-primary');
+    } else if(state === 'warning') {
+      el.removeClass('has-color-red-500 has-color-primary').addClass('has-color-yellow-700');
+    } else if(state === 'error') {
+      el.removeClass('has-color-yellow-700 has-color-primary').addClass('has-color-red-500');
+    } else {
+      el.removeClass('has-color-red-500 has-color-yellow-700 has-color-primary');
+    }
```
ArtOfCode-, line 16: style: spaces before parens, new blocks on new lines:  
lint: strict equality checks (`===`) in JS unless you're dealing with `null/undefined`:

```js
if (a === 'b') {
  doAThing();
}
else if (a === 'c') {
  doAnotherThing();
}
```

There's a number of these across the PR - worth a double-check for them.

app/assets/javascripts/character_count.js
```diff
@@ -4,57 +4,60 @@ $(() => {
...
+    if(gtnMax || ltnMin) {
+      setCounterState($counter, 'error');
       setIcon($icon, 'fa-times');
-      if ($button) {
-        $button.attr('disabled', true).addClass('is-muted');
-      }
-    }
-    else if (count > 0.75 * max) {
-      $counter.removeClass('has-color-red-500 has-color-primary').addClass('has-color-yellow-700');
+      setSubmitButtonState($button, true);
```
ArtOfCode-, line 50: comment: I had to double check the implementation of `setSubmitButtonState` here - passing `true` when you want to disable the input feels weird. Makes sense when you actually look, but might be worth doing something about. Possibilities:

 * Rename the `state` parameter in each of these functions to `isDisabled`, or similar
 * `setSubmitButtonState($button, { disabled: true })` (might be overkill)
 * Invert the parameter in each function `const disabled = !state;`

Or something else!

app/views/shared/_char_count.html.erb
```diff
@@ -0,0 +1,28 @@
...
+
+<%
+  # defaults & normalization
+  cur ||= defined?(cur) && !cur.nil? ? cur.to_i : 0
+  max ||= defined?(max) && !max.nil? ? max.to_i : 255
+  min ||= defined?(min) && !min.nil? ? min.to_i : 0
+  threshold ||= defined?(threshold) && !threshold.nil? ? threshold.to_f : 0.75
+%>
+
+<span class="has-float-right has-font-size-caption js-character-count-<%= type %>"
```
ArtOfCode-, line 20: question: Am I missing something, or is the `type` not actually used anywhere? Do we need it?

also: the JS triggers on elements with a `data-character-count` attribute, which this doesn't have - what am I missing here?

app/views/shared/_char_count.html.erb
```diff
@@ -0,0 +1,28 @@
+<%#
```
ArtOfCode-, line 1: :D I like that this is its own file now - makes it much easier to work with!

ArtOfCode- (changes requested):

app/assets/javascripts/character_count.js
Oaphi: @ArtOfCode- still a draft, haven't bothered with linting it properly (my personal personal preference is same-line, so that's probably why), will do when I have enough time to spare to finalize it. As for the strict equality - yeah, seems like a typo, I don't think I've ever intentionally used lax equality.

app/views/shared/_char_count.html.erb
Oaphi: As for the `data-character-count` attribute - we actually do have it :) It's just set with an underscore, IIRC: https://github.com/codidact/qpixel/blob/47a4d2153851bb5009774f0535e0cae3012a8c39/app/views/posts/_form.html.erb#L65

ArtOfCode-: Ah, okay, I misunderstood the view. So this partial is _only_ the actual counter part - you still have to set the data attribute on the relevant form field separately?

Oaphi: @ArtOfCode- yup, that's how it currently works as the event listener is delegated. That said, I am open to extending the scope of the PR a little further and devise a way to get rid of the necessity to set `data-character-count`  the class name matching the one of the count (also related to removing the need for the `type` parameter).

app/assets/javascripts/character_count.js
Oaphi: Upon further inspection, I think the state can simply be a union of string literals `'enabled' | 'disabled'` with a `state === 'disabled'` check under the hood. Should make the intent clear as day (I don't like boolean / magic number parameters either)

app/views/shared/_char_count.html.erb
cellio: @ArtOfCode-  are you still requesting a change here or is this thread resolved?

ArtOfCode- (approved):

PR #1193, opened 2023-08-24
1186 tag merge creates duplicates

### LINKED ISSUE #1186: Merging a tag creates duplicate tagging on posts which previously had both tags
**Describe the bug**
If tagA and tagB are both present on a post, and then tagB is merged into tagA, the post now has tagA twice. This only shows once in the rendered post, but shows twice in the tags field during editing the post.

**To Reproduce**
1. Add 2 tags to a post.
2. Merge 1 tag into the other.
3. Edit the post.
4. See that the tags field contains the other tag twice.

**Expected behavior**
When tagB is merged into tagA, tagA should be added to all posts that previously had tagB, but only if each post does not already have tagA.

Currently tagA is added to all posts that previously had tagB, even if each post already has tagA.

Taeir:
# Depends on #1185

First replacement query is altered to not alter posts that have both the primary tag and subordinate tag.
Second replacement query removes the subordinate tag from posts that still have it (those not affected by the first query).

Fixes #1186

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -126,64 +126,68 @@ class TagsController < ApplicationController
 
     @subordinate = Tag.find params[:merge_with_id]
 
-    AuditLog.moderator_audit event_type: 'tag_merge', related: @primary, user: current_user,
-                             comment: "#{@subordinate.name} (#{@subordinate.id}) into #{@primary.name} (#{@primary.id})"
+    Post.transaction do
+      AuditLog.moderator_audit event_type: 'tag_merge', related: @primary, user: current_user, comment:
+        "#{@subordinate.name} (#{@subordinate.id}) into #{@primary.name} (#{@primary.id})"
 
-    # Replace subordinate with primary, except when a post already has primary (to avoid giving them a duplicate tag)
-    posts_sql = 'UPDATE posts INNER JOIN posts_tags ON posts.id = posts_tags.post_id ' \
-                'SET posts.tags_cache = REPLACE(posts.tags_cache, ?, ?) ' \
-                'WHERE posts_tags.tag_id = ? ' \
-                'AND post_tags.post_id NOT IN (SELECT post_id FROM posts_tags WHERE tag_id = ?)'
-    exec_sql([posts_sql, "\n- #{@subordinate.name}\n", "\n- #{@primary.name}\n", @subordinate.id, @primary.id])
+      # Replace subordinate with primary, except when a post already has primary (to avoid giving them a duplicate tag)
+      posts_sql = 'UPDATE posts INNER JOIN posts_tags ON posts.id = posts_tags.post_id ' \
+                  'SET posts.tags_cache = REPLACE(posts.tags_cache, ?, ?) ' \
+                  'WHERE posts_tags.tag_id = ? ' \
+                  'AND post_tags.post_id NOT IN (SELECT post_id FROM posts_tags WHERE tag_id = ?)'
+      exec_sql([posts_sql, "\n- #{@subordinate.name}\n", "\n- #{@primary.name}\n", @subordinate.id, @primary.id])
 
-    # Remove the subordinate tag from posts that still have it (the ones that were excluded from our previous query)
-    posts2_sql = 'UPDATE posts INNER JOIN posts_tags ON posts.id = posts_tags.post_id ' \
-                 'SET posts.tags_cache = REPLACE(posts.tags_cache, ?, ?) ' \
-                 'WHERE posts_tags.tag_id = ?'
-    exec_sql([posts2_sql, "\n- #{@subordinate.name}\n", "\n", @subordinate.id])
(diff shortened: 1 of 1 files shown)


app/controllers/tags_controller.rb
```diff
@@ -129,25 +129,32 @@ def merge
     AuditLog.moderator_audit event_type: 'tag_merge', related: @primary, user: current_user,
```
trichoplax, line 129: Would it be useful to make this log entry after the merge, so that it only gets recorded when successful? I don't know to what extent Rails already handles logging what was being attempted if there is an error. Do we need to check for and log errors ourselves?

app/controllers/tags_controller.rb
```diff
@@ -129,25 +129,32 @@ def merge
...
-    exec([sql.gsub('$TABLENAME', 'post_history_tags'), @primary.id, @subordinate.id])
-    exec([sql.gsub('$TABLENAME', 'suggested_edits_tags'), @primary.id, @subordinate.id])
-    exec([sql.gsub('$TABLENAME', 'suggested_edits_before_tags'), @primary.id, @subordinate.id])
+    exec_sql([sql.gsub('$TABLENAME', 'posts_tags'), @primary.id, @subordinate.id])
+    exec_sql([sql.gsub('$TABLENAME', 'categories_moderator_tags'), @primary.id, @subordinate.id])
+    exec_sql([sql.gsub('$TABLENAME', 'categories_required_tags'), @primary.id, @subordinate.id])
+    exec_sql([sql.gsub('$TABLENAME', 'categories_topic_tags'), @primary.id, @subordinate.id])
+    exec_sql([sql.gsub('$TABLENAME', 'post_history_tags'), @primary.id, @subordinate.id])
+    exec_sql([sql.gsub('$TABLENAME', 'suggested_edits_tags'), @primary.id, @subordinate.id])
+    exec_sql([sql.gsub('$TABLENAME', 'suggested_edits_before_tags'), @primary.id, @subordinate.id])
```
trichoplax, line 157: Could all of these SQL queries be enclosed in a single transaction, so that if any fail we don't leave the database in an incorrect state? I don't know whether Rails' capacity for running raw SQL extends to using transactions.

app/controllers/tags_controller.rb
```diff
@@ -129,25 +129,32 @@ def merge
...
 
     # Nuke it from orbit
     @subordinate.destroy
```
trichoplax, line 160: Is it possible to check for success of the previous queries at this point, and only destroy if the queries were successful, to avoid simply deleting a tag without replacing it with anything?

I don't know whether a failure in one of the previous queries would simply prevent the process from every reaching this line of the code, so I'm asking just in case.

trichoplax (changes requested):
I've suggested some potential changes to account for failed attempts. I'm not familiar with how Rails and SQL interact, so my suggestions are a little vague. Please let me know if the intention is unclear.

app/controllers/tags_controller.rb
```diff
@@ -129,25 +129,32 @@ def merge
...
                              comment: "#{@subordinate.name} (#{@subordinate.id}) into #{@primary.name} (#{@primary.id})"
 
-    # Take the tag off posts
+    # Replace subordinate with primary, except when a post already has primary (to avoid giving them a duplicate tag)
     posts_sql = 'UPDATE posts INNER JOIN posts_tags ON posts.id = posts_tags.post_id ' \
                 'SET posts.tags_cache = REPLACE(posts.tags_cache, ?, ?) ' \
-                'WHERE posts_tags.tag_id = ?'
-    exec([posts_sql, "\n- #{@subordinate.name}", "\n- #{@primary.name}", @subordinate.id])
+                'WHERE posts_tags.tag_id = ? ' \
+                'AND post_tags.post_id NOT IN (SELECT post_id FROM posts_tags WHERE tag_id = ?)'
```
trichoplax, line 136: Typo found during testing:

```suggestion
                'AND posts_tags.post_id NOT IN (SELECT post_id FROM posts_tags WHERE tag_id = ?)'
```

trichoplax (changes requested):

app/controllers/tags_controller.rb
Taeir: Good remark (though not code I touched :P). Really, all of this should be wrapped in a transaction. That means it either all applies, or none of it applies, to avoid exactly the case where we get partial execution. I will commit that.

app/controllers/tags_controller.rb
Taeir: If any query fails then an exception is thrown in rails and the execution stops (and error 500 is reported to the user).

app/controllers/tags_controller.rb
Taeir: Yes, had the same idea based on your previous comment. I have added it in this PR, and tested in a rails console whether the transaction actually applies for the raw SQL as well (it does).

app/controllers/tags_controller.rb
Taeir: Applied the change

trichoplax (approved):

PR #1194, opened 2023-08-24
Rename exec to exec_sql

Taeir:
For consistency with changes made in #1185


trichoplax (approved):
Tested locally. The script fails, but the failure is pre-existing and unaffected by this pull request. I am therefore approving this pull request, with the error raised as a separate issue #1195.

PR #1196, opened 2023-08-24
Fix 500 error on /users/me/filters

Taeir:
Check whether user is signed in before serving page.


trichoplax (approved):
Navigating to `users/me/filters` while not signed in now redirects to the sign in page as expected.

I've also confirmed that the same is true of `users/me/notifications` and `users/me/preferences`.

PR #1197, opened 2023-08-30
Skip min title validation for help/policy posts

Taeir:
Seeding was no longer creating help/policy posts with title lengths < 15, due to the recent validations changes. This PR excludes help and policy documents from the min title length requirement, to allow help articles like "FAQ" to be created again.

### DIFF
--- a/app/models/concerns/post_validations.rb
+++ b/app/models/concerns/post_validations.rb
@@ -8,7 +8,7 @@ module PostValidations
     validate :maximum_tag_length, if: -> { post_type.has_tags }
     validate :no_spaces_in_tags, if: -> { post_type.has_tags }
     validate :stripped_minimum_body, if: -> { !body_markdown.nil? }
-    validate :stripped_minimum_title, if: -> { !title.nil? && !['HelpDoc', 'PolicyDoc'].include?(post_type.name) }
+    validate :stripped_minimum_title, if: -> { !title.nil? }
     validate :maximum_title_length, if: -> { !title.nil? }
     validate :required_tags?, if: -> { post_type.has_tags && post_type.has_category }
   end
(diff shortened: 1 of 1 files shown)


app/models/concerns/post_validations.rb
```diff
@@ -8,7 +8,7 @@ module PostValidations
     validate :maximum_tag_length, if: -> { post_type.has_tags }
     validate :no_spaces_in_tags, if: -> { post_type.has_tags }
     validate :stripped_minimum_body, if: -> { !body_markdown.nil? }
-    validate :stripped_minimum_title, if: -> { !title.nil? }
+    validate :stripped_minimum_title, if: -> { !title.nil? && !['HelpDoc', 'PolicyDoc'].include?(post_type.name) }
```
MoshiKoi, line 11: Do we want to put this validation directly in `stripped_minimum_title` instead?

https://github.com/codidact/qpixel/blob/c7907f575d9f1ed0dafb08de4d2b810245e2e0d8/app/models/concerns/post_validations.rb#L48-L53

MoshiKoi: Also, it looks like this skips all validation, so it would accept 0-length titles (not sure if this is actually the case; I haven't tested). This probably wouldn't be a problem either way since these posts aren't generally user-editable, but it's something I noticed.

Oaphi: To follow up on @MoshiKoi's comment, proposed adjustment (not necessarily as implemented):

```ruby
def stripped_minimum_title
    if ['HelpDoc', 'PolicyDoc'].include?(post_type.name)
        min_title = 1
    else
        min_title = category.nil? ? 15 : category.min_title_length
    end

    if (title&.gsub(/(?:^[\s\t\u2000-\u200F]+|[\s\t\u2000-\u200F]+$)/, '')&.length || 0) < min_title
        errors.add(:title, "must be more than #{min_title} non-whitespace characters long")
    end
end
```

The above won't skip the rest of the steps the validation method runs, as well as ensures the seeded docs require at least one character in the title (initially I thought it should be 3, but since those aren't user-editable, I think it's fine to just check if there's at least one. Might be worth abstracting into a config option as well).

Taeir: Good remark, and thanks for the suggestions. I have applied them.

Oaphi (approved):
Changes LGTM (I personally prefer ternary expressions, but my background mostly includes languages where you can't assign the result of evaluating the `if...else` directly anyway; as it passes rubocop, all good)

Taeir:
> Changes LGTM (I personally prefer ternary expressions, but my background mostly includes languages where you can't assign the result of evaluating the `if...else` directly anyway; as it passes rubocop, all good)

@Oaphi I was thinking about the options, but since we have 3 cases here ternary expressions seemed a bit much, and this if-else-end expression is what I had to do to make rubocop happy, so yeah 😅

PR #1203, opened 2023-09-26
Add a notice when tags and attributes are rejected

### LINKED ISSUE #1140: Inform the user of disallowed HTML
There have been a lot of question on Discord and Meta about posts being different from what was shown in the preview. Currently, due to the way we handle submissions, differences between the post preview and the actual post comes from sanitization.

Could we have something that automatically detects when the sanitizer would remove content, and inform the user? We could do this either client-side or server side.

On the client side, we could have a script to detect when the preview contains disallowed tags and attributes, in accordance with https://meta.codidact.com/posts/277420. This would give live feedback, but could suffer from any discrepancies between the "client sanitizer" and the true sanitizer.

On the server side, we check that the submitted HTML is the same before and after sanitization, and if different, we would inform the users similar to how we inform the user when they are submitting an image without alt text. This would be accurate, but only catchable when the user hits the submit button.

MoshiKoi:
Resolves #1140 

image

Currently waiting for a help article to be created so I can link to that instead of my own post.

### DIFF
--- a/app/assets/javascripts/posts.js
+++ b/app/assets/javascripts/posts.js
@@ -6,6 +6,19 @@ const ALLOWED_ATTR = ['id', 'class', 'href', 'title', 'src', 'height', 'width', 
   'start', 'dir'];
 
 $(() => {
+  DOMPurify.addHook("uponSanitizeAttribute", (node, event) => {
+    const rowspan = node.getAttribute("rowspan");
+    const colspan = node.getAttribute("colspan");
+
+    if (rowspan && Number.isNaN(+rowspan)) {
+      event.keepAttr = false;
+    }
+
+    if (colspan && Number.isNaN(+colspan)) {
+      event.keepAttr = false;
+    }
+  });
+
   const $uploadForm = $('.js-upload-form');
 
   const stringInsert = (str, idx, insert) => str.slice(0, idx) + insert + str.slice(idx);
--- a/app/assets/javascripts/posts.js
+++ b/app/assets/javascripts/posts.js
@@ -163,7 +163,7 @@ $(() => {
 
         $tgt.parents('form')
           .find('.rejected-elements')
-          .toggleClass('hide', removedElements.length === 0)
+          .toggleClass('hide', removedElements.length === 0 && removedAttributes.length === 0)
           .find('ul')
           .empty()
           .append(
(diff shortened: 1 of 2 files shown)


cellio:
Is someone working on that help article?  If not, what do we need -- an edit to the formatting topic, a new topic, something else?

MoshiKoi:
I don't know if anyone is working on it; pinging @codidact/documentation

MoshiKoi:
I'm not sure what exactly is needed; I think any of your suggestions would work fine. I'll leave it to the writers to figure it out 🙂

superplane39:
Help article written; it currently exists on Meta and #1212 adds it to seeds for the network

app/assets/javascripts/posts.js
```diff
@@ -6,6 +6,19 @@ const ALLOWED_ATTR = ['id', 'class', 'href', 'title', 'src', 'height', 'width',
...
+
+    if (rowspan && Number.isNaN(+rowspan)) {
+      event.keepAttr = false;
+    }
+
+    if (colspan && Number.isNaN(+colspan)) {
+      event.keepAttr = false;
+    }
+  });
+
```
Oaphi, line 21: I've added a hook to also sanitize values of the `rowspan` and `colspan` attributes to disallow non-numeric values (might be worth to allow only ints too) - since we have such a powerful tool now, why not?

app/assets/javascripts/posts.js
```diff
@@ -149,6 +162,27 @@ $(() => {
...
+        const removedAttributes = [...new Set(DOMPurify.removed
+          .filter(entry => entry.attribute)
+          .map(entry => [
+            entry.attribute.name + (entry.attribute.value ? `='${entry.attribute.value}'` : ''),
+            entry.from.localName
+          ]))]
+
+        $tgt.parents('form')
+          .find('.rejected-elements')
+          .toggleClass('hide', removedElements.length === 0 && removedAttributes.length === 0)
```
Oaphi, line 179: Fixed a small issue with the notice not displaying when there are no removed elements, but some of the attributes are.

Oaphi (approved):
I've added a small fix & a minor expansion for the PR (someone else might want to review them), so other than that - LGTM! Played around with the changes locally, seems to be working perfectly fine.

cellio:
@MoshiKoi  can you review the latest changes and let us know if you're happy with them?

MoshiKoi:
Looks good to me

PR #1212, opened 2023-10-04
Add advanced formatting help to seeds

superplane39:


Oaphi (approved):
LGTM! I took the liberty to update the HTML to conform to what is specified in code standards (read: slightly linted) and bring the branch up-to-date with develop while I was at it. We probably want to lint the other seeded posts too at some point in the future.

As an additional note (out of scope of this PR, of course), I am thinking whether it'd be beneficial to add examples of the tags being used alongside them.

PR #1220, opened 2023-10-19
added guidance on score values

cellio:
Added a little guidance about scores after seeing [some user confusion](https://meta.codidact.com/posts/290048).

I tried to make "min score" and "max score" links, but those were clickable before and triggered the form field, and that behavior overrides the link.

I also tried to move the help link closer to the form fields, to more clearly "go with" the previous row, but couldn't figure out how.  I tried adding an extra layer of form grouping to include that row plus this link, but it laid out as one row with (now) five elements.  We have a design framework, so I don't think resorting to direct HTML hackery is the right way to go about it.  I think having the link is better than not having it, and having it in a slightly different place would be better than this.

Update: here's a screenshot:

Min Score (0-1), Max Score (0-1), link below fields

### DIFF
--- a/app/views/search/_filters.html.erb
+++ b/app/views/search/_filters.html.erb
@@ -25,10 +25,10 @@
   <% if user_signed_in? && defined? @category %>
     <%= label_tag :save_as_default, 'Is default for this category?' %>
     <%= check_box_tag :save_as_default, @category.id, false, { class: 'filter-is-default form-checkbox-element' } %>
-    <% end %>
+  <% end %>
   <div class="form-group-horizontal">
     <div class="form-group">
-      <%= label_tag :min_score, 'Min Score (8-1)', class: "form-element" %>
+      <%= label_tag :min_score, 'Min Score (0-1)', class: "form-element" %>
       <%= number_field_tag :min_score, @active_filter[:min_score],
           min: 0, max: 1, step: 0.01, class: 'form-element form--filter',
           data: { name: 'min_score' } %>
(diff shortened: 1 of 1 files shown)


app/views/search/_filters.html.erb
```diff
@@ -25,16 +25,16 @@
   <% if user_signed_in? && defined? @category %>
     <%= label_tag :save_as_default, 'Is default for this category?' %>
     <%= check_box_tag :save_as_default, @category.id, false, { class: 'filter-is-default form-checkbox-element' } %>
-  <% end %>
+    <% end %>
   <div class="form-group-horizontal">
     <div class="form-group">
-      <%= label_tag :min_score, 'Min Score', class: "form-element" %>
+      <%= label_tag :min_score, 'Min Score (8-1)', class: "form-element" %>
```
michaelblyons, line 31: Maybe 0-1 instead of 8-1?

app/views/search/_filters.html.erb
```diff
@@ -25,16 +25,16 @@
   <% if user_signed_in? && defined? @category %>
     <%= label_tag :save_as_default, 'Is default for this category?' %>
     <%= check_box_tag :save_as_default, @category.id, false, { class: 'filter-is-default form-checkbox-element' } %>
-  <% end %>
+    <% end %>
```
michaelblyons, line 28: Accidental dedent?

app/views/search/_filters.html.erb
cellio: Whoops!  Typo + vision failure, thanks.

app/views/search/_filters.html.erb
cellio: Looks like all the other `<% end %>` lines are aligned with the statements in the block and this was the outlier.  I didn't notice this change, but I think, based on the rest of the file, it was inconsistent before.  (I don't care, so if someone wants to speak authoritatively on our coding standards, just let me know what to do.)

Oaphi (approved):
LGTM

PR #1221, opened 2023-10-22
Translation files must have a top-level locale key (fixes #1210)

### LINKED ISSUE #1210: "Translation missing" instead of an error message
meta:288508

A user hit a rate limit and got this error message:

translation missing + key

That's not right.  I can't tell whether we have a missing entry in a translation file (where?) or a problem with how that error is rendered.  I tried grepping the code to find where these strings are defined, but failed.

Oaphi:
This minor PR fixes the issue with translation strings missing (all config YAML files for those must start with a top-level key denoting the locale, which were missing, causing the "translation missing" error to pop up).

### DIFF
--- a/config/locales/strings/en.rate_limit.yml
+++ b/config/locales/strings/en.rate_limit.yml
@@ -1,10 +1,11 @@
-rate_limit:
-  new_user_posts: >
-    You may only post :count :level posts per day. Once you have some well-received posts, your limit will increase.
-  posts: >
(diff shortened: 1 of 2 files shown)


cellio:
Looks fine to me, but I don't know enough about how translation works (or how to test this) to sign off myself, hence the request for other reviewers.  This is a pair of one-line changes, so should be a quick review for people who know what they're doing. :-)

Taeir (approved):

PR #1222, opened 2023-10-22
Prevent keyboard events with modifier keys from triggering shortcuts

### LINKED ISSUE #1206: Keyboard shortcuts shouldn't trap ctrl-(letter)
meta:288400

Keyboard shortcuts allow single-letter navigation, like `n` for "new post".  Browsers have many ctrl/command-letter shortcuts, like `ctrl-n` for "new window".

Typing `ctrl-n` triggers the qpixel shortcut, along with the browser one.  But the qpixel shortcut isn't `ctrl-n`; it's `n`.  Can we not react if there's a modifier key?

Oaphi:
Closes #1206.

Additional issue I've noticed is that the "a" shortcut no longer works (causing a runtime error) as there's no element with the `answer_body_markdown` id (post fields are currently all id'd as `post_body_markdown`). Fixing that is outside of the scope of the PR - let's tackle it separately.

### DIFF
--- a/app/assets/javascripts/keyboard_tools.js
+++ b/app/assets/javascripts/keyboard_tools.js
@@ -77,8 +77,27 @@ $(() => {
       }
     });
 
+    /**
+     * Checks common modifier states on a given keyboard event
+     * @param {KeyboardEvent} e
+     * @returns {boolean}
+     */
+    const getModifierState = (e) => {
+      return !!e.altKey || !!e.ctrlKey || !!e.metaKey || !!e.shiftKey;
+    };
+
+    /**
+     * Handles the "home" keyboard state
+     * @param {KeyboardEvent} e
+     */
     function homeMenu(e) {
-      if (e.key === "?") {
+      const isHelp = e.key === "?";
+
+      if (!isHelp && getModifierState(e)) {
+        return;
+      }
+
+      if (isHelp) {
         _CodidactKeyboard.dialog(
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/keyboard_tools.js
```diff
@@ -77,8 +77,27 @@ $(() => {
       }
     });
 
+    /**
+     * Checks common modifier states on a given keyboard event
+     * @param {KeyboardEvent} e
+     * @returns {boolean}
+     */
+    const getModifierState = (e) => {
```
Taeir, line 85: Perhaps `isModifierPressed` is a better name for this function?

Taeir (commented):
I made a minor suggestion for a function rename, but otherwise the changes look good to me!

app/assets/javascripts/keyboard_tools.js
Oaphi: @Taeir probably, but I was trying to align the signature with the built-in API (which, unfortunately, lacks the required functionality of "is any modifier pressed?" as it can only check by key): https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState. I am not particular as to what to name it - if you, given the context, still think it's better to rename, will gladly do so!

Taeir: Ah, in that case I think this is fine :)

Taeir (approved):

Issue #1209, opened 2023-10-03
Make it easier to show question titles when linking to them


cellio:
meta:288499  (third item)

Sometimes people want to include, in the body of a post or comment, a link to another Codidact post.  The page URL is of the form `site/posts/###`, so the URL itself doesn't tell you what the post is.  Can we improve on this?

- SE converts such links at render time to show the post title instead of the raw URL.  Is that hard?  Is it fragile?
- Maybe we can instead handle it in the editor somehow.  Button?  Automatic, but at post-creation time not render time so it become static text?  Something else?   (This only helps with the post editor, not comments or the next bullet.)
- The linked meta post requests a way to generate the markdown and copy it to the clipboard.  This could then be pasted into a post, or it could be pasted somewhere else that supports Markdown, like social media.  We would need to work out what the UI should be (some suggestions on the linked post).

How should we approach this general class of requests?  What's practical?

PR #1223, opened 2023-10-22
Post titles improvements as per #1208, #1209

### LINKED ISSUE #1208: Make post titles (the h1) links and don't truncate page title
meta:288499

We've gotten this request several times; this is one recent one.  People are used to being able to right-click on a title to get its URL or open the page in a new tab.  (I don't just mean on SE; many blogs also have this property.)  We do have "copy link" under the post, which is also important (and the only way to link an answer), but could we also add the link in the h1 itself?

A related request from the same meta post: in the page `<title>`, we truncate the text if it's long.  If we didn't do this, then apparently some browser extensions for constructing markdown links would work better.  Since we don't automatically render page titles when somebody links to another Codidact post, it seems helpful to provide this.  I think this is in the same "neighborhood" as the H1 request, but if it should be split off, we can do that.

I'm going to spin off the third request from this meta post into its own issue (https://github.com/codidact/qpixel/issues/1209).

Oaphi:
Closes #1208 and #1209

### General improvements

- post titles are now self-linked;
- <title> element no longer truncates post titles to 50 characters (technically, we don't need to truncate at all there, but I opted for max title length truncation just in case);

### Copying links

Below are screenshots of the initial implementation (raw URL & Markdown link copying).

Per-post (expanded) view:

Screenshot from 2023-10-22 07-31-18

Post history view:

Screenshot from 2023-10-22 13-59-48

The above is subject to change over the course of fleshing the idea out.

TODO:

- [x] create a partial for rendering a reusable component;
- [x] switch expanded post view to the partial;
- [x] switch post history view to the partial;
- [X] at least show a link that can be manually copied if JS is disabled;
- [X] decide on the appropriate title override for Markdown share links for post history;
- [X] update Code Golf's script to not rely on `.js-permalink`*;
- [ ] add rich link copying;

---
\* Instead of updating the script, I ended up using a hidden anchor tag for scripts to be able to hook to it as usual. A better idea, probably, is to set a data attribute on the button, though.


Oaphi:
After some deliberation, I don't think there's a need for a separate rich link option (raw link & markdown cover all common cases). The rest should work as described.

ArtOfCode- (approved):

Issue #1207, opened 2023-10-01
When updating live votes, also update the tooltip


cellio:
meta:288443

When you vote on a post, the page is updated client-side to show your vote, but this update doesn't include the tooltip with the post score.  Could it?  (It is updated after page reload.)

PR #1224, opened 2023-10-23
Dynamically update post score tooltip (#1207)

Oaphi:
This PR ensures the post score tooltips are updated on vote success. I opted for adding a new `score` field to the success response (re-finding the post is needed to get the updated score after the `recalc_score` method is run - if anyone has a better idea for doing so, I'd appreciate that). 

[Screencast from 23.10.2023 04:16:50.webm](https://github.com/codidact/qpixel/assets/34833340/8a6455de-801b-46fc-b4c2-2a9a10f9fb7a)

### DIFF
--- a/app/controllers/votes_controller.rb
+++ b/app/controllers/votes_controller.rb
@@ -48,8 +48,7 @@ class VotesController < ApplicationController
               vote_id: vote.id,
               upvotes: post.upvote_count,
               downvotes: post.downvote_count,
-              score: post.score
-            }
+              score: post.score }
 
     render json: state
   end
@@ -70,8 +69,7 @@ class VotesController < ApplicationController
       render json: { status: 'OK',
                      upvotes: post.upvote_count,
                      downvotes: post.downvote_count,
-                     score: post.score
-                   }
+                     score: post.score }
     else
       render json: { status: 'failed', message: vote.errors.full_messages.join('. ') }, status: :forbidden
     end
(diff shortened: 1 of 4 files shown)


app/controllers/votes_controller.rb
```diff
@@ -41,9 +41,14 @@ def create
 
     AbilityQueue.add(post.user, "Vote Change on ##{post.id}")
 
+    post = Post.find(params[:post_id])
```
Taeir, line 44: You can use the reload method to reload an object from the database, which I think would be preferable for this case (still triggers the same query though).

```suggestion
    post.reload
```

app/controllers/votes_controller.rb
```diff
@@ -58,8 +63,13 @@ def destroy
     end
 
     if vote.destroy
+      post = Post.find(post.id)
```
Taeir, line 66:
```suggestion
      post.reload
```

Taeir (commented):
Code looks good. I have no good solution for avoiding requerying the database for the updated score, but I do suggest to use the reload function to get the updated post attributes instead of using find to grab a new object.

app/controllers/votes_controller.rb
Oaphi: @Taeir yeah, `.reload` is cleaner than `Post.find(post.id)`. I don't have a better solution for this either at this point other than recalculating the score manually for the affected post without actually triggering `recalc_score` - should save us a query - this could work, I think?

Taeir: Aah, I just looked at `recalc_score` and realized why Rails doesn't magically fix things for us. Its because we literally execute a raw SQL query there...

I guess that is done to let the database handle atomicity?

Taeir (approved):

PR #1225, opened 2023-10-23
Prevent unnamed subscriptions in the subscriptions dashboard

### LINKED ISSUE #1191: Prevent unnamed subscriptions
meta:288220

When creating a subscription, you supply a name.  We allow this name to be empty, which makes the UI for managing subscriptions confusing.  We should either require a name or automatically generate one if the name is blank ("unnamed #" or similar).

Oaphi:
Closes #1191

Here's how unnamed subscriptions look like in the dashboard with the update (the only named one is "Test name" [it's of type "user" and was created the 2nd, hence 1 & 3 for unnamed user subscriptions]):

image

Not sure if how I approached the querying of row number is ideal, though, thoughts?


app/models/subscription.rb
```diff
@@ -10,6 +10,18 @@ class Subscription < ApplicationRecord
 
   validate :qualifier_presence
 
+  def summary
+    unless name.empty?
```
Taeir, line 14: blank? works with nil (NULL), empty string, and strings that consist of only whitespace.

```suggestion
    unless name.blank?
```

app/views/subscriptions/index.html.erb
```diff
@@ -5,7 +5,7 @@
 
 <% @subscriptions.each do |sub| %>
```
Taeir, line 6: I would change this line to do what you want to achieve:

```ruby
<% @subscriptions.order(created_at: :asc).each_with_index do |sub, index| %>
  <details data-sub-id="<%= sub.id %>">
    <summary><%= sub.summary(index + 1) %></summary>
    <p>Subscription to <%= phrase_for sub.type, sub.qualifier %>, emailed every <%= pluralize(sub.frequency, 'day') 
```

(Or similar)

Now index+1 will be the nth subscription and can be used in the view, and you wouldn't need to do the query.

Oaphi: Applied your other suggestion as per our discussion in Discord - adjusted it a little to make sure the content of the subscription is accessible in the loop, but it's still pretty much a `flat_map` -> `each_with_index` -> `map` -> `sort_by` -> `each` (xD)  chain.

app/views/subscriptions/index.html.erb
```diff
@@ -3,10 +3,19 @@
...
-    <summary><%= sub.name %></summary>
-    <p>Subscription to <%= phrase_for sub.type, sub.qualifier %>, emailed every <%= pluralize(sub.frequency, 'day') %>.</p>
+<% @subscriptions.group_by(&:type).flat_map { 
+  |type, subs| subs.each_with_index.map { 
+    |sub, index| [
+      sub.created_at, 
+      [sub.name.present? ? sub.name : "#{type.capitalize} subscription #{index + 1}",sub]
+    ] 
+  } 
+}.sort_by { |a| a }.map { |_, v| v }.each do |name, sub| %>
```
ArtOfCode-, line 13: style: for multiline blocks, prefer `do...end` over braces:

```suggestion
<% @subscriptions.group_by(&:type).flat_map do |type, subs|
  subs.each_with_index.map do |sub, index|
    [
      sub.created_at, 
      [sub.name.present? ? sub.name : "#{type.capitalize} subscription #{index + 1}",sub]
    ] 
  end
end.sort_by { |a| a }.map { |_, v| v }.each do |name, sub| %>
```

ArtOfCode- (changes requested):

ArtOfCode- (approved):

PR #1227, opened 2023-10-23
Move 'start new comment thread' button under listed threads

### LINKED ISSUE #1072: Input box is not on screen after pressing "Start new comment thread" with expanded comments
https://meta.codidact.com/posts/288456

Expanded comment threads move the input box for a new comment thread off the screen. Could either move the input to above existing comment threads, or the "start new" and "follow" buttons to below them.

Oaphi:
Closes #1072

This is how the proposed positioning looks like (since there's now only one button on the top, I suggest we move it on the same line as the thread count):

Screenshot from 2023-10-23 09-38-12

Locked post state with admin permissions:

Screenshot from 2023-10-23 09-43-13

With no preexisting comment threads:

Screenshot from 2023-10-23 09-44-41

### DIFF
--- a/app/assets/stylesheets/comments.scss
+++ b/app/assets/stylesheets/comments.scss
@@ -65,6 +65,17 @@
   font-style: italic;
 }
 
+.post--comments-header {
+  align-items: center;
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: 0.75rem;
+}
+
+.post--comments-container {
+  margin-bottom: 1rem;
+}
+
 .post--comments-thread.is-inline {
   padding: 0.5rem 0.25rem;
   display: flex;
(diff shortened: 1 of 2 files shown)


Taeir (approved):
Definitely better than before (when a thread is expanded). Though I think that comments in general may benefit from a bit of a redesign. I feel they take up a lot of real-estate/show a lot of clutter for what they are and are used for.

Oaphi:
> Definitely better than before (when a thread is expanded). Though I think that comments in general may benefit from a bit of a redesign. I feel they take up a lot of real-estate/show a lot of clutter for what they are and are used for.

@Taeir yeah, I believe they are in need of a redesign - was out of scope of the PR, so I only tweaked alignment a little. Happy to put it on my TODO list, though 😇

PR #1228, opened 2023-10-23
Header buttons should have tooltips set on the outer link element

### LINKED ISSUE #1044: title text on buttons only works for the icon, not the whole button
https://meta.codidact.com/posts/287698

The buttons in the top bar (like inbox) consist of images surrounded by whitespace, with the whitespace being part of the clickable area of the button.  These buttons have tooltips, but they're only active for the icon, not the whole button.  Hovering anywhere on the button changes the rendering of the button, though, so it's a little confusing to not get the tooltip everywhere.

I hope this is just a matter of moving the attachment for the tooltip "out" one level, from the icon to the containing `div`.

Oaphi:
Closes #1044 

This PR moves title text of the header buttons to the outer anchor tag ensuring tooltips are shown not only when hovering the icons. Also adds a list of missing button titles for other header buttons.

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -22,10 +22,12 @@
     </div>
     <nav class="header--menu">
       <% unless @community.is_fake %>
-        <%= link_to users_path, class: 'header--item' do %>
+        <%= link_to users_path, class: 'header--item', title: 'Open users list' do %>
           <i class="fas fa-fw fa-users"></i> Users
         <% end %>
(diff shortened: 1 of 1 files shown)


cellio:
Thanks for finding and fixing the missing ones while you were in there.

cellio (approved):
I checked out this branch, ran it, and ran the mouse over the things with new/moved tooltips.  All looks good and the code change is straightforward, so I'm approving.

PR #1230, opened 2023-10-30
Post history pagination & permalinks

### LINKED ISSUE #1229: Persistent references to post history entries & pagination
**Describe the bug**

The issue is two-fold:

1. If the number of history entries for a given post exceeds the number of items per page, there's no UI to navigate to the rest of the entries.
2. Links to history entries rely on indices which makes them not true permalinks as what constitutes an Nth entry is determined on the fly and is page-relative.

**To Reproduce**

Opening any history view for a post with more than 20 history entries, for example, https://codegolf.codidact.com/posts/287238/history.

**Expected behavior**

1. Pagination component present.
2. Immutable links to history entries.

**Additional context**

Special thanks to @trichoplax for noticing the issue.

Oaphi:
Closes #1229


trichoplax:
I'm not familiar with Rails' pagination functionality so I don't know how to completely fix this, but just in case it helps I believe the numbering of the edits can be calculated using information which already seems to be available:

- `@count`: the total number of edits
- `per_page`: the number to show per page in the pagination
- `@page`: the currently displayed page

If you know the number of an item within the current page (from 1 to per_page), say `number_in_page`, then the number to display (the edit number) can be calculated as:

```ruby
@count - (number_in_page - 1) - (@page - 1) * per_page
```

The 1s are due to `number_in_page` and `@page` being 1-indexed rather than 0-indexed. They can be omitted if those variables are actually 0-indexed.

PR #1235, opened 2023-11-14
clarified intent for redaction (use sparingly for PII etc)

cellio:
We've seen some uses of redaction where there was nothing sensitive to hide, which is understandable because we didn't give guidance in the UI.

### DIFF
--- a/config/locales/strings/en.posts.yml
+++ b/config/locales/strings/en.posts.yml
@@ -49,7 +49,7 @@ en:
     redact_label: >
       Redact
     redact_explanation: >
-      Redact original content by hiding the previous versions from history? Use sparingly, for example to hide passwords or personal information.
+      Redact original content by hiding the previous versions from history? Use only for private information such as passwords or personally identifiable information.
     licence_label: >
       License
     unsaved_changes_confirmation: >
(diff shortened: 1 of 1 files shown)


config/locales/strings/en.posts.yml
```diff
@@ -49,7 +49,7 @@ en:
     redact_label: >
       Redact
     redact_explanation: >
-      Redact original content by hiding the previous versions from history?
+      Redact original content by hiding the previous versions from history? Use sparingly, for example to hide passwords or personal information.
```
trichoplax, line 52: I've confirmed that the new wording shows up correctly in the user interface, so my only question is whether to fine tune the wording. It's already a big improvement, but I wonder if some people would read "Use sparingly ... " as "Use your own judgement, the rest of this sentence is just rough guidance".

We could be more imperative, such as "Use only for passwords and personally identifiable information". The problem with this is that we've moved from a list of examples to an exhaustive list, so we'd have to add in any other reasons that might ever apply.

trichoplax (approved):
I'm happy from a code perspective (I've tested it works). Although I've added a comment about the wording, I've also approved as the wording is a definite improvement, and my suggestion might be worse.

If someone finds a better wording later we can always make another pull request, so I'm happy for this to be merged if there are no immediate suggestions.

config/locales/strings/en.posts.yml
cellio: Good point.  I've changed that sentence to: "Use only for private information such as passwords or personally identifiable information."  That's a little more verbose but keeps the "example" framing.

PR #1236, opened 2023-11-21
fixed typo in help seed

cellio:
One-character typo in the seed; already fixed in our live help.


Oaphi:
@cellio I threw in a couple more fixes: there was some stray markdown & an Oxford comma was missing.

Issue #1106, opened 2023-07-09
User profile: access all posts without having to scroll all the way down


cellio:
https://meta.codidact.com/posts/287867

The bottom of the list of posts on the user profile page has a button to see all posts.  Sorting is gated behind that button, so it would be nice to be able to get there more easily.  Can we make the "Posts" h2 on that page a link to the full posts list, the same target as the button at the bottom?

The request contains an implicit request for sorting controls, and I've long had aspirations of being able to sort, on the profile page directly, by score, age, category, and post type.  That's a much bigger task that would involve some UX analysis to make it not cumbersome.  In the meantime, we have the "all posts" page as a separate page; let's make it easier to jump to particularly on a phone.

The relevant code is in `app/views/users/show.html_erb`; I'm just not quite sure how to make the link instead of the button.

PR #1237, opened 2023-11-21
on profile page, repeat button for all posts at the top of the list

cellio:
On the user profile page, copies the "show all" button from the bottom of the post list to the top as well.  We could make the heading a link instead of having the button, but since it's the same operation maybe we want the same UI presentation.  I'm open to changing that, though we only want the link if there _are_ posts, so it complicates the view a little.  I tried putting a link next to or immediately below the heading instead, but it looked ugly -- maybe just a matter of spacing.

Fixes https://github.com/codidact/qpixel/issues/1106.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -85,6 +85,9 @@
       <% if @posts.size == 0 %>
         <p><span dir="ltr"><%= rtl_safe_username(@user) %></span> hasn't posted anything yet.</p>
       <% else %>
+        <%= link_to user_posts_path(@user), class: "button is-muted", 'aria-label': "View all posts by #{rtl_safe_username(@user)}" do %>
+          See all &raquo;
+        <% end %>
         <div class="item-list user-profile-post-list">
           <% @posts.each do |a| %>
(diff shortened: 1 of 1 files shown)


Oaphi (changes requested):
How about we do something like this?

Screenshot from 2023-11-24 22-49-48

Example implementation:

show.html.erb

```ruby
      <div class="user-profile-heading">
        <h2 class="has-margin-0">Posts</h2>
        <% if @posts.size > 0 %>
          <%= link_to user_posts_path(@user), class: "button is-muted", 'aria-label': "View all posts by #{rtl_safe_username(@user)}" do %>
            See all &raquo;
          <% end %>
        <% end %>
      </div>
      <% if @posts.size == 0 %>
        <p><span dir="ltr"><%= rtl_safe_username(@user) %></span> hasn't posted anything yet.</p>
      <% else %>
```

users.scss

```postcss
.user-profile-heading {
  padding: 0.5em 0 0.5em 0.5em;
  border-bottom: 1px solid #ddd;
  margin-bottom: 0;
  display: flex;
  justify-content: space-between;
}
```

cellio:
> How about we do something like this?

That would be much better!  I couldn't figure out how to lay that out.  Thanks.

Issue #1251, opened 2023-11-26
Let people with Edit Tags create new tags from the tags page


cellio:
meta:289114

On the tags page, moderators see a "new" button so they can directly create tags without attaching them to posts.  This is helpful for tag curation, for example defining hierarchies.

In the interest of empowering responsible community members more, let's make that button available to anyone with Edit Tags.  Note that until we address some other issues with tags, Edit Tags has to be manually granted by moderators.  So this effectively allows moderators to delegate tag-curation powers including defining them, not just editing their wkis.

PR #1254, opened 2023-11-27
allow users with the edit-tags ability to create new tags from the tags list

cellio:
Currently only moderators and admins get the "new" button on the category tags page to be able to create new tags directly.  This also gives it to users with the "edit tags" ability.

Fixes https://github.com/codidact/qpixel/issues/1251 .

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -2,7 +2,8 @@ class TagsController < ApplicationController
   before_action :authenticate_user!, only: [:new, :create, :edit, :update, :rename, :merge, :select_merge]
   before_action :set_category, except: [:index]
   before_action :set_tag, only: [:show, :edit, :update, :children, :rename, :merge, :select_merge, :nuke, :nuke_warning]
-  before_action :verify_moderator, only: [:new, :create, :rename, :merge, :select_merge]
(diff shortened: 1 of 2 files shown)


app/controllers/tags_controller.rb
```diff
@@ -213,4 +214,21 @@ def tag_params
   def exec_sql(sql_array)
     ApplicationRecord.connection.execute(ActiveRecord::Base.sanitize_sql_array(sql_array))
   end
+
+  def verify_tag_editor
+    if !user_signed_in? ||
+       !(current_user&.privilege?(:edit_tags) || current_user&.is_moderator || current_user&.is_admin)
```
ArtOfCode-, line 220: suggestion: use an `unless` block instead. This condition is hard to read naturally; switching the conditional to `unless` makes it a bit easier:

```suggestion
    unless user_signed_in? && (current_user.privilege?(:edit_tags) || current_user.is_moderator || current_user.is_admin)
```

ArtOfCode- (changes requested):

app/controllers/tags_controller.rb
cellio: Applied and tested.

PR #1257, opened 2023-11-27
owners can approve edits and thus should not see the notice saying they can't

cellio:
Fixes meta:290299

### DIFF
--- a/app/views/suggested_edit/show.html.erb
+++ b/app/views/suggested_edit/show.html.erb
@@ -5,7 +5,7 @@
 
 <h1>Review Suggested Edit</h1>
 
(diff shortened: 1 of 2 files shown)


app/views/suggested_edit/show.html.erb
```diff
@@ -5,7 +5,7 @@
 
 <h1>Review Suggested Edit</h1>
 
-<% if params[:show_decided] != '1' && !current_user&.privilege?('edit_posts') %>
+<% if params[:show_decided] != '1' && !current_user&.privilege?('edit_posts') && !may_decide %>
```
Oaphi, line 8: Can't we simply replace `!current_user&.privilege?('edit_posts')` with `!may_decide` as it does the same thing?

PR #1268, opened 2023-12-05
Display existing flags in flag dialog

### LINKED ISSUE #1264: When you have an outstanding flag on a post, show an indicator on that post.
meta:289233

If you raise a flag on a post and later view that post while the flag remains active, you get no indication of the flag.  You also don't get any indicator if you start to flag it again.  This can result (and has resulted) in the same person flagging the same post more than once due to not remembering.

Curators see an indicator like this:

links under post include Show 1 flag

Could regular users who are not curators, who have active flags, see "Show my flags" there?  We need to make sure that this would show _only_ that user's flags, and showing them should not offer the "accept", "decline", and "escalate" options that curators have.

A curator who can see other people's flags should not also see "See my flags"; that'll be covered in "Show N flags", which include your own.  We don't want two controls in that case.

Users can see all of their flags (active and handled) via a link on the profile page, but the goal here is to put the "you've already flagged this, and here's what you said" indicator where the user will notice it before accidentally re-flagging.

ArtOfCode-:
Fixes #1264

image

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -144,18 +144,18 @@ class PostsController < ApplicationController
       return not_found
     end
 
-    #@post = @post.includes(:flags, flags: :post_flag_type)
+    # @post = @post.includes(:flags, flags: :post_flag_type)
     @children = if current_user&.privilege?('flag_curate')
                   Post.where(parent_id: @post.id)
                 else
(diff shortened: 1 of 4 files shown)


cellio:
LGTM but I don't know if I qualify as an approver here.  I did look at the code, and the screenshot here is also helpful.

PR #1271, opened 2023-12-09
Validation for edit comment length

### LINKED ISSUE #1258: Post edit comment length is not validated at all
**Describe the bug**

Length of edit comments (both privileged edits and suggested edits) is not validated neither client-side nor server-side. As a result, editors can submit up to 65 535 (privileged edits) or 255 (suggested edits) characters. 

Additionally, because there are no validations, if a user attempts to submit more characters than the limits mentioned above, they'll get a raw query error due to the DB rejecting the value for the `comment` column.

**To Reproduce**

1. Log in as an unprivileged user.
2. Make an edit suggestion to any post.
3. Include more than 255 characters in the comment field.
4. Observe the raw query error.

---

There definitely needs to be _some_ limit (and an accompanying character counter client-side), although the _exact_ limit is up for debate (it can be made configurable, but it still leaves us with the question of what would be the default). I propose a soft cap of 150 characters (similar to commit messages) with a somewhat higher hard cap (200 or 255).

Oaphi:
Closes #1258

---
- [x] character count for edit comment fields;
- [x] site setting for configuring max edit comment length;
- [x] direct edit comment length validation for single post updates;
- [x] suggested edit comment length validation;
- [x] internationalization for the error message;
- [x] direct edit comment length validation for push-to-network updates;
- [x] fix updates to Help Center with network push option turned on;
- [x] add ability to resolve edit validation issues when approving pending edits;
---

Validation error display with > 1 max characters limit:

Screenshot from 2024-01-04 00-07-28

Site setting for updating the new limit (255 by default):

Screenshot from 2024-01-04 00-08-55

Validation error display with = 1 max character limit:

Screenshot from 2024-01-04 00-10-07

Validation error display when approving a suggested edit fails due to validation:

Screenshot from 2024-01-05 19-27-53
---

Additional changes include:

- formatting settings for VSCode users;
- fix for incorrect char count value on initial render when editing post titles;
- fix for `min_title_length` view helper for Help Center posts (min length for them is always 1);


cellio:
Question: I see the addition of the seed to define the length, but I don't see a change to do the checking.  Were we already checking but didn't have a default value set (so that check was falling back to the DB limit)?  How does the limit in the seed get to code that uses it?

app/controllers/posts_controller.rb
```diff
@@ -160,6 +160,67 @@ def show
...
+  # @param edit_post_params [ActionController::Parameters] edit parameters
+  # @param body_rendered [String] new post body
+  # @return [Boolean] status of the operation
+  def do_update(post, user, edit_post_params, body_rendered)
+    post.update(edit_post_params.merge(body: body_rendered,
+                                       last_edited_at: DateTime.now,
+                                       last_edited_by: user,
+                                       last_activity: DateTime.now,
+                                       last_activity_by: user))
+  end
```
ArtOfCode-, line 175: question: what's the benefit of having this method over just calling `post.update` manually?

app/controllers/posts_controller.rb
```diff
@@ -160,6 +160,67 @@ def show
 
   def edit; end
 
+  # Attempts to update a given post
+  # @param post [Post] post the user is attempting to update
+  # @param user [User] user attempting to update the post
+  # @param edit_post_params [ActionController::Parameters] edit parameters
+  # @param body_rendered [String] new post body
+  # @return [Boolean] status of the operation
+  def do_update(post, user, edit_post_params, body_rendered)
```
ArtOfCode-, line 169: suggestion: I'd refactor this signature to the following for Ruby conventions:

```suggestion
  def do_update(post, user, body_rendered, **edit_post_params)
```

This means you can call it as `do_update(post, user, body, some: 'data', more: 'data')` instead of `do_update(post, user, { some: 'data', more: 'data' }, body)`. Little bit nicer.

app/controllers/posts_controller.rb
```diff
@@ -160,6 +160,67 @@ def show
...
+
+  # Attempts to update a given post network-wide. The update is manual to avoid
+  # skipping validations and fail early if at least one validation fails.
+  # @param post [Post] post from which the network push is initiated
+  # @param posts [ActiveRecord::Result] network posts to be updated
+  # @param user [User] user attempting to push updates to network
+  # @param edit_post_params [ActionController::Parameters] edit parameters
+  # @param body_rendered [String] new post body
+  # @return [Boolean] status of the operation
+  def do_update_network(post, posts, user, edit_post_params, body_rendered)
```
ArtOfCode-, line 185: suggestion: likewise on the signature

app/controllers/posts_controller.rb
```diff
@@ -160,6 +160,67 @@ def show
...
+
+  # Checks if a given user can directly update a given post
+  # @param post [Post] updated post (owners can unilaterally update)
+  # @param user [User] user attempting to update the post
+  # @param post_type [PostType] type of the post (some are freely editable)
+  # @return [Boolean]
+  def can_update(post, user, post_type)
+    user.privilege?('edit_posts') || user.is_moderator || user == post.user || \
+      (post_type.is_freely_editable && user.privilege?('unrestricted'))
+  end
```
ArtOfCode-, line 222: suggestion: these should probably be methods on the `Post` or `User` model rather than in a controller. The `do_update` methods could as well, to be fair.

app/helpers/posts_helper.rb
```diff
@@ -28,9 +28,14 @@ def max_body_length(_category)
   end
 
   # @param category [Category, Nil]
+  # @param post_type [PostType]
   # @return [Integer] the minimum length for post titles
-  def min_title_length(category)
-    category&.min_title_length || 15
+  def min_title_length(category, post_type)
+    if ['HelpDoc', 'PolicyDoc'].include?(post_type.name)
```
ArtOfCode-, line 34: suggestion: use your new `PostType#system?` method?

ArtOfCode- (changes requested):

.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
Oaphi: Actually, I specifically wanted to push those settings as they enforce conformance with the standard in terms of spacing at least - in light of that, do you think we should?

app/helpers/posts_helper.rb
Oaphi: Yeah, good idea, forgot I even added it!

app/controllers/posts_controller.rb
Oaphi: Frankly, not much at this point. Initially, I expected both `do_update` and `do_update_network` to include history updates, but it turned out to not be desirable. Right now it's like that mostly for consistency with `do_update_network` that _does_ do more heavy lifting + to ease understanding of what the `update` method does. If that's fine with you, I'd keep it that way as its complexity is likely to grow?

app/controllers/posts_controller.rb
Oaphi: Yeah, let's do that (same with the `do_update_network`), keep forgetting the double splat operator exists, a moment.

.vscode/settings.json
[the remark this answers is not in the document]
ArtOfCode-: I'd still avoid it, I think - I don't want to prescribe other people's editor settings for them.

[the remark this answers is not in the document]
Oaphi: Fine by me if you think so - removed

app/controllers/posts_controller.rb
Oaphi: Yup, they should definitely not be here, an artifact of abstracting the logic from the `update` method (otherwise, with the changes it starts to fail the cyclomatic complexity rule of RuboCop [which is fair, the method is a heavy one]). Done and done, with a couple of unit tests as well.

Oaphi: That said, I kept the `do_update` methods on the controller _for now_ as I am a bit worried about shuffling it around right now. I'd like to tackle the move separately later as I also have a couple of other improvements in mind (trying not to scope creep this one).

ArtOfCode- (approved):

PR #1278, opened 2023-12-26
add total pending edits to dashboard if you can see them

cellio:
meta:290471

Adds the number of pending edits to the dashboard and links, for users who can review edits.  Thanks @Oaphi for getting me started and @MoshiKoi for the clue about per-category tallies.

screen shot

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -133,7 +133,7 @@ class User < ApplicationRecord
       return false
     end
 
-    cu = community_users.find { |cu| cu.community_id == community_id && cu.id == id }
+    cu = community_users.find { |cuf| cuf.community_id == community_id && cuf.id == id }
 
     if cu.nil?
       return false
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 # Represents a user. Most of the User's logic is controlled by Devise and its overrides. A user, as far as the
 # application code (i.e. excluding Devise) is concerned, has many questions, answers, and votes.
 class User < ApplicationRecord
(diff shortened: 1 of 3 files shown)


app/views/application/dashboard.html.erb
```diff
@@ -27,10 +27,18 @@
               <% if user_signed_in? && cat.new_posts_for?(current_user) %>
                 <span class="badge is-status" title="unread activity in category"></span>
               <% end %>
-            <% end %>
+	      <% if current_user&.has_ability_on(c.id, 'edit_posts') %>
+	      <% sug_edits = SuggestedEdit.where(post: Post.undeleted.where(category: cat), active: true).count %>
+	      <% if sug_edits != 0 %>
+	      &nbsp;&nbsp;(<%= sug_edits %> pending edits)
```
cellio, line 33: I admit this is ugly.  What's the _proper_ way to get some breathing room around the new-posts badge before diving into the edits text?

cellio, line 33: Pluralization fix?

cellio:
Ok I think this is really ready for review now.  No more hard-coded whitespace or other abominations. :-)

app/models/user.rb
```diff
@@ -122,6 +122,30 @@ def is_admin
...
+
+    if cu.nil?
+      return false
+    end
+
+    if cu.is_moderator || cu.privilege('mod')
+      return true
+    end
+
+    UserAbility.where(community_user_id: cu.id, ability_id: ability.id, is_suspended: false).exists?
```
ArtOfCode-, line 146: At a first glance, this method feels heavy to me - Rails is usually good at doing this kind of thing in a couple of lines.

We want this method to return true if (correct me if I'm wrong):

 * user is a global mod or admin
 * user is a community mod or admin
 * user has specified ability on specified community

I'd probably go about this by grabbing the community user first, then you can probably do the rest as one or two conditionals. Something like:

```ruby
def has_ability_on(community_id, ability_internal_id)
  # where instead of find so that we're using the database rather than Ruby
  cu = community_users.joins(:community).where(communities: { id: community_id }).first
  if is_admin || is_global_moderator || cu.is_moderator # can't remember the names of all these checks - will need checking!
    true
  else
    UserAbility.joins(:ability).where(community_user_id: cu.id, is_suspended: false, ability: { internal_id: ability_internal_id }).exists?
  end
end
```

(Fair warning: I haven't tested any of that, so if you do change it it'll need testing!)

cellio: Updated and tested with the following user types: none (logged out), regular (not editor), editor, moderator, admin.  All behaved as expected.

What I have not been able to test in a dev environment is multiple communities.  I don't know how to set that up when everything runs off of localhost not DNS.  I assume some port-wrangling would do it, but maybe we could test this on the dev server?

ArtOfCode- (approved):

cellio:
Turns out this is not working right on the dev server.  The check for edits is applied on the _local_ site, not the target one, and for non-admins the dashboard is busted.  I guess we need to revert this and figure out how to test in a multi-community environment.

PR #1281, opened 2023-12-28
Fix broken Docker build

Oaphi:
Simple fix for the Docker build failing due to using Ruby 2.7.6 image while the current Gem list requiring Ruby 3.0.0 and higher.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.7.6
+FROM ruby:3.0.0
 
 # docker build -f docker/Dockerfile -t qpixel_uwsgi .
 


cellio:
Should this and https://github.com/codidact/qpixel/pull/1280 be reviewed together?  (They're different changes to the same Docker config file.)

Oaphi:
> Should this and #1280 be reviewed together? (They're different changes to the same Docker config file.)

@cellio yup, they should, I only noticed that someone else had opened a PR for the same issue earlier today after submitting it. Should've checked, the existence of the second one turned out to be quite a surprise, d'oh.

ArtOfCode- (approved):

PR #1282, opened 2023-12-31
Fix has_ability_on checks

ArtOfCode-:
`has_ability_on`:
 * Use `User#is_moderator` method instead of each check individually
 * UserAbility query correctly scoped with regard to Ability

`dashboard.html.erb`/`application_controller.erb`
 * Move SuggestedEdit query to controller to avoid querying for each category individually
 * Scope SuggestedEdit query correctly


cellio (approved):
LGTM. Thank you for the education.

Issue #1279, opened 2023-12-27
On page listing suggested edits awaiting review, add links and counts for other categories


cellio:
Continuation of https://github.com/codidact/qpixel/issues/452, specifically:

> Preferable: when you go to the "edits" page, either it shows all categories or it lists all the categories and shows the silver circle for each one with pending edits. (If we show all categories, we need to include the category name with each suggestion in the list.)

After https://github.com/codidact/qpixel/pull/1278, it seems possible to add the counts and links for pending edits in all categories on each category's pending-edits page.  This way, when you visit one of them you'll see that there are also pending edits elsewhere.  That's easier than unifying the pages, both technically and to help users keep the category context in mind when reviewing.

PR #1286, opened 2024-01-03
on suggested-edits page, link to other category pages

cellio:
Fixes https://github.com/codidact/qpixel/issues/1279.

Increase visibility of pending edits in other categories by cross-referencing the per-category pages.  The new list shows up when you're looking at the "pending" tab (default) but not the "decided" tab; I don't see much value in xrefs in the latter case, particularly when you had to click through the view with the xrefs to get to "decided".

New table below buttons and above the list of edits:

screen shot

The table includes the current category (because the count is useful), but it's not a link because that's this page.

"Decided" view is unchanged:

screen shot

Also adjusted the top-level text since it's not just suggestions for this category any more.

### DIFF
--- a/app/views/suggested_edit/category_index.html.erb
+++ b/app/views/suggested_edit/category_index.html.erb
@@ -1,5 +1,5 @@
 <h1>Suggested Edits</h1>
-<p class="is-lead">This is a list of suggested edits on posts in this category.</p>
+<p class="is-lead">Suggested edits for review.</p>
 
 <% categories = Category.unscoped.where(community: @community).order(sequence: :asc, id: :asc) %>
 <% current_cat = current_category %>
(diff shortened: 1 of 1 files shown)


app/views/suggested_edit/category_index.html.erb
```diff
@@ -15,6 +18,41 @@
...
+      <% if cat == current_cat %>
+        <%= cat.name %>
+      <% else %>
+        <%= link_to suggested_edits_queue_url(cat) do %>
+        <%= cat.name %>
+	<% end %>
+      <% end %>
+      </td>
+      <td>
+      <% if cat == current_cat %>
```
cellio, line 42: I couldn't find a way to avoid the repetition here.  Wrapping the link around the pair of table cells did not work.

ArtOfCode- (approved):

PR #1290, opened 2024-01-07
Add a MathJax button

### LINKED ISSUE #1266: Putting $$ around text when pressed
**Is your feature request related to a problem? Please describe.**
When reformatting posts containing math (but not yet in LaTeX) into LaTeX, I'll need to place a dollar sign on both ends of the math expression. I think this can be simplified and made more accessible.

When a user has some text selected and the `$` key is pressed, it would be nice if the text select would have dollar signs placed before and after it, similar to how the `ctrl+b` shortcut works for `*`.

An alternative would be to add a shortcut (like `ctrl+b` for `*`). As far as I know however, there is no established shortcut for adding `$` so users would have to learn this shortcut.

ArtOfCode-:
Adds a MathJax button to the Markdown editor on communities where MathJax is enabled.

Fixes #1266.

### DIFF
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -30,7 +30,8 @@ $(() => {
       numbered: ['\n 1. ', null],
       heading: ['\n# ', null],
       hr: ['\n\n-----\n\n', null],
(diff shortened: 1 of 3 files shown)


Oaphi (approved):
LGTM. Am I correct that we don't have any way to manage undo in the editor in general yet?

PR #1306, opened 2024-02-11
paginate handled flags

cellio:
I noticed that the (moderator) page of handled flags was getting a little sluggish on our most flag-heavy site.  This one-liner adds pagination.  The pages for pending flags and user flag history were already paginated.  I couldn't find where the list of escalations is produced.  We should check that at some point, but having enough pending escalations to slow down page load also feels like a very rare case.

Proof of pagination (hey, when you need to manufacture 50+ flags to test, the reasons get a little, uh, minimal):

Screenshot

### DIFF
--- a/app/views/flags/handled.html.erb
+++ b/app/views/flags/handled.html.erb
@@ -14,3 +14,5 @@
 <% @flags.each do |flag| %>
   <%= render 'handled', flag: flag %>
 <% end %>
+
+<%= will_paginate @flags, renderer: BootstrapPagination::Rails %>


ArtOfCode- (changes requested):
Is `@flags` paginated in the controller action for this view?

cellio:
> Is `@flags` paginated in the controller action for this view?

Yes, assuming that https://github.com/codidact/qpixel/blob/e26a930960a1af68c8f7e66bc36a092329027319/app/controllers/flags_controller.rb#L57 is the right place to look.  I see pagination for queue, handled, and escalated there.

ArtOfCode-:
Huh, wonder why this wasn't done before if it was provided in the controller 🤷🏻

ArtOfCode- (approved):

PR #1309, opened 2024-02-13
Update css_parser gem and its dependencies to fix 1308

### LINKED ISSUE #1308: Rails error "wrong number of arguments"
**Describe the bug**
Trying to run Rails with either `rails s` or `rails c` gives an error `wrong number of arguments (given 3, expected 1..2)`.

This may be due to the change in [how positional arguments are handled in Ruby 3](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/).

**To Reproduce**
1. Install QPixel following the installation instructions, with Ruby 3.3.0 (I used RVM on Fedora Linux).
2. Try to use `rails c` and see the following error message:
```
/home/trichoplax/.rvm/gems/ruby-3.3.0/gems/css_parser-1.11.0/lib/css_parser/regexps.rb:11:in `initialize': wrong number of arguments (given 3, expected 1..2) (ArgumentError)
```

This appears to be fixable by upgrading the `css_parser` gem. I'm not sure why my previous installations of QPixel with Ruby 3 have worked without this error though.

trichoplax:
Fixes #1308


trichoplax:
I've tested that I can set up a new development instance and create users and posts now. Is there any reason to avoid this upgrade?

trichoplax:
Following ArtOfCode's comment on the issue, I'm converting this pull request to a draft as it will not be necessary until we upgrade Ruby.

In the meantime you can avoid the error by using Ruby 3.1.

ArtOfCode- (approved):

PR #1316, opened 2024-02-24
Rewording to specify that editing follows the goals of the post, not the poster

ws909:
Clarify that it's the goal of the post itself which matters, and not that of its author. The author's goals may change, or they may have created a post which turned out not to be aligned with the goals they had. Once a post has been committed to Codidact, the author's goals cease, and the network's goals with the post take effect.

### DIFF
--- a/db/seeds/abilities/edit_posts.html
+++ b/db/seeds/abilities/edit_posts.html
@@ -6,4 +6,4 @@
 <img src="/assets/ability-help/edit-link.png" alt="Edit post button">
 
 <h2>How do I earn this ability?</h2>
-<p>To earn this ability, you need to have at least a 95% approval rate for your suggested edits, with a hard minimum of 30 approved suggested edits (these numbers may vary from site to site).</p>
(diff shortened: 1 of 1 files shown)


ws909:
Oh, it's been a while since I edited directly on Github, so I managed to mess up the commit titles. Should I fix that, or will they be squashed at merge?

trichoplax:
Is there any existing documentation that supports this assertion? If not, we could have a Meta discussion to see if there is support for changing from the current stance that an author owns their own posts.

ws909:
As far as I’m concerned, that posts are owned by the network, and not the author, is one of the pillars of Codidact, and any quality curated Q/A.

Other than this very document, I have never seen any established stance that authors are fully in control of their posts, and that would also run counter to the goals of the network, the way I understand it. 

Certainly, you’d grant original authors a greater level of autonomy in deciding the fate of their posts in the meta category, but that would be the exception, and not the rule.

ArtOfCode-:
Suggestion: instead of "respect the goal of the post" (which reads odd - a person can have goals, but a post?), consider "preserve the intention of the post".

ws909:
That's much better wording, indeed.

ArtOfCode- (approved):

Issue #1317, opened 2024-02-25
No indication that subscription was created


peterelbert:
**Describe the bug**
The Subscribe button allows you to subscribe to certain kinds of content. After subscribing, there is no immediately obvious on-screen feedback that you are already subsribed.

![image](https://github.com/codidact/qpixel/assets/96020591/67b2b4c3-6f2b-435d-81ee-7f9c34d9a96d)

**To Reproduce**
1. Go to https://meta.codidact.com/categories/3/tags
2. Click on tag "comments" - https://meta.codidact.com/categories/3/tags/952
3. Subscribe to receive updates once a month.

![image](https://github.com/codidact/qpixel/assets/96020591/e2ab4400-432d-4f63-bf13-7e21265b0712)

4. Press submit. There is no visual change in the tag screen when you return to it. You can create the same subscription again, with no message that you already subscribed.

**Expected behavior**
The subscribe button should change visually, with a check mark indicating that you are already subscribed. You should be able to edit the existing subscription if you want.

PR #1318, opened 2024-02-25
if already subscribed to a tag, change the button label and link

cellio:
Fixes https://github.com/codidact/qpixel/issues/1317.

If you're already subscribed to the tag, the button now says "subscribed" and links to the subscriptions page on your profile.

I noticed while working on this that the tooltips are not right -- everything on that row has just the number of posts in the tooltip, despite the various elements having their own more-informative tooltips.  I remembered this problem from the heading on the category page, so I spent far too long banging my head against divs and spans trying to fix this based on that example.  I couldn't crack it and this part was already like this, so I'd like to defer that and meanwhile make the subscription button clearer.


cellio:
Can anyone help me understand the errors?  They look like build errors but I was able to run this locally with no obvious problems.

ArtOfCode- (approved):

PR #1336, opened 2024-04-08
Make it directly usable in Docker

katomaso:
Hi, I am trying to use your app semi-production in Docker and this is necessary changes to be able to run it painlessly on stock Ubuntu server with generic database.

Are any of those changes interesting for you that you would consider porting them to your code?


cellio:
@Taeir I'm not directly assigning you as a reviewer (though qpixel-reviewers should cover you), but pinging you in case you have both specific knowledge here and time to review.  If not, don't worry about it -- this is why we have a reviewers group. :-)

tsengia:
I am very interested in this PR. Being able to deploy via Docker would be great! Let me know if I can be of help somehow.

cellio:
@ArtOfCode-  would you be able to help shepherd this along?  I don't know if you're the right reviewer but you have a better sense of this than I do.  Thanks.

@tsengia thanks for your patience. Note that our Docker setup wasn't intended for production systems and you might run into some issues.  Most notably, Redis data isn't stored outside of Docker, and Redis is where we store some state info (like when you last viewed a category) and also some user preferences.  Qpixel also can't send email from inside Docker.  For more caveats, see https://collab.codidact.org/posts/289900/289902#answer-289902 .  You said "semi-production", so maybe none of these are issues for you, but I just wanted to make sure you know.

I don't know what it would take to change any of these restrictions; if you have ideas on how to make that better, we'd welcome PRs.  But this is where it stands now to the best of my limited knowledge.

cellio:
@katomaso I'm sorry for the delay.  This PR includes both Docker changes and other changes, but it's not clear (to me at least) how the other changes are related.  Can you say more about how the Ruby and db/seeds changes are related to the Docker issue?  Or are they unrelated, and we could merge the Docker changes independently of the others?  Thanks.

docker/Dockerfile
```diff
@@ -17,27 +15,21 @@ RUN apt-get update && \
...
 
 # Add core code to container
 WORKDIR /code
-COPY . /code
+COPY Gemfile Gemfile.lock /code/
 RUN gem install bundler -v 2.4.22 && \
     bundle install
 
+COPY . /code
+RUN bundle exec rails assets:precompile && \
```
Oaphi, line 30: NB!
---
If anyone decides to port this change, please note that we do not have (yet) a separate Dockerfile for development _and_ precompiled assets are not excluded from git tracking via `.gitignore`, meaning that without changing the above, doing so will heavily pollute the working directory.

db/migrate/202309141441_disable_needs_author_attention_flag.rb
Oaphi, line 1: Quick question - can you clarify a little as to what was the reasoning behind removing this migration? I presume there was a compelling reason for doing so, but lack context.

docker/Dockerfile
```diff
@@ -17,27 +15,21 @@ RUN apt-get update && \
...
-RUN wget https://nodejs.org/dist/v12.18.3/node-v12.18.3-linux-x64.tar.xz && \
-    tar xf node-v12.18.3-linux-x64.tar.xz && \
-    wget https://imagemagick.org/archive/binaries/magick && \
-    chmod +x magick && \
-    mv magick /usr/local/bin/magick
-
-ENV NODEJS_HOME=/opt/node-v12.18.3-linux-x64/bin
-ENV PATH=$NODEJS_HOME:$PATH
+        libgdbm-dev \
+        nodejs \
```
Oaphi, line 19: NB!
---
If anyone decides to port this change, please, review that the Node.js version pulled is compatible or, better, yet, fix it to a specific version range (as we should do with all other dependencies, tbh).

Oaphi: Outdated due to merge of #1430 - we no longer install imagemagick and Node.js directly (bullseye has imagemagick out-of-the-box, and Node.js is now copied over from a multi-step build).

docker/Dockerfile
Oaphi: Partially outdated due to #1430 - we now allow custom Dockerfiles to be used so as users can choose what's best for them. Still haven't excluded precompiled assets from tracking, though.

PR #1340, opened 2024-05-12
show profile creation date/time to owner and mods

cellio:
Based on a question in chat, this PR adds the date and time a profile was created, only for the owner and moderators.  (Owner because it's yours and why not; mods because it might help with investigating certain kinds of abuse.)

Some sites show account age to everyone.  If we decide to do this, I would suggest showing the full info to the owner and mods but showing at most only the date to everyone else.  I don't know if there's a privacy reason to fuzz this more, so for now I'm not showing it publicly at all.

I considered adding this to the table of stats (top-level posts, answers, edits, etc), but all of those values are short while timestamps are long, so it flowed weirdly.  I saw no clean way to fit it in there, so I made a separate new mini-section.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -147,6 +147,14 @@
         <td><%= @user.metric 'E' %></td>
       </tr>
     </table>
+    
+    <% if current_user&.id == @user.id || current_user&.is_moderator %>
+    <table class="table is-full-width">
+      <tr>
+	<td>User since <%= @user.created_at %></td>
+      </tr>
+    </table>
+    <% end %>
 
     <% unless @abilities.empty? %>
       <h2>Earned Abilities</h2>


ws909:
I think it would be reasonable to hide the exact account creation date for everybody else, and instead show: 
- Created X days ago (if x <= 60)
- Created Y months ago (if y <= 12)
- Created Z years and _y months ago 

This provides a little bit of privacy such that users cannot get the exact date after the user has been a member for a while (unless their profile page was archived within 60 days), and it still shows others approximately for how long somebody has been a member (which can be useful in various situations).

trichoplax:
> Well, I guess people can also monitor user accounts for when the month number changes, but, I’d say that’s not too bad.

Because of this possibility, the vague version of creation date does not add real privacy, and worse, it gives the false impression of privacy. I would prefer that the privacy level be clear and explicit - either showing the full date to everyone or showing nothing to anyone other than the user and moderators.

We could add randomness to try to prevent getting a specific date out of the vague version, but that seems to invite the possibility of privacy bugs, so I'd prefer to avoid that and stick to something simple.

I like the approach implemented here (only the user and moderators see the creation date). It might be nice to have a checkbox that allows people to make their account creation date public if they wish, but that can be a future feature if someone chooses to raise it. I see no reason to implement it in this pull request.

app/views/users/show.html.erb
```diff
@@ -147,6 +147,14 @@
         <td><%= @user.metric 'E' %></td>
       </tr>
     </table>
+    
+    <% if current_user&.id == @user.id || current_user&.is_moderator %>
+    <table class="table is-full-width">
+      <tr>
+	<td>User since <%= @user.created_at %></td>
```
trichoplax, line 154: I can imagine 2 potential improvements to the display format, but since this is already fully working I recommend merging as is, and raising these 2 improvements as separate issues if desired. If you like either of them I'm happy to raise them myself - just let me know.

1. Make the final space between the time and the timezone a [non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space), so when width is limited it displays as:
    ```text
    User since 2024-05-11
    12:51:22 UTC
    ```
    rather than:
    ```text
    User since 2024-05-11 12:51:22
    UTC
    ```
2. Show a less accurate but easier to read time elapsed, for consistency with other datetimes, with the full datetime shown on click or tap. This could use the `time_ago_in_words` function to show:
    ```text
    User for 16 days
    ```
    changing when clicked to:
    ```text
    User since 2024-05-11 12:51:22 UTC
    ```

trichoplax (approved):
Approved with potential improvements that I recommend be left for future issues as this is self-contained.

app/views/users/show.html.erb
cellio: Both of these are definite improvements -- please do raise them, thanks!  (How do we do the first part -- grab the string, work back from the end to the first space, and then replace it and return the new string?  Or do we have manipulation functions for times/dates/timezones that we can use to extract the pieces and build the string we want?)

trichoplax: I've now raised #1344 and #1345

As for how to approach the non-breaking space, your first suggestion is definitely possible (and is the only way I currently know how to do it). Ideally we would use your second suggestion, which would be more robust in case of other QPixel instances choosing to use different datetime formats that might be in a different order or might use different separators than spaces. I imagine Ruby has such built in functions (many languages do) but I'm not familiar with them yet.

PR #1346, opened 2024-05-31
adds post count to 'see all' button on user profile

cellio:
Addresses https://meta.codidact.com/posts/291590 (minimally).  I couldn't find where in the code the threshold of 15 posts to show on the main profile is set, so I didn't add "showing N of M" 'cause I don't know how to get N and don't want to hard-code.


trichoplax (approved):
Looks good.

Correctly works with deleted posts where the total number differs for users and moderators. In either case the total number shown on the button matches the total number shown on the resulting Posts page.

cellio:
This worked in branch testing, but after deploying I now see "see all 15" instead of "see all (correct number)" for users with more than 15 posts.  So there's still a bug here.

Issue #1354, opened 2024-06-26
Add a flag for curators to change post state (reopen, lock, delete)


cellio:
meta:291860, specifically [this comment thread](https://meta.codidact.com/comments/thread/9737).

Currently there is no way, outside of a custom flag, to indicate that you think a closed question should be reopened.  This means that only moderators can handle such requests, and only if someone thought to write a custom flag about it.

We'd like the community to have a stronger role in closing and reopening, especially while those remain one-vote actions.  Users with the Curate ability can close/reopen, and they can also see standard flags -- but not the custom moderator-only flags.  If we had a flag on top-level posts, then curators would see the flag and could act on it.

This isn't just about reopening closed posts (the specific case raised in the linked Meta thread).  Curators can do other things too: they can lock or unlock posts, and they can delete or undelete them.  Therefore, my proposal is:

1. Rename "other" (the mod-only one) to "moderator attention" and add some text that only moderators will see this flag.
2. Add "change post status" (or something similar) with text like "post should be reopened/closed, locked/unlocked, or deleted/undeleted. Please elaborate..." (last part as for "other").

PR #1355, opened 2024-06-27
Added flag type for changing post status (like reopen, but also lock, delete, and inverses)

cellio:
Fixes https://github.com/codidact/qpixel/issues/1354.

Added 'change post status' flag, changed 'other' to 'moderator attention' and added that it's mod-only to the description (not curators). Also changed the button on the flag modal from 'flag for moderator attention' to just 'flag for attention' because most flags are for curator attention too.

New flag reason:

change post status: This post should be reopened/closed, locked/unlocked, or deleted/undeleted. Please elaborate in the details field below.

New version of "other":

moderator attention: Please elaborate in the details field below. This flag will be seen only by moderators and administrators, not curators.

The "not curators" part in that description is intentional, as signal to people who might not realize that their other flags have a wider audience.  That's something that Codidact does differently from other sites, so a little reinforcement seems helpful.


app/views/posts/_expanded.html.erb
```diff
@@ -371,9 +371,9 @@
                 </div>
                 <div class="grid--cell is-flexible">
                   <label class="form-element has-margin-0" for="flag-reason-other_<%= post.id  %>">
-                    other reason
+                    moderator attention <br/>
```
trichoplax, line 374: I like the simplicity of just "moderator attention" rather than padding it out with extra words like "it needs moderator attention".

Would this be a good opportunity to make the other options consistent with this? Currently the first two start with "it's", the third starts with "is", then the rest don't have any extra words. Consistency is my main aim, but my personal preference would be the shortest form, with no "it's" or "is".

I'd be happy to approve this pull request with either consistency applied here, or a separate issue raised for applying the consistency later.

db/seeds/post_flag_types.yml
```diff
@@ -26,9 +26,17 @@
   active: true
   post_type_id: <%= Question.post_type_id %>
 
+- name: change post status
+  description: >
+    This post should be reopened/closed, locked/unlocked, or deleted/undeleted. Please elaborate in the details field below.
```
trichoplax, line 31: This wording is showing up correctly in my development environment, so I can select this option when flagging, but unlike the "moderator attention" flag, this one does not insist on the details text field being filled in.

I can see 2 potential solutions:
1. Separate the options sufficiently that each one requires no further details. I guess 3 options would suffice:
    - Open/close
    - Lock/unlock
    - Delete/undelete
2. Keep just one option for all 6 actions, and add an extra property for custom flag types. In addition to "confidential", "active", and "post_type_id", there could also be "requires_details". This would also require adding code elsewhere to make this work.

Option 1 is easy to implement if you consider it sufficient for now. Option 2 seems unnecessary unless you think there will be other cases that need it in future. It doesn't seem worth it just for this particular case, as I'm guessing this flag will be a temporary measure until a more automated approach is implemented (for example, a reopen button available to everyone, that acts similarly to the edit button in that it either takes the action if you have authority, or adds it to a queue if not).

app/views/posts/_expanded.html.erb
```diff
@@ -315,7 +315,7 @@
 
         <div class="post--action-dialog js-flag-box">
           <div class="widget">
-            <div class="widget--header">Why does this post require moderator attention?</div>
+            <div class="widget--header">Why does this post require attention from curators or moderators?</div>
```
trichoplax, line 318: I like the clarity here that both curators and moderators can see some flags.

Note that the pop up message still says "Thanks! A moderator will review your flag" regardless of whether it's a moderator only flag or a curator also flag.

trichoplax (changes requested):
Looking good. I've added some comments with a few suggestions.

app/views/posts/_expanded.html.erb
cellio: I agree we should make it consistent.  My only concern is the issue I hit in https://collab.codidact.org/posts/291874 -- when we push those changes to the network, are we going to get *two* versions of each of those flags?  The only reason it didn't happen here with other -> moderator attention is that that case is baked into the code, not in the YAML file that gets seeded.

I'd like to make the flag text consistent, if we can do it without making a mess.  I don't know how to do that yet. :-(

db/seeds/post_flag_types.yml
cellio: Option 3: keep the one flag option, figure out how the code is enforcing the details text for mod flags, and do that.  I see that some flag names are hard-coded, which is something we might need to address in the future, but we could in the meantime do the same thing for this flag type.

app/views/posts/_expanded.html.erb
trichoplax: I don't see any reason for this to hold up this pull request. It's not making anything worse. If an issue can be raised then we can work it out when we get to it, and just focus on what's essential for this pull request.

db/seeds/post_flag_types.yml
trichoplax: If option 3 turns out to be easier then I'm fine with that. I'm not worried about hard-coding something that will only be temporary.

*(If anyone reading this is cringing, suspecting that "temporary" may last forever, feel free to raise an issue so it doesn't get forgotten, but I see no reason to fix it in this pull request.)*

app/views/posts/_expanded.html.erb
trichoplax: Reading this again, I hope it doesn't sound like a request for a context reactive message. This message that pops up is just an acknowledgement to confirm the flag was raised successfully, so it doesn't need to be specific about who will process it. I expect something along the lines of "Thanks! Someone will review your flag" or even "Thanks! We'll review your flag" would be fine to cover both cases.

cellio: Don't worry; I didn't take it that way, and already changed it to "we".

db/seeds/post_flag_types.yml
cellio: I decided to hard-code this new curator-attention flag, like we do for the moderator-attention flag.  This code enforces the requirement for the details field, which can't be done through the YAML file currently.  I've removed the reason from the YAML file (and thus we don't need to worry about problems in seeding).

cellio:
> Looking good. I've added some comments with a few suggestions.

Updates:

- Moved the new flag from the YAML file to be hard-coded (like "other" aka "moderator attention") already was).  The YAML file is still showing up in the diff; need to figure out how to remove it from this changeset or unchange whatever it thinks changed (whitespace?).
- Now that the flag reason is in the code, it enforces the requirement for details.
- Generalized the confirmation message on submission since it's not just moderators.

app/views/posts/_expanded.html.erb
cellio: Let's file a separate issue for the wording inconsistencies.

db/seeds/post_flag_types.yml
```diff
@@ -32,3 +32,4 @@
   confidential: false
   active: true
   post_type_id: null
+
```
trichoplax, line 35: This is just a newline at the end of the file. If you delete the newline then this file will disappear from the diff.

trichoplax (changes requested):
- I've added a comment about the YAML file.
- The details are now required in my development environment - looks good.
- I like the new pop up message.

Since the new flag reason was added to my local database, I'm now seeing the status change flag twice, but this should be fine in production where the new flag was never added to the YAML file.

cellio:
> Since the new flag reason was added to my local database, I'm now seeing the status change flag twice, but this should be fine in production where the new flag was never added to the YAML file.

Same -- I'm seeing the extra one in my dev environment because I previously seeded it and haven't figured out how to back out just that change, but it shouldn't affect prod.  We can confirm that on the dev server after merge and before deploy.

db/seeds/post_flag_types.yml
cellio: Gone now, thanks.

app/views/posts/_expanded.html.erb
cellio: Do you want to file an issue for that?  Can we resolve this conversation and approve?

trichoplax: Added #1359.

trichoplax (approved):

PR #1364, opened 2024-07-24
profile remove category column

cellio:
Mitigates https://meta.codidact.com/posts/291869 as discussed in [comments](https://meta.codidact.com/comments/thread/9741).

The user profile page has 8 tabs (for your own profile, when logged in).  Some of those sub-pages omit the right column but others don't, and the addition of the right column causes some unfortunate wrapping in both the tabs row and some page content.  Except on the main profile page (which is not changed by this PR), the right column is the "category" column that you see on most pages on the site.  The content there -- bulletins, hot posts, subscription links, etc -- makes sense in the context of a list of posts or even an individual post, but it's less relevant when looking at your vote summary.

This PR removes that column from most of the profile sub-pages that had it, specifically: votes summary,  preferences, filters, notifications.  Activity and edit already omitted it, and the main profile page is special.  That leaves the account sub-page (where you change your password etc), for which I couldn't find the relevant code.  Since all of this is mitigation for wrapping that will still happen on smaller screens, I figure an incremental fix is better than nothing at all.

Before: see meta post.

After: 

Screenshot, single tabs line without wrapping

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -67,9 +67,10 @@ class UsersController < ApplicationController
         @community_prefs = prefs[:community]
       end
       format.json do
-        render json: current_user.preferences
+        render json: current_user.preferences 
       end
     end
+    render layout: 'without_sidebar'
   end
 
   # Helper method to convert it to the form expected by the client
@@ -108,6 +109,7 @@ class UsersController < ApplicationController
         render json: filters_json
       end
     end
+    render layout: 'without_sidebar'
   end
 
   def set_filter
@@ -581,6 +583,7 @@ class UsersController < ApplicationController
                    end \
                    .paginate(page: params[:page], per_page: 15)
     @votes
+    render layout: 'without_sidebar'
   end
 
   def avatar
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -6,7 +6,7 @@ class NotificationsController < ApplicationController
     @notifications = Notification.unscoped.where(user: current_user).paginate(page: params[:page], per_page: 100)
                                  .order(Arel.sql('is_read ASC, created_at DESC'))
     respond_to do |format|
-      format.html { render :index }
+      format.html { render :index, layout: 'without_sidebar' }
       format.json { render json: @notifications, methods: :community_name }
     end
   end


cellio:
Maybe instead of this (which is also causing some test failures), we should attack the original problem of the list of tabs being too wide (and thus wrapping) by constraining it to one line with a "..." at the end to get to whatever didn't fit.  That would presumably work on phones too, but could be pretty annoying and less discoverable.  I don't know what the tradeoffs are.

app/controllers/users_controller.rb
```diff
@@ -70,6 +70,7 @@ def preferences
         render json: current_user.preferences
       end
     end
+    render layout: 'without_sidebar'
```
trichoplax, line 73: This method has a section for what to do if HTML is requested, and a section for what to do if JSON is requested. The method is only allowed to have 1 render call. Putting a render call inside the HTML section will work fine, but putting it here outside both sections means that when JSON is requested (which already has its own render call) this will call render twice.

Moving this line inside the `format.html` block should remove the errors that are making the tests fail. It needs to be the last line in that block so that the variables being defined there are available to it.

app/controllers/users_controller.rb
```diff
@@ -108,6 +109,7 @@ def filters
         render json: filters_json
       end
     end
+    render layout: 'without_sidebar'
```
trichoplax, line 112: This render call has the same problem as the preferences one: there is already a render call in the JSON section so this would be a second render call, which is not allowed. The same solution applies - this line just needs to be inside the `format.html` section, after the `authenticate_user`.

app/controllers/users_controller.rb
```diff
@@ -581,6 +583,7 @@ def vote_summary
                    end \
                    .paginate(page: params[:page], per_page: 15)
     @votes
+    render layout: 'without_sidebar'
```
trichoplax, line 586: This one looks good.

trichoplax (changes requested):
A couple of the methods have two sections, one for HTML and one for JSON. People viewing the page in their browser (the only context where a sidebar makes sense) will be seeing the HTML version, so that's the section that will need the suppression of the sidebar. I've added comments to the 2 places that are affected by this.

cellio:
> A couple of the methods have two sections, one for HTML and one for JSON. People viewing the page in their browser (the only context where a sidebar makes sense) will be seeing the HTML version, so that's the section that will need the suppression of the sidebar. I've added comments to the 2 places that are affected by this.

Oh!  I see now.  Thank you for the clues.  I was puzzled by the complaint about multiple calls because JSON seemed like a different path, except duh, it wasn't.

trichoplax (approved):
All working now. The only slight difference is there is now a JavaScript error in the developer console for the tabs that have had the sidebar removed. However, I don't see this as a reason to delay merging this pull request as that error is already present on the `develop` branch for all of the tabs that already have no sidebar. I see it as a pre-existing problem.

I've raised #1366 so someone can investigate that existing error separately.

PR #1365, opened 2024-07-24
fixed typo in abilities help seed (+ clarification)

cellio:
Fixed typo, and while in here anyway, added a sentence about new-site mode to avert confusion.  (We've seen people be confused about this, and also, a two-character change didn't seem like enough justification for a PR.)

Fixes https://meta.codidact.com/posts/290825.


trichoplax (approved):
Looks good. The only thing I wonder is where to use "site" and where to use "community" but I guess that would affect more places so best left for another pull request, so no changes needed here.

PR #1367, opened 2024-07-24
revert: profile remove category column

cellio:
https://github.com/codidact/qpixel/pull/1364 introduced a server error if you are not logged in and access a login-restricted tab directly (e.g. /users/me/filters).  The correct (and previous) behavior is to prompt for login.  This PR removes those changes.


Oaphi (approved):
As per discussion - LGTM, reverting for now.

PR #1368, opened 2024-07-24
Fix for pull request 1364

trichoplax:
When reviewing pull request 1364 I overlooked that the filters tab controller method would error when not signed in due to containing both `authenticate_user!` and a `render` call. The changes from that pull request were therefore reverted. Changing from an explicit `authenticate_user!` to a `before_action` fixes this.


Oaphi (approved):
LGTM as per discussion

PR #1384, opened 2024-08-23
don't try to update rep for the nil user

cellio:
A moderator tried to delete an answer by a deleted user and got a 500 error, specifically:

```
    NoMethodError: undefined method `reputation' for nil:NilClass
        user.update(reputation: user.reputation - Vote.total_rep_change(votes))
                                    ^^^^^^^^^^^

/var/apps/qpixel/app/models/post.rb:264:in `modify_author_reputation'
/var/apps/qpixel/app/controllers/posts_controller.rb:402:in `delete'
```

I added a guard against trying to change the rep for a deleted user.  This might not be the best way to do it.  Also, I was not able to reproduce the bug in a local dev environment, so it's possible this isn't really the cause of the error.  But server errors are bad, so let's try to fix this one.


app/models/post.rb
```diff
@@ -260,10 +260,12 @@ def modify_author_reputation
...
-        user.update(reputation: user.reputation - Vote.total_rep_change(votes))
-      else
-        user.update(reputation: user.reputation + Vote.total_rep_change(votes))
+      if user
+        if deleted
+          user.update(reputation: user.reputation - Vote.total_rep_change(votes))
+        else
+          user.update(reputation: user.reputation + Vote.total_rep_change(votes))
+        end
       end
```
Oaphi, line 269: Let's go for a cleaner option and use optional chaining (aka "safe navigation" in Ruby) instead to avoid the "hadouken" problem:

```ruby
if deleted
    user&.update(reputation: user.reputation - Vote.total_rep_change(votes))
else
    user&.update(reputation: user.reputation + Vote.total_rep_change(votes))
end 
```

Oaphi (changes requested):

app/models/post.rb
ArtOfCode-: The `user` *should *always exist - deleted users aren't deleted, they just have a `deleted` flag set on them. Clearly that's failing in some cases, though. I'd add a SNO for the reputation as well. My suggestion would be:

```suggestion
      if deleted
        user&.update(reputation: (user&.reputation || 1) - Vote.total_rep_change(votes))
      else
        user&.update(reputation: (user&.reputation || 1) + Vote.total_rep_change(votes))
      end
```

cellio: Oh, I wonder if it was a destroyed user.  I didn't try that test case yet.  (Would that make a difference?)

cellio: Ok, that's god an error, though I got a 418, not a 500.

Repro: post something from a scratch account, log in as admin, *destroy* (not delete) the scratch account, then delete the post.

cellio:
@Oaphi  I applied @ArtOfCode- 's change. The key seems to have been destroyed (rather than deleted) accounts.  After I 418ed myself and figured out how to do the db surgery to get back to a working dev environment, I think this works now.

Oaphi (approved):
LGTM!

Issue #1353, opened 2024-06-14
Tag search returns 500 under some circumstances


cellio:
meta:291242 (and [another reproducer](https://software.codidact.com/categories/38/tags?q=hax)
Also https://electrical.codidact.com/posts/291745

Not fully characterized.

Searching for a tag that doesn't exist (the EE report) or for a substring of a tag that does exist can produce 500 errors from the server.  We made a fix (that I can't find now) after the initial meta report but it did not fix the whole problem.  This needs more investigation.

MoshiKoi:
Another case: https://meta.codidact.com/posts/292574

cellio:
I triggered this in a dev environment and got the following complaint.  The console highlighted the call to `will_paginate` as where the problem occurred.  This is the same pagination function used elsewhere in the code (only the argument is different).  Also note that we are within the scope of an "unless nil" block:

```
uwsgi-1  | ActionView::Template::Error (undefined method `total_pages' for #<ActiveRecord::AssociationRelation []>):
uwsgi-1  |     53:   <% end %>
uwsgi-1  |     54: <% end %>
uwsgi-1  |     55: 
uwsgi-1  |     56: <%= will_paginate @tags, renderer: BootstrapPagination::Rails %>
uwsgi-1  |     57: 
uwsgi-1  |     58: <% end %> <%# unless @tags == nil %>
uwsgi-1  |     59: 
uwsgi-1  |   
uwsgi-1  | app/views/tags/category.html.erb:56
uwsgi-1  | app/models/application_record.rb:53:in `block in with_lax_group_rules'
uwsgi-1  | app/models/application_record.rb:49:in `with_lax_group_rules'
uwsgi-1  | app/views/tags/category.html.erb:5
```

PR #1397, opened 2024-09-08
Expanding post draft logic to handle fields other than the body

### LINKED ISSUE #1395: When saving drafts for posts, we should also save the title and tags
meta:292446

When saving drafts, we only consider the post body. It's, indeed, a problem that users can lose some (although not much) work over an issue that causes them to reload the page.

Oaphi:
closes #1395 

In addition to the post's body, we now save (all fields are optional, so we only save them if the context allows for it):

- post title;
- post tags;
- post license;
- edit comment ;
- tag wiki;
- tag excerpt;
- tag name;

Those fields are stored individually, I opted to do so to not invalidate drafts created prior to the change, but open to switching to a hash value instead.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -614,6 +614,12 @@ class PostsController < ApplicationController
       RequestContext.redis.expire license_key, expiration_time
     end
 
+    if params.key?(:tag_name)
+      tag_name_key = "saved_post.#{current_user.id}.#{params[:path]}.tag_name"
+      RequestContext.redis.set tag_name_key, params[:tag_name]
+      RequestContext.redis.expire tag_name_key, expiration_time
+    end
+
     if params.key?(:tags)
       tags_key = "saved_post.#{current_user.id}.#{params[:path]}.tags"
       RequestContext.redis.sadd tags_key, params[:tags]
@@ -698,6 +704,7 @@ class PostsController < ApplicationController
     excerpt_key = "saved_post.#{current_user.id}.#{path}.excerpt"
     license_key = "saved_post.#{current_user.id}.#{path}.license"
     tags_key = "saved_post.#{current_user.id}.#{path}.tags"
+    tag_name_key = "saved_post.#{current_user.id}.#{path}.tag_name"
     title_key = "saved_post.#{current_user.id}.#{path}.title"
     saved_at = "saved_post_at.#{current_user.id}.#{path}"
 
@@ -707,6 +714,7 @@ class PostsController < ApplicationController
       excerpt_key,
       license_key,
       tags_key,
+      tag_name_key,
       title_key,
       saved_at
     ]
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -584,14 +584,43 @@ class PostsController < ApplicationController
     render json: { status: 'success', success: true }
   end
 
+  # saving by-field is kept for backwards compatibility with old drafts
   def save_draft
-    key = "saved_post.#{current_user.id}.#{params[:path]}"
+    expiration_time = 86_400 * 7
+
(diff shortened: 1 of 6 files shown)


ws909:
I can't see a mention of tag synonyms, tag name ("post title" covers that, maybe?), and parent tag. Are these three included?

Oaphi:
> I can't see a mention of tag synonyms, tag name ("post title" covers that, maybe?), and parent tag. Are these three included?

Parent tag and synonyms are a no-go for now, I tried, but the scope of the change would be quite significant, so probably some time later. As for the name, yeah, seems to be worth adding.

ws909:
In that case, this might be related:
[500, internal server error, as I try to create a tag](https://meta.codidact.com/posts/292502)

Oaphi:
> In that case, this might be related: [500, internal server error, as I try to create a tag](https://meta.codidact.com/posts/292502)

I don't think this one's in scope of the change either, there is an issue open for that (https://github.com/codidact/qpixel/issues/1353). I may be the one fixing it, or maybe not, depending on what my availability is. Doesn't seem like something complex.

ws909:
I meant that it may be related to the way that synonyms and parent tag are implemented, not that they're in scope here, after you ruled those two out. But it seems you've located the issue, then, and that the problem is that the system tries searching for the synonyms before creating the tag.

Oaphi:
@ws909 ah, I thought you wanted me to make the change in this PR, sorry. If I get some time to spare, will look into it, just not as part of this change to avoid scope creep.

app/controllers/posts_controller.rb
```diff
@@ -584,14 +584,55 @@ def feature
...
+
     saved_at = "saved_post_at.#{current_user.id}.#{params[:path]}"
-    RequestContext.redis.set key, params[:post]
     RequestContext.redis.set saved_at, DateTime.now.iso8601
-    RequestContext.redis.expire key, 86_400 * 7
-    RequestContext.redis.expire saved_at, 86_400 * 7
-    render json: { status: 'success', success: true, key: key }
+    RequestContext.redis.expire saved_at, expiration_time
+
+    if params.key?(:comment)
```
ArtOfCode-, line 599: Feels like this set of `if`s could be made into a loop over an array of the field names. Something like:

```ruby
[:comment, :excerpt, :license, :tag_name, :tags, :title].each do |key|
  if params.key? key
    key_name = "saved_post.#{current_user.id}.#{params[:path]}.#{key.to_s}
    RequestContext.redis.set key_name, params[key]
    RequestContext.redis.expire key_name, expiration_time
  end
end
```

app/controllers/posts_controller.rb
```diff
@@ -658,9 +699,27 @@ def unless_locked
...
+    keys = [
+      body_key,
+      comment_key,
+      excerpt_key,
+      license_key,
+      tags_key,
+      tag_name_key,
+      title_key,
+      saved_at
+    ]
```
ArtOfCode-, line 720: Also feels like this could be a map over just the part names - at least for the specific parts, with the body key added after.

```ruby
[:comment, :excerpt, :license].map { |key| "saved_post.#{current_user.id}.#{path}.#{key.to_s}" }
```

(only some fields included, others left as an exercise for the reader :) )

ArtOfCode- (changes requested):

app/controllers/posts_controller.rb
Oaphi: Good idea - I was adding ghem one-by-one, so never noticed they ended up matching 1:1, will do!

app/controllers/posts_controller.rb
Oaphi: Yeah, agreed

cellio:
@ArtOfCode- did the updates here address your concerns?

ArtOfCode-:
LGTM

ArtOfCode- (approved):

PR #1399, opened 2024-09-10
make tag rss link more visible

cellio:
The tag page already has an RSS link at the bottom, but that's not very obvious (see https://software.codidact.com/posts/291434).  This PR adds the RSS link to the top of the list, parallel to the top of the category post list.  The category post list repeats the RSS link at the bottom, so I've left that in place on the tag page too.

Also added/improved tooltips, among things clarifying that "subscribe" means email.

### DIFF
--- a/app/views/tags/show.html.erb
+++ b/app/views/tags/show.html.erb
@@ -32,7 +32,7 @@
 
 <p class="has-color-tertiary-900 has-font-weight-normal has-margin-0 has-font-family-brand">
   <% if @tag.parent_id.present? %>
-      Subtag of <%= link_to @tag.parent.name, tag_path(id: @category.id, tag_id: @tag.parent_id),
+    Subtag of <%= link_to @tag.parent.name, tag_path(id: @category.id, tag_id: @tag.parent_id),
                             class: tag_classes(@tag.parent, @category) %>
   <% end %>
   <% child_count = @tag.children.count %>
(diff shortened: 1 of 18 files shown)


app/views/tags/show.html.erb
```diff
@@ -92,17 +92,29 @@
...
-    <%= short_number_to_human post_count, precision: 1, significant: false %>
-    <%= 'post'.pluralize(post_count) %>
+    <span title="<%= post_count %> posts with this tag">
+      <%= short_number_to_human post_count, precision: 1, significant: false %>
+      <%= 'post'.pluralize(post_count) %> &nbsp;
+    </span>
+
+    <span title="RSS feed for this tag">
+      <%= link_to tag_path(id: @category.id, tag_id: @tag.id, format: 'rss') do %>
+      <i class="fas fa-rss"></i> RSS &nbsp;
```
Oaphi, line 102: Whitespace should be moved out of the link, otherwise we end up with this on hover:

image

Oaphi (changes requested):

app/views/tags/show.html.erb
cellio: Fixed.

Oaphi:
I've added a followup PR that expands on your changes (a couple of additional fixes not related to your changes & a partial for RSS links) - if you'd like, we could merge them in together: https://github.com/codidact/qpixel/pull/1413

Oaphi (approved):
LGTM

Issue #1349, opened 2024-06-04
remove the list of communities in the footer


cellio:
meta:290804

The list of communities in the footer is a single column and pretty long.  We could adjust the layout, but that list already [isn't entirely accurate](https://github.com/codidact/qpixel/issues/1216), and also, we have the community switcher at the top.  Further, the left column in the footer already links to the list of communities.

Let's remove the individual communities from the footer as redundant.

Stretch goals that could be considered in other issues if they're not easy:

1. On desktop, lay out the remaining links to use the space better (multiple shorter columns).
2. Make the target of the "other communities" link be configurable at the network level instead of hardcoding codidact.com.  (Maybe that is already the case?  I'm not sure where to look.)

When we do this, close https://github.com/codidact/qpixel/issues/1216 as overcome by events.

Issue #1216, opened 2023-10-05
Community list in page footer does not match Communities drop-down in header


cellio:
meta:288624

The newest community on our network, Proposals, is missing from the list of communities shown in the page footer, but is present in the communities drop-down widget in the header.

The footer HTML:

```html
      <div class="grid--cell is-6 is-12-md is-12-sm">
        <h3>Other Codidact Communities</h3>
        <ul>
          <% Rails.cache.persistent('codidact_sites').each do |site| %>
            <li><%= link_to site['name'], site['canonical_url'] %></li>
          <% end %>
        </ul>
      </div>
```
I understand why we would want to cache that; otherwise every single page load would be affected.  Is the cache stuck?  I assume it could be manually flushed, but is there something we could do in our code to avoid the need for manual intervention?

(Aside: if someone's editing that anyway, can we change "Codidact" to "Network" in the heading, in keeping with other attempts to generalize strings in the code?)

The drop-down code, for reference, seems to compute the list dynamically.  I assume that would be a terrible idea in the footer:

```html
  <h3 class="h-m-t-1">Communities</h3>
  <div class="community-header-list">
    <% logo_paths = SiteSetting.all_communities('SiteLogoPath') %>
    <% Community.where(hidden: false).each do |c| %>
      <div class="community-header-list-entry">
        <% logo_setting = logo_paths[c.id] %>
        <%= link_to "//#{c.host}" do %>
          <% if logo_setting.present? %>
            <div class="img">
              <img src="<%= logo_setting %>" alt="<%= c.name %>" class="community-header-list-image">
            </div>
          <% end %>
          <div class="label">
            <%= c.name %>
          </div>
        <% end %>
      </div>
    <% end %>
  </div>
```

cellio:
Community list has been removed from the footer.

PR #1402, opened 2024-09-10
generalized footer

cellio:
Sorted the footer links into "this network" and "Codidact project".  Replaced the generated list of communities (which was getting long and was a little buggy) with a dashboard link.  Added some wayfinding to the dashboard page, including pointer to the communities switcher.

This change removes the baked-in `codidact.com` link that is incorrect for other networks.  I added the GitHub link under "Codidact project" to make the OSS platform a little more visible on other networks (maybe they want to contribute too).

Closes https://github.com/codidact/qpixel/issues/1349 and (indirectly) https://github.com/codidact/qpixel/issues/1216.

### DIFF
--- a/app/views/application/dashboard.html.erb
+++ b/app/views/application/dashboard.html.erb
@@ -1,7 +1,7 @@
 <%= content_for :title, 'Dashboard' %>
 
 <h1>Dashboard</h1>
-<p>All communities, with categories, on this network.  You can also see a list of all communities from the dropdown arrow in the top right corner of this page.</p>
+<p>All communities, with their categories. You can also see the communities from the dropdown arrow at the top right of each page.</p>
 <div class="grid community-list">
   <% @communities.each do |c| %>
     <% categories = Category.unscoped.where(community: c).order(sequence: :asc, id: :asc) %>
--- a/app/views/application/dashboard.html.erb
+++ b/app/views/application/dashboard.html.erb
@@ -1,6 +1,7 @@
 <%= content_for :title, 'Dashboard' %>
 
 <h1>Dashboard</h1>
+<p>All communities, with categories, on this network.  You can also see a list of all communities from the dropdown arrow in the top right corner of this page.</p>
 <div class="grid community-list">
   <% @communities.each do |c| %>
     <% categories = Category.unscoped.where(community: c).order(sequence: :asc, id: :asc) %>
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -2,26 +2,21 @@
   <div class="footer--container container">
     <div class="grid">
       <div class="grid--cell is-6 is-12-md is-12-sm">
-	<h3>This Network</h3>
 	<ul>
-          <li><%= link_to 'All communities', '/dashboard' %></li>
-          <li><%= link_to 'Terms of Service', '/policy/tos' %></li>
-          <li><%= link_to 'Privacy Policy', '/policy/privacy-policy' %></li>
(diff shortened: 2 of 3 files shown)


Oaphi (approved):
LGTM. The only change I'd make, though, is changing "This Network" to something else. If it confuses _me_, pretty sure it will be even more confusing to outside users. Non-blocking, though (and I honestly have no idea how to word it better).

cellio:
> LGTM. The only change I'd make, though, is changing "This Network" to something else. If it confuses _me_, pretty sure it will be even more confusing to outside users. Non-blocking, though (and I honestly have no idea how to word it better).

I'm certainly open to changing it -- I'm just not sure to what. "Communities on this server" doesn't seem much clearer, and I don't think "instance" is widely understood.

trichoplax:
> > LGTM. The only change I'd make, though, is changing "This Network" to something else. If it confuses _me_, pretty sure it will be even more confusing to outside users. Non-blocking, though (and I honestly have no idea how to word it better).
> 
> I'm certainly open to changing it -- I'm just not sure to what. "Communities on this server" doesn't seem much clearer, and I don't think "instance" is widely understood.

Short answer:  
Go with "the network" or (my preference) leave it out altogether.

Long answer:  
Websites are understood to be self contained. A supermarket website listing "All products" doesn't need to mention "(by the way, this is only all products for our company - products from other companies are not listed here)". Users of any particular instance of QPixel don't need to know what an instance is, or that there are other networks, any more than they need to know what programming language we use.

In the paragraph on the dashboard, I would probably say something like:

> All communities, with their categories. You can also see the communities from the dropdown arrow at the top right of each page.

(Less to read and less to be confused about.)

In the footer, I would expect the separation into two sections to have no meaning for many users, and to be a source of confusion. Unless I already know about Codidact, the heading "The Codidact Project" is unlikely to change my expectation that the link "contact us" will be to contact the owners of the website. For every instance except one, this will be incorrect.

My suggestion would be to remove the two section headings, and only include links that relate to the instance (ideally in a way that is configurable without needing a code change). "About us" and "Contact us" should both relate to the instance. Perhaps in future work they could become relative URLs like the others.

There can then be a single link to codidact.org with text that distinguishes it from all the other links, which are instance-specific. Perhaps something like "Powered by Codidact" (please someone suggest something better). This could possibly be separate from the bullet point list of links (I can imagine it in the version label below the footer:

> Powered by [Codidact](https://codidact.org/). Version [604d40c6](https://github.com/codidact/qpixel/commit/604d40c6c03ce445032f58cec1f67e4c445d0351) (2024-09-11 04:12:24 +0100)

If I want to read about or contact Codidact, I can follow that link and understand that any links on that new page relate to Codidact, which powers the website I just left. I think this will be clearer to users than trying to have both a "Contact Codidact" and "Contact Acme" (for the instance) in the same footer.

Parts of this that sound reasonable might be best in other issues - I'm not suggesting we do everything in one pull request.

cellio:
@trichoplax I've made some adjustments, thanks, including moving support/contact info into the network FAQ where other instances can edit it.  On-site config would be even better...someday.

Oaphi:
Minor nit for the network FAQ: specifying "software" twice in a row is unnecessary, and a more appropriate term would be "engine", imo:

> This community is built with the [open-source QPixel engine](https://github.com/codidact/qpixel) provided by [The Codidact Foundation](https://codidact.org/). The software is free, open-source, and highly customizable.

Oaphi:
@cellio to follow up on my note about the timestamp format, here is a quick fixup: https://github.com/codidact/qpixel/pull/1402

Issue #1356, opened 2024-06-28
Default mod warning message refers to old proposals category


trichoplax:
**Describe the bug**
When issuing a moderator warning for "Many off-topic posts", there is a default wording that the moderator can modify or send as is. This contains a link to the category on Meta where proposed new sites were posted in the past, before this was replaced with a separate Proposals community.

There is also a link to chat, which does not contain a destination URL. This would be easy to overlook, resulting in a broken link that directs back to the page you are already on.

**To Reproduce**
1. As a moderator, go to a user page and click Moderator Tools, then "warn or suspend user".
2. Select the "Many off-topic posts" template.
3. Under "Review the message", notice the problems mentioned, in the third paragraph.

**Expected behavior**
- The out of date link could be updated to point to the new Proposals community.
- The chat link could either be changed to plain text so that forgetting to enter a destination doesn't lead to a confusing link back to the page you are already on, or a valid URL could be included in the default wording.

cellio:
`db/seeds/warning_templates/off-topic.md`.

PR #1404, opened 2024-09-11
updated link in mod-message template

cellio:
Updated the link in the off-topic template for how to propose a community.  This isn't ideal because it's specific to codidact.com -- so was the previous version, so it's not a regression, but this is something other instances will need to edit.  It would be better to have a generic template here and then for our prod to modify from there, but we need a better framework for doing that.  Meanwhile, at least the link isn't wrong now.

Fixes https://github.com/codidact/qpixel/issues/1356.

### DIFF
--- a/db/seeds/warning_templates/off-topic.md
+++ b/db/seeds/warning_templates/off-topic.md
@@ -3,7 +3,9 @@ We've noticed that you've written a number of posts about topics that are beyond
 You can find out about what's on-topic and what's off-topic on $SiteName in the [help center](/help/faq).
 
 This is just a gentle reminder that we expect posts on this site to stay focused on the topic on-hand.
-We have a [Network of communities](https://codidact.com/) that you are free to use; you may find one of our other communities more suitable to some of your posts. If you would like to talk about the possibility of creating a site for a subject not currently covered, please write a post on [meta](https://meta.codidact.com/categories/10) with your request.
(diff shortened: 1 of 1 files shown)


Oaphi (changes requested):
nit: there is a typo in Codidact's TLD, should read "com", not "con"

db/seeds/warning_templates/off-topic.md
```diff
@@ -3,7 +3,9 @@ We've noticed that you've written a number of posts about topics that are beyond
 You can find out about what's on-topic and what's off-topic on $SiteName in the [help center](/help/faq).
 
 This is just a gentle reminder that we expect posts on this site to stay focused on the topic on-hand.
-We have a [Network of communities](https://codidact.com/) that you are free to use; you may find one of our other communities more suitable to some of your posts. If you would like to talk about the possibility of creating a site for a subject not currently covered, please write a post on [meta](https://meta.codidact.com/categories/10) with your request.
+We have a [Network of communities](/dashboard) that you are free to use; you may find one of our other communities more suitable to some of your posts.
+If you would like to explore creating a site for a subject not currently covered, see [the Proposals process on codidact.con](https://proposals.codidact.com/help/proposals).
```
Oaphi, line 7: test

Oaphi (changes requested):

cellio:
🤦  Thanks for catching that.

Oaphi (approved):
LGTM

Issue #284, opened 2020-11-12
Expose a network profile


cellio:
You have a network *account*, and a *profile* on each community that you've visited while logged in.  But there's nothing that ties them together; to see your posts (or other information) on a community you first have to visit that community and then click on your user link.

Can we add a network profile that, at minimum, lists and links to your community profiles?  This network profile would also be a good place to attach network-wide preferences when we get to them.  And we have a [request for a unified subscription list](https://meta.codidact.com/questions/276419), which could be made available here.  (I'll file a separate issue for that.  This current one is about enablement -- we need a network profile at *all* before we can do stuff on it.)

The network profile should be linked from the community profile and publicly visible.

ArtOfCode-:
This is a good one for someone new - either new to developing, to Ruby, or to the project. It's nice and isolated, with few changes required to existing views. It'll involve creating a new controller action and a view to go with it, plus linking to it from a few places.

Trilarion:
I would be interested in working on this issue. However, I have no experience with Ruby yet (data scientist with Python, Java experience in daily work). Therefore I would need to find free time, get into Ruby, get into QPixel and into this issue. This would probably take a bit of time.

However, it's marked as good first issue and I like the proposed feature.

I will come back with more feedback once I'm more into it. But if you happen to come around implementing the feature before that, that would be fine with me too.

cellio:
Welcome!  Take a look, and please let us know if you need guidance.

Trilarion:
Setting up rails and this project with all required dependencies is quite a bit of work. I ended up compiling ImageMagick by myself, but finally I got to the "Set up QPixel" part and 

`rails db:create`

fails with

```
Cannot load database configuration:
Could not load database configuration. No such file - ["config/database.yml"]
```

I guess one has to rename config/database_sample.yml or so. Maybe the README could be adapted a bit.

One of the problems was that the Gemfile specifies Ruby 2.6.6 while I had Ruby 2.7.0. Why does the project need Ruby 2.6.6?

Trilarion:
Found an updated version of that text at:

[How to set up a new (development) instance of Codidact/QPixel](https://collab.codidact.org/posts/280451/280452#answer-280452)

ArtOfCode-:
That guide has more detail than the README, so that should work for you @Trilarion.

We don't specifically need 2.6.6, but that's the version it's built with so that's the version that's guaranteed to work, at least in theory.

Trilarion:
[...]

PR #1409, opened 2024-09-17
add profile links from other communities

cellio:
Minimally addresses https://github.com/codidact/qpixel/issues/284 by adding an "all communities" tab to the user profile to hold links to that user's profile on all other network communities where a profile exists. 

Here's a moderator viewing its own tab:

tab with table: community profile link, mod status where applicable, number of posts, rep

And here's how someone else's profile looks:

same table columns, fewer tabs

The guts of the profile tab are separated from the "container" (`network.html.erb` renders `_network.html.erb`) to make it easier to pick this up and move it to a truly global place later if we can work out how to do that.  (See discussion on the linked issue.)  Right now, your network profile is part of your user page on each specific community, just like your inbox is, even though these things are really "higher up".

To support this, I added several "accessor" functions to the user model to go fetch information on specific communities.  I followed the pattern of the pre-existing `has_ability_on` function.

Currently this table shows the number of (undeleted) posts on each community.  I was going to separate that out by post type, but post types can be defined in configuration, so that means walking the list of all post types to get their names and counts on each community (some of which might not be enabled on some communities), and I decided to defer that until we decide we need it.  If the consensus from user feedback is that we should have that, we can do it as followup work.

This initial implementation does nothing special with sorting, and I believe the order will automatically be the same as the site switcher and the dashboard.  This implementation also does not give the user a way to change the sort order.  A full implementation would be user-customizable, which requires storing settings somewhere.  Also, we need to work out the UI for custom ordering.

This implementation does not yet have any user-private information.  I once proposed a design that would show you things like where you have new inbox items or flag responses.  I'd still like to do that, but even without it, I think a basic list is better than what we have today.  As a moderator I have sometimes needed to look at a user's activity on other communities, and neither URL editing nor navigating the users list is satisfactory.

I learned a lot about Ruby/Rails while doing this and probably failed to learn other things.  I suspect there are stylistic anomalies in my code ("no don't do it _that_ way!"), which I hope reviewers will point out so I can fix them.

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -62,7 +62,7 @@ module UsersHelper
   def user_link(user, url_opts = nil, **link_opts)
     url_opts ||= {}
     anchortext = link_opts[:anchortext]
-    if anchortext != nil
+    if !anchortext.nil?
       link_to anchortext, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts)
     elsif deleted_user?(user)
       link_to 'deleted user', '#', { dir: 'ltr' }.merge(link_opts).except(:anchortext)
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -61,10 +61,13 @@ module UsersHelper
 
   def user_link(user, url_opts = nil, **link_opts)
     url_opts ||= {}
-    if deleted_user?(user)
-      link_to 'deleted user', '#', { dir: 'ltr' }.merge(link_opts)
+    anchortext = link_opts[:anchortext]
+    if anchortext != nil
+      link_to anchortext, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts)
+    elsif deleted_user?(user)
+      link_to 'deleted user', '#', { dir: 'ltr' }.merge(link_opts).except(:anchortext)
     else
-      link_to user.rtl_safe_username, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts)
+      link_to user.rtl_safe_username, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts).except(:anchortext)
     end
   end
 
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -62,12 +62,13 @@ module UsersHelper
   def user_link(user, url_opts = nil, **link_opts)
     url_opts ||= {}
     anchortext = link_opts[:anchortext]
+    link_opts_reduced = { dir: 'ltr' }.merge(link_opts).except(:anchortext)
     if !anchortext.nil?
       link_to anchortext, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts)
     elsif deleted_user?(user)
-      link_to 'deleted user', '#', { dir: 'ltr' }.merge(link_opts).except(:anchortext)
+      link_to 'deleted user', '#', link_opts_reduced
     else
-      link_to user.rtl_safe_username, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts).except(:anchortext)
+      link_to user.rtl_safe_username, user_url(user, **url_opts), link_opts_reduced
     end
   end
 
--- a/app/models/community_user.rb
+++ b/app/models/community_user.rb
@@ -28,7 +28,7 @@ class CommunityUser < ApplicationRecord
   def post_count
     Post.unscoped.where(community_id: community_id).where(user: user).undeleted.count
   end
-  
+
   # Calculation functions for privilege scores
   # These are quite expensive, so we'll cache them for a while
   def post_score
--- a/app/models/community_user.rb
+++ b/app/models/community_user.rb
@@ -24,6 +24,11 @@ class CommunityUser < ApplicationRecord
     false
   end
 
+  # All undeleted posts on this community by this user.
+  def post_count
+    Post.unscoped.where(community_id: community_id).where(user: user).undeleted.count
+  end
+  
   # Calculation functions for privilege scores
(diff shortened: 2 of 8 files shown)


ws909:
For bookkeeping, I’d suggest renaming this to «Listing profile information from across the network», since it isn’t actually a network-profile.

app/helpers/users_helper.rb
```diff
@@ -61,10 +61,13 @@ def rtl_safe_username(user)
...
-    if deleted_user?(user)
-      link_to 'deleted user', '#', { dir: 'ltr' }.merge(link_opts)
+    anchortext = link_opts[:anchortext]
+    if !anchortext.nil?
+      link_to anchortext, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts)
+    elsif deleted_user?(user)
+      link_to 'deleted user', '#', { dir: 'ltr' }.merge(link_opts).except(:anchortext)
     else
-      link_to user.rtl_safe_username, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts)
+      link_to user.rtl_safe_username, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts).except(:anchortext)
```
Oaphi, line 70: Optional: let's DRY this a little and extract `{ dir: 'ltr' }.merge(link_opts).except(:anchortext)` into a variable available for all branches and pass it instead.

Oaphi, line 70: Optional: let's DRY this a little and extract `{ dir: 'ltr' }.merge(link_opts).except(:anchortext)` into a variable available for all branches and pass it instead.

app/models/user.rb
```diff
@@ -138,6 +138,31 @@ def is_admin
...
+  end
+
+  def reputation_on(community_id)
+    cu = community_users.where(community_id: community_id).first
+    cu&.reputation
+  end
+
+  def post_count_on(community_id)
+    cu = community_users.where(community_id: community_id).first
+    cu&.post_count
```
Oaphi, line 154: Optional: let's be extra sure and return `0`  by default to avoid accidental bugs with consumers of the API doing a safe navigation check. Has an additional benefit of preventing "type hopping" by providing a stable API contract to consumers ("always returns an integer", etc).

app/models/user.rb
```diff
@@ -138,6 +138,31 @@ def is_admin
...
 
+  # Used by network profile: does this user have a profile on that other comm?
+  def has_profile_on(community_id)
+    cu = community_users.where(community_id: community_id).first
+    !cu&.user_id.nil?
+  end
+
+  def reputation_on(community_id)
+    cu = community_users.where(community_id: community_id).first
+    cu&.reputation
```
Oaphi, line 149: Optional: same as with the `post_count_on` and `reputation_on` suggestions, let's default to `1` here in case `cu` fails the safe navigation check.

app/models/user.rb
```diff
@@ -138,6 +138,31 @@ def is_admin
     is_global_admin || community_user&.is_admin || false
   end
 
+  # Used by network profile: does this user have a profile on that other comm?
+  def has_profile_on(community_id)
+    cu = community_users.where(community_id: community_id).first
+    !cu&.user_id.nil?
```
Oaphi, line 144: Optional: same as with the `post_count_on` and `reputation_on` suggestions, let's default to `false` if `cu` fails the safe navigation check.

app/models/user.rb
```diff
@@ -138,6 +138,31 @@ def is_admin
...
+    cu&.post_count
+  end
+
+  def is_moderator_on(community_id)
+    cu = community_users.where(community_id: community_id).first
+    if cu&.is_moderator || cu&.privilege?('mod')
+      true
+    else
+      false
+    end
```
Oaphi, line 163: `if ... true else false` is redundant and should never be done like that, you can simply return the boolean value directly: `cu&.is_moderator || cu&.privilege?('mod')` (that's it, without the whole construct).

Oaphi (changes requested):
Other than a single redo item (the rest are optional suggestions), LGTM at a quick glance!

cellio:
> For bookkeeping, I’d suggest renaming this to «Listing profile information from across the network», since it isn’t actually a network-profile.

When you say "renaming", where do you mean?

ws909:
> When you say "renaming", where do you mean?

PR title; currently named "basic network profile".

app/models/user.rb
cellio: Heh.  For this line I was rubocopped -- I was more explicit originally. :-)

cellio:
> Other than a single redo item (the rest are optional suggestions), LGTM at a quick glance (at least from the initial implementation standpoint)!

I addressed your comments -- with luck, in the ways you intended.

Oaphi (approved):
Leaving for others to take a quick look too, but otherwise - LGTM

app/models/community_user.rb
```diff
@@ -24,6 +24,11 @@ def suspended?
     false
   end
 
+  # All undeleted posts on this community by this user.
+  def post_count
+    Post.unscoped.where(community_id: community_id).where(user: user).undeleted.count
```
ArtOfCode-, line 29: This is going to be expensive when run for multiple users and/or communities on one page. We're _probably_ alright at current scale, but we should ultimately cache this on the CommunityUser model for each community.

cellio: It occurs to me that the community-specific post count shows up other places, including on the profile, where we call `@user.posts.undeleted.count`.  But I'm not quite sure how to do that for a non-local user -- any advice?  We're calling this from a user on site A and asking about the user on site B, C, D... .

cellio: Filed https://github.com/codidact/qpixel/issues/1411 after discussion with @ArtOfCode- .

PR #1413, opened 2024-09-20
RSS link improvements

Oaphi:
@cellio I'd like to expand a bit on your [PR](#1399) regarding the RSS links. This PR targets _your branch_ and adds:

- adds a proper configurable partial for RSS links;
- fixes the trailing whitespace underlining on hover (that's not related to your fix in response to my review, it's a general issue with our templates);
- solves the annoying misalignment of the header (see screenshot):

If you'd like, we can merge the changes into your branch and then merge to `develop`.

### DIFF
--- a/app/assets/stylesheets/categories.scss
+++ b/app/assets/stylesheets/categories.scss
@@ -16,6 +16,12 @@
     align-items: center;
     justify-content: space-between;
   }
+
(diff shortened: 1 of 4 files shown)


cellio:
LGTM, thanks!

PR #1423, opened 2024-10-02
Clean up spammers automatically

ArtOfCode-:
Clean up spam users automatically. Runs every night to clean up users who have posted, but whose posts have all since been deleted with at least one helpful spam flag on them. Feeds to STAT and soft-deletes the users, in the same way a moderator can.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -6,10 +6,14 @@ every 1.day, at: '02:05' do
   runner 'scripts/mail_uncaptured_donations.rb'
 end
 
-every 6.hours do
-  runner 'scripts/recalc_abilities.rb'
-end
-
 every 1.day, at: '02:10' do
   runner 'scripts/prune_email_logs.rb'
+end
+
+every 1.day, at: '02:15' do
+  runner 'scripts/run_spam_cleanup.rb'
+end
+
+every 6.hours do
+  runner 'scripts/recalc_abilities.rb'
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -338,13 +338,7 @@ class UsersController < ApplicationController
         return
       end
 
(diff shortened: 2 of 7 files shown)


cellio:
Do you mean all posts deleted and there was at least one helpful spam flag on one of them, or do you mean that all posts are deleted and each deleted post had at least one helpful spam flag?  If it's the latter this sounds good to me, but if it's the former, there could be edge cases where profile deletion is too strong (need to think more about it).  I think from the diff that you mean the latter (every post is deleted and had a spam flag) but I want to confirm.

With this change, I think (as followup) we need to raise the priority of change requests that make deleted users more visible -- profile links instead of "deleted user" for mods, and maybe a "recent user deletions" page in the mod tools so that mods can figure out what happened if they come across a case.

config/schedule.rb
```diff
@@ -6,10 +6,14 @@
...
 
-every 6.hours do
-  runner 'scripts/recalc_abilities.rb'
-end
-
 every 1.day, at: '02:10' do
   runner 'scripts/prune_email_logs.rb'
 end
+
+every 1.day, at: '02:15' do
```
cellio, line 13: I assume this script runs as System?  (Asking for attribution purposes on the delete.)

ArtOfCode-: It would if I'd remembered to do that...

ArtOfCode-:
> Do you mean all posts deleted and there was at least one helpful spam flag on one of them, or do you mean that all posts are deleted and each deleted post had at least one helpful spam flag?

It's the latter. All posts must be deleted and every post must have at least one helpful spam flag on it.

cellio:
From a functionality perspective, LGTM.  Someone more qualified should review the code.  (I looked at it but I don't feel qualified to sign off on it.)

Oaphi (approved):
LGTM

PR #1424, opened 2024-10-07
Add Cloudflare bot authorization

ArtOfCode-:
In mitigating for the incident over the weekend, I made the server simply ignore if it couldn't fetch other Codidact sites and move on. This has left us without a functioning communities list in the footer. This change adds an authorization header that Cloudflare should recognise and allow the request.


trichoplax:
I'm not sure if `communities.json` is used anywhere else, but a recent pull request removed the list of communities from the footer, so it isn't needed for that purpose. This was #1402 and `develop` has been deployed since that was merged, so production is no longer displaying communities in the footer.

cellio:
The community switcher also pulls a list of communities from somewhere, presumably here (but I haven't checked).

ArtOfCode-:
I can't say I know off the top of my head everywhere this is used, but seems sensible to restore it to working order until we can review properly whether it's still required.

cellio:
Let's get this reviewed by someone more competent than I am. :-)  How do we test it?

config/initializers/zz_codidact_sites.rb
```diff
@@ -4,7 +4,13 @@
   if Rails.env.development? || Rails.env.test?
     []
   else
-    response = Net::HTTP.get_response(URI('https://codidact.com/communities.json'))
+    uri = URI('https://codidact.com/communities.json')
+    req = Net::HTTP::Get.new(uri)
```
Oaphi, line 8: Let's maybe take the opportunity and untie development from codidact being up by making this configurable and not hardcoded? Also makes sense if we ever switch domains or restructure. If I am able to get some time off for that, I can push here or barring that - separately.

Oaphi (approved):
LGTM

Issue #829, opened 2022-08-13
Automatically focus search bar after clicking search (search-slide)


Taeir:
**Is your feature request related to a problem? Please describe.**
When you click the search button in the header, you cannot immediately start typing, because it does not focus the actual search field.

**Describe the solution you'd like**
Automatically focus the search field when the search button is clicked and the search bar is shown (`search-slide`)

**Additional context**
I tried messing around with some javascript to make this happen and it is slightly tricky. Sliding the header down does not trigger any events on which one could listen. If I add an additional onclick handler to the search button, I can make it focus the search, but only if the handler is added after the one which slides the element down.

As a proof of concept, adding the following javascript to `app/views/layouts/_header.html.erb` works.

```js
<script>
  $(document).ready(function () {
      $('a[data-header-slide="#search-slide"]').on('click', focusSearch);
  });

  var search = false;
  function focusSearch() {
      search = !search;
      if (search) {
          $('#search').focus();
      }
  }
</script>
```

While this works, it seems a bit dirty. One potential solution would be if co-design triggers some event, e.g. `slide_open` or `slide_close`. Then the listening behavior could be defined elsewhere.

cellio:
@luap42 can you weigh in on whether this should be done in co-design vs. qpixel?  Thanks.

cellio:
@ArtOfCode- can you weigh in on the question about doing this in co-design vs qpixel?

ArtOfCode-:
Probably better to do this in QPixel. Makes more sense for an application-specific use case than doing it as part of the design framework.

PR #1426, opened 2024-10-09
automatically focus in the textbox when you start a search

cellio:
Fixes https://github.com/codidact/qpixel/issues/829 by implementing Taeir's suggestion.  (Art approved the approach in the comments there.)


app/views/layouts/_header.html.erb
```diff
@@ -2,6 +2,20 @@
 <% mobile_logo_path = SiteSetting['MobileLogoPath'] %>
 <% sticky_header = user_preference('sticky_header', community: false) == 'true' %>
 
+<script>
+  $(document).ready(function () {
+      $('a[data-header-slide="#search-slide"]').on('click', focusSearch);
+  });
+
+  var search = false;
+  function focusSearch() {
```
Oaphi, line 11: A couple of notes:

- avoid using `var` at all costs, `const` for unchangeable cariables and `let` for reassignable is a must:
- the `focusSearch` function has a "scope bleed" - it should be called something like `toggleSearchFocus` and accept a single boolean parameter, which you then pass in the handler function. I'll provide example implementation a bit later

Oaphi (dismissed):

app/views/layouts/_header.html.erb
```diff
@@ -2,6 +2,22 @@
 <% mobile_logo_path = SiteSetting['MobileLogoPath'] %>
 <% sticky_header = user_preference('sticky_header', community: false) == 'true' %>
 
+<script>
+const toggleSearchFocus = () => {
+  const $input = $('#search');
+
+  if ($input.is(':focus')) {
+    $input.blur();
+  } else {
```
ArtOfCode-, line 11: style: new line for if-else blocks

```suggestion
  }
  else {
```

ArtOfCode- (changes requested):

trichoplax (changes requested):
I recommend this pull request be made to Co-Design instead, in the file `js_src/header.ts`, either in the method `toggle`, which starts on [line 32](https://github.com/codidact/co-design/blob/develop/js_src/header.ts#L32), or otherwise in the constructor for the class `HeaderSlideToggle`, which starts on [line 20](https://github.com/codidact/co-design/blob/develop/js_src/header.ts#L20).

The `toggle` method is added as an event listener callback to every slider, so adding the code here when it is only used for the Search slider would be slightly wasteful, but likely imperceptibly so.

Instead modifying the constructor would allow adding the existing `toggle` method to all of the other sliders, and creating a new callback function just for the Search slider, consisting of calling `toggle` and then focusing the text box. This would avoid adding the unneeded code to the other sliders, and would avoid the decision being made each time a slider is opened, instead deciding just once at page load.

trichoplax:
Straight after recommending making this change in Co-Design instead, I've just seen [Art's comment on the issue this pull request fixes](https://github.com/codidact/qpixel/issues/829#issuecomment-1384651342):

> Probably better to do this in QPixel. Makes more sense for an application-specific use case than doing it as part of the design framework.

This is a good point, and making a special case in the design framework is a little messy. Maybe it's better to keep this in QPixel for now, and avoid changing Co-Design until #1432 is worked on (which is unlikely to be soon as it seems low priority). At that point there would be no special case in Co-Design, as the browser would handle the autofocus automatically with no JS required.

trichoplax (approved):

ArtOfCode- (approved):

PR #1427, opened 2024-10-09
Ignore emacs temp files

cellio:
Can haz ignore for emacs backup files plz? :-)  We already have vim, and not having this gives me a lot of pollution in a `git status`.  I've been locally modifying it, but why not just commit it?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,9 @@ qpixel-import.tar.gz
 # Ignore Vim stuff.
 *.swp
 
+# Ignore emacs stuff.
+*~
+
 dump.rdb
 
 # Ignore IRB files


.gitignore
```diff
@@ -52,6 +52,9 @@ qpixel-import.tar.gz
 # Ignore Vim stuff.
 *.swp
 
+# Ignore emacs stuff.
+*~
```
Oaphi, line 56: Just so as we don't get bit in the rear down the line - are we sure there is no chance we get a file ending with a tilde at project root that's not related to emacs? LGTM otherwise

Oaphi (approved):
LGTM

.gitignore
Oaphi: Nvm, it's a standard pattern, all good

Issue #1400, opened 2024-09-10
On 2FA confirmation page, auto-focus on the input field


cellio:
meta:291584

When you visit the page to confirm a 2FA code that's been sent, you have to click into the text field before entering it.  On many sites that use 2FA, the field already has focus so you can send the code and then just start typing.  Can we do that too?  (I know how to do that in pure HTML but when I searched for how to do it in Ruby on Rails, I found lots of other people with that question...)

PR #1428, opened 2024-10-10
on 2fa code-response page, focus on textbox

cellio:
Focus on code input box instead of requiring a click.  There is nothing else to do on this page so that's a safe assumption.

Fixes https://github.com/codidact/qpixel/issues/1400.

### DIFF
--- a/app/views/users/sessions/verify_2fa.html.erb
+++ b/app/views/users/sessions/verify_2fa.html.erb
@@ -1,4 +1,11 @@
 <h1>Two-factor authentication</h1>
+
+<script>
+  window.onload = function() {
+      document.getElementById("code").focus();
+  };
+</script>
+
 <p>
   Your account has two-factor authentication enabled. Enter a code from your authenticator app here.
 </p>
--- a/app/views/users/sessions/verify_2fa.html.erb
+++ b/app/views/users/sessions/verify_2fa.html.erb
@@ -1,9 +1,13 @@
 <h1>Two-factor authentication</h1>
 
 <script>
-  window.onload = function() {
-      document.getElementById("code").focus();
-  };
+  addEventListener('load', () => {
+      const codeElem = document.getElementById("code");
+
+      if (codeElem) {
+	  codeElem.focus();
+      }
+  });
 </script>
 
 <p>
--- a/app/views/users/sessions/verify_2fa.html.erb
+++ b/app/views/users/sessions/verify_2fa.html.erb
@@ -1,7 +1,7 @@
 <h1>Two-factor authentication</h1>
 
 <script>
-  addEventListener('load', () => {
+  window.addEventListener('load', () => {
       const codeElem = document.getElementById("code");
 
       if (codeElem) {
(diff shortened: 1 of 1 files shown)


app/views/users/sessions/verify_2fa.html.erb
```diff
@@ -1,4 +1,11 @@
 <h1>Two-factor authentication</h1>
+
+<script>
+  window.onload = function() {
```
Oaphi, line 4: `on*`-style listeners should not be used in modern code (they have a lot of issues, for example, there can only be one at a time). Let's change it to the idiomaitc, fully supported, and newer `addEventListener('load', () => { ... })`. Same argument for switching from `function` to arrow function `() => {}` syntax.

Oaphi (changes requested):

app/views/users/sessions/verify_2fa.html.erb
```diff
@@ -1,4 +1,11 @@
...
+      document.getElementById("code").focus();
```
Oaphi, line 5: Let's do optional chaining or an `if` here: we do not want to hard fail if the element is not found (it is guaranteed now, but future-proofing it is beneficial). Either `document.getElementById("code")?.focus()` or (arguably better as optional chaining support is a bit lower still):

```javascript
const codeElem = document.getElementById("code");

if (codeElem) {
    codeElem.focus();
}
```

Oaphi (changes requested):

app/views/users/sessions/verify_2fa.html.erb
cellio: Ok.  There's a tiny bit of visible lag that I didn't see before, but if this is the correct way to do it now, I assume that's ok.  And maybe I only noticed it because I was watching for the focus.  Thanks for the education.

app/views/users/sessions/verify_2fa.html.erb
```diff
@@ -1,4 +1,15 @@
 <h1>Two-factor authentication</h1>
+
+<script>
+  addEventListener('load', () => {
```
Oaphi, line 4: Minor nit: `window.addEventListener` :) It's not wrong, but best use global explicitly

Oaphi (changes requested):

trichoplax (changes requested):
HTML also provides the [autofocus attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus) that indicates which element is to have focus on page load, which avoids the need for JavaScript. In addition to being faster, this will also work for people who have JavaScript disabled.

I don't know if there's a reason to also have the JavaScript approach. It appears that `autofocus` works with all modern browsers.

cellio:
> I don't know if there's a reason to also have the JavaScript approach. It appears that `autofocus` works with all modern browsers.

I didn't know how to do that with Ruby-generated HTML.  If it were pure HTML I'd know what to do, but the relevant code is this:

```
  <div class="form-group">
    <%= label_tag 'code', 'Code', class: 'form-element' %>
    <%= text_field_tag 'code', '', class: 'form-element', autocomplete: 'one-time-code' %>
  </div>
```

Doing this without Javascript would be better, I agree.  I'm just having a little trouble searching for it.  Ruby forums sent me to Javascript.  I probably need better search terms.

Oaphi (commented):
LGTM

Oaphi (approved):

trichoplax (approved):

PR #1430, opened 2024-10-12
Improvements for Docker setup

### LINKED ISSUE #1398: Docker configuration needs changes to work on Mac (M1+)
I tried to set up QPixel with Docker on a Mac M1 with MacOS 14.2.1 (Sonoma) and ran into many issues.  With help we got it working; this issue is to track changes for both the Docker setup and our installation doc to make it easier for the next person.  I was working with (and leaning heavily on) @Oaphi , who will probably have more to add to this list.

- We had to edit `Dockerfile` to be more explicit about architecture, replacing `# FROM ruby:3.0.0` with `FROM --platform=linux/arm64 ruby:3.0.0`.  `Dockerfile` has to work for everyone so we can't edit it in the repo, but we should document this requirement in the readme file.
- Similarly, in `docker-compose.yml`, in the `uwsgi` block, we had to add `platform: linux/amd64`.  Again, we should document this.
- We had to edit `manifest.js` to add: `//= link application.css`.  I don't understand the failure that led to this so I don't know if this is specific to this platform, a more general problem, or a gremlin.  Needs more review.
- We had issues with mysql setup, but I don't know if they were general bugs or problems with interference from the host system.  We had to drop and recreate the `root@%` and `qpixel@%` users, give `root` the ability to do grants, and then do the qpixel grants manually.  Needs more review (or testing, if we can get a clean Mac environment to test on).
- In the `uwsgi` container, we had to do `dpkg --add-architecture amd64` and then manually install `libc6:amd64` and `g++-multilib`.  I think the configuration should be able to do that for us, but if not and we need to document this, the cycle is: `apt install...`, `rake tmp:clear`, `rm -rf ./public/assets`, `rm -rf ./tmp/cache/assets`, `rails assets:precompile`, `rails restart`.

Oaphi:
Closes #1398

This PR is an overhaul of our Docker setup to make it more flexible and resilient. It includes:

- change of container `restart` policy to `on-failure:3` (meaning 3 consequent failures cause containers to exit). Our current policy of `always` sends containers in a restart loop if there is a fatal error, making it hard to debug and resolve issues;
- removed obsolete `version` field from docker-compose (just getting rid of an annoying warning);
- removed mention of `sleep` command from Docker docs as there is none (probably legacy?);
- removed manual installation of imagemagick in the container as it comes packed with ruby-3.1.2-bullseye;
- added Dockerfile.dev for setting up Docker specifically for local development purposes;
- added Dockerfile.arm for aarch64-specific setup out of the box;
- upgraded Ruby image to 3.1.2-bullseye (matches our production setup in version);
- when built with Dockerfile.dev, the uwsgi container no longer automatically starts Rails when run;
- made Dockerfile configurable via `.env` (automatically created by local-setup.sh from compose-env).

Rationale for no longer auto-starting Rails when building with Dockerfile.dev:

- allows for live debugging in case there are issues that cause hard failures (currently, the container exits);
- prevents restart loops in case of hard failures (partially mitigated already by the restart policy change);
- the additional step (connecting to the container & running a single command) is fully documented in README.md. 

Rationale for making Dockerfile used for the uwsgi container configurable:

- allows use of custom Docker build configurations as one sees fit as long as they set the `CLIENT_DOCKERFILE` environment variable;
- required to facilitate the use of Dockerfile.dev for development setups without the need to hard-code the changes;
- is backwards-compatible due to the default value of compose-env set to `docker/Dockerfile`;

Related: #1336


ArtOfCode- (approved):
I know nothing about Docker so if you say this works I believe you... I don't see anything that throws any red flags at me in this.

cellio:
For what it's worth, I don't know much about Docker either, but I was able to get this running on my Mac following the updated instructions.  It was iterative, not a clean start from the final state of the PR, but I think that's ok.

Oaphi:
> I know nothing about Docker so if you say this works I believe you... I don't see anything that throws any red flags at me in this.

Thanks! Whether there are any obvious red flags that I might've glossed over due to being busy with trying to get the aarch64 setup working is one of the main reasons I wanted as many sets of eyes as possible on this, @ArtOfCode- :) Appreciate the vote of confidence! The other reason being that there are no objections to the direction I am taking our Docker setup in (configurable Dockerfiles, switching to a bullseye image, changing our restart policy for the container with Rails, etc).

There's going to be a bunch more changes before I can mark it as ready with clear conscience (in particular, switching Node.js to be taken from a stable image rather than being pulled and installed directly during build), but nothing drastic.

Issue #1436, opened 2024-10-19
Tag synonym already exists error, when removing and re-adding with different capitalisation


Wouter17:
**Describe the bug**
When removing a tag synonym, and re-adding it in the same action (so without hitting save),  but with different capitalisation, it will give a "There were some errors while saving this tag: Tag synonyms A tag synonym with the name X already exists." (where the new tag name replaces X). It will also show both versions as being synonyms.

**To Reproduce**
1. Create a tag
2. Add a synonym
3. Remove the synonym (do not hit save)
4. Add a version of the synonym with different capitalisation
5. Press "Save"

**Expected behavior**
The new capitalisation version is saved correctly.

**Screenshots**
After removing the old one (AD) and adding the new one (ad) and pressing "save":

![Error on tag edit page as described in "Describe the bug"](https://github.com/user-attachments/assets/3057ca39-0dd0-4c0c-8078-1eab73235789)

ArtOfCode-:
Looks like this is expected behaviour. Tags with different capitalizations are considered the same tag, so it makes sense that synonyms would behave the same.

What's behind this issue? Has there been a case of a synonym not applying because someone created a tag with a different capitalization than the synonym used?

trichoplax:
My impression from the issue description is that this is a problem with renaming a synonym during editing a tag itself, rather than changing the tags during editing a specific post.

I can reproduce the error. Here are my slightly more specific steps:

1. Edit a tag to add a synonym.
2. Save.
3. Edit the same tag to remove the synonym (by pressing the "X" but not saving) and add a new synonym which has a different mix of upper and lower case.
4. Save.
5. See error.

Having a tag synonym with different case than you would like does not prevent it from functioning correctly. It will work regardless of what case a user saving a post enters. However, the system does already allow changing the case of a synonym, it just requires an additional save to make it work. For this reason it seems reasonable to improve the system, to allow making the change in a single step without saving in between.

The workaround in the meantime is as follows:

1. Edit the tag to remove the synonym that you don't like the case of.
2. Save.
3. Edit the tag again to add the synonym with the case that you prefer.
4. Save.

I believe that the intention of this issue is to avoid this need to save twice.

trichoplax:
Although changing the case of a synonym is never necessary in order to make it work correctly, it seems reasonable to be able to change it for clarity or to fix typos that only affect case.

[...]

PR #1437, opened 2024-10-20
Silently substitute synonyms for tags

ArtOfCode-:
Instead of showing an error when a user tries to use a synonym instead of the tag it points to, just substitute the synonym for the primary tag.

May assist with #1436, but I'm not sure it's the full answer.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -200,16 +200,20 @@ class Post < ApplicationRecord
 
   private
 
-  # Updates the tags association from the tags_cache.
+  ##
+  # Before-validation callback. Update the tags association from the tags_cache.
   def update_tag_associations
     tags_cache.each do |tag_name|
-      tag, name = Tag.find_or_create_synonymized name: tag_name, tag_set: category.tag_set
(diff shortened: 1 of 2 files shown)


Oaphi (approved):
LGTM

trichoplax:
The error described in #1436 relates to editing a tag itself, not editing the tags on a post, so I've added some comments there including more detailed steps to reproduce.

Note that:
- Before this pull request, editing a post to add a tag synonym would already silently replace the synonym with the relevant tag (this was already being done on the front end).
- The error described in #1436 still happens after this pull request (it is an error when editing a tag, not editing a post).

ArtOfCode-:
@trichoplax when I dug into this it wasn't _quite_ being handled on the front end - it would suggest the primary tag instead of the synonym, but it still gives the option to create a new tag. This just cleans up that loose end - anyone selecting the new tag anyway will have it replaced with the synonym's primary.

trichoplax:
Thanks for explaining - I didn't pick up that when testing it. I just pressed enter after typing in the tag name, and it was automatically replaced, so I overlooked the option to override that behaviour. Good to have this fixed.

Issue #1411, opened 2024-09-20
Cache CommunityUser post counts


cellio:
https://github.com/codidact/qpixel/pull/1409 adds the "all communities" tab on the user profile, which shows that user's profiles on the rest of the network with some stats, currently just reputation and number of posts.  Getting the number of posts per profile could get expensive at scale.  Instead, we should cache this and update when posts are added or deleted.

We are already using [counter_culture](https://github.com/magnusvk/counter_culture) in a couple other places and could use it here to address this.  We need to add a `post_count` counter to the CommunityUser model; CU doesn't have a direct relation to Post but can get there through User (`has_many :posts, through: :user` should do that, I think).  The rest of the work is to add the `counter_culture` declaration and make it work for adding or deleting posts.  That's the part I couldn't figure out, so we're spinning this off as a new issue.

PR #1438, opened 2024-10-20
Add counter for undeleted posts on CommunityUser

ArtOfCode-:
See #1411.

Adds an automatic counter for undeleted posts on CommunityUser.

> **Deploy note**  
> When this is deployed a manual process will be needed to set initial values for these counters.

### DIFF
--- a/app/models/community_user.rb
+++ b/app/models/community_user.rb
@@ -24,11 +24,6 @@ class CommunityUser < ApplicationRecord
     false
   end
 
-  # All undeleted posts on this community by this user.
-  def post_count
-    Post.unscoped.where(community_id: community_id).where(user: user).undeleted.count
-  end
-
   # Calculation functions for privilege scores
   # These are quite expensive, so we'll cache them for a while
   def post_score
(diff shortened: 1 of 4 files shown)


Oaphi (changes requested):
I presume we also no longer need the `post_count` method on the [`community_user` model](https://github.com/codidact/qpixel/blob/64aeaa700b90b2955b9d7f02d287937f8a538706/app/models/community_user.rb#L28-L30) then (apologies for a no-preview link, [review comments on unchanged files](https://github.com/orgs/community/discussions/9099) are still not supported)?

ArtOfCode-:
> I presume we also no longer need the `post_count` method on the [`community_user` model](https://github.com/codidact/qpixel/blob/64aeaa700b90b2955b9d7f02d287937f8a538706/app/models/community_user.rb#L28-L30) then?

You would be correct, removed :)

Oaphi:
LGTM!

Oaphi (approved):

cellio:
> > **Deploy note**
> > When this is deployed a manual process will be needed to set initial values for these counters.

What happens if this gets deployed and that doesn't happen?  Does it just not get cached (yet), or does something break?  (This is a question about timing: does somebody need to be able to jump on this at the time of the next deploy?)

PR #1449, opened 2024-11-11
install guide updates

cellio:
I just did a fresh dev setup using our installation doc and discovered a couple things that could be improved.  Somewhere along the way the problem with having to manually configure tag categories must've gotten fixed (it worked for me out of the box), so I removed that.  I noticed that the policy docs were not initially seeded, but the "update" seed command added them, so I added that.  And because the removed tag-set doc had included a screenshot of what qpixel in dev mode actually looks like, I added one elsewhere.


INSTALLATION.md
```diff
@@ -198,36 +200,32 @@ While being logged into your administrator account, go to [http://localhost:3000
 Review the settings (if you want; you can change them later) and click "Save and continue" to complete
 setting up the dev server.
 
-### Configure Categories
+## Create a Post
 
-Before you try to create a post we need to configure categories!
-Go to `http://localhost:3000/categories/`
+You should then be able to create a post! There are character requirements for the
+body and title, and you are required at least one tag.
```
Oaphi, line 206: I think you meant something along the lines of "and you have to add at least one tag", not that the poster is required to be tagged at least once :)

Oaphi (changes requested):

INSTALLATION.md
cellio: Well, somebody before me, but I should have caught it. :-)

Issue #1451, opened 2024-11-17
"All communities" does not show correct moderator status for global mods/admins


cellio:
The "all communities" tab on a user profile indicates which communities that use is a moderator on.  This is correct for regular appointed mods (i.e. most of them), but for people who are also global moderators or admins, the information is sometimes missing.  The *local* community always shows with mod status, but other communities do not, except where there were explicit grants.  (For example, I was explicitly granted mod status on Writing way back when and was later a global mod; Writing always shows up, and the community where you're viewing the profile does too, but others do not.)

Info from a Discord debugging session (thanks @Oaphi ):

> is_moderator on the user model is defined as is_global_moderator || community_user&.is_moderator || is_admin || community_user&.privilege?('mod') || false while in is_moderator_on we only do cu&.is_moderator || cu&.privilege?('mod')  
> we should align is_moderator_on with is_moderator, methinks  
> is_moderator is an explicit DB column
so, the full check is:
> 1. Is user a global moderator?
> 2. Is user a moderator via the is_moderator DB flag?
> 3. Is user an admin via is_admin DB flag?
> 4. Is user a moderator via the mod ability?

PR #1452, opened 2024-11-17
fix global mod check

cellio:
Fixes https://github.com/codidact/qpixel/issues/1451.


Oaphi (approved):
LGTM - we should probably change `is_moderator` method on the `user` model to accept a community user as a parameter to be able to check generically (DRY), but that's outside the scope of this PR and would require a bunch of carefully tested updates throughout the codebase (just something to keep in mind).

PR #1456, opened 2024-11-25
add policy links

cellio:
Partially in response to https://meta.codidact.com/posts/292975, added links to the ToS, CoC, spam policy, and "any other policies" to the sign-in page.  I was going to add those links to the confirmation email, but the confirmation link takes you straight to the sign-in page and this way the links are visible after the first sign-in too, so I thought this was a better place to put them. 

Screenshot of new notice

The three explicit links are "safe" in that those pages are seeded for new installations, but if anybody adds more policies (like a university with an IRB policy or a workplace with an NDA), they wouldn't be able to add those without forking the code, hence the catch-all.

### DIFF
--- a/app/views/devise/sessions/new.html.erb
+++ b/app/views/devise/sessions/new.html.erb
@@ -8,8 +8,7 @@
 <% end %>
 
 <div class="notice is-info has-margin-bottom-4">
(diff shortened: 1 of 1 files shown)


Oaphi (approved):
checked, LGTM

PR #1458, opened 2024-12-01
Fix empty ephemeral tags in post drafts if the draft is saved with no tags

### LINKED ISSUE #1457: Drafts for posts without tags load an ephemeral empty tag
meta:293037

When creating a new post without tags, draft saving receives `null` (`nil` server-side) instead of an array, causing an ephemeral non-existent tag to be shown when the draft is loaded (due to a `nil`-valued key in Redis).

Oaphi:
closes #1457 

This PR fixes an issue with saving post drafts with no tags. Due to a `null` value coming from a completely empty tag field, we end up persisting a `nil` value in Redis. This results in conversion to an empty string on load, which in turn ends up showing an empty tag box in the field.

Also introduces a daily job for automatic cleanup of drafts (for now, the only action is to non-destructively fix existing drafts).

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -602,7 +602,9 @@ class PostsController < ApplicationController
       key_name = [:body, :saved_at].include?(key) ? base_key : "#{base_key}.#{key}"
 
       if key == :tags
-        RequestContext.redis.sadd(key_name, params[key])
+        if params[key].present?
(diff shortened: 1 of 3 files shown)


cellio:
I tested and do not see the bug on this branch, but I do on `develop`.

ArtOfCode- (approved):

PR #1468, opened 2024-12-05
when search turns up nothing, say so

cellio:
Addresses https://meta.codidact.com/posts/291715.

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -22,10 +22,14 @@
               role: 'button', 'aria-label': 'Sort by age' %>
 </div>
 
-<div class="user-list">
-  <% @users.each do |user| %>
-    <%= render 'user', user: user %>
-  <% end %>
-</div>
+<% if @users.count == 0 %>
+  <p>No users found.</p>
+<% else %>
+  <div class="user-list">
+    <% @users.each do |user| %>
+      <%= render 'user', user: user %>
+    <% end %>
+  </div>
+<% end %>
 
 <%= will_paginate @users, renderer: BootstrapPagination::Rails %>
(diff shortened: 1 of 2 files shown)


app/views/users/index.html.erb
```diff
@@ -22,10 +22,14 @@
               role: 'button', 'aria-label': 'Sort by age' %>
 </div>
 
-<div class="user-list">
-  <% @users.each do |user| %>
-    <%= render 'user', user: user %>
-  <% end %>
-</div>
+<% if @users.count == 0 %>
+  <p>No users found.</p>
```
Oaphi, line 26: It's best that we do not introduce new hardcoded English-only end user-visible strings. Let's switch to `I18n.t` helper and move the message itself to a localization file (e.g., `en.edits.yml`). We should probably add a separate file containing search (`en.search.yml`) strings.

Oaphi (changes requested):

app/views/users/index.html.erb
cellio: Good idea, done.  

I thought to add i18n strings for the general and tags searches while I was in there, but: (a) tags search for a non-existent tag is still broken (500), and (b) main search has text with parameters and I don't yet know how to do that with i18n.

Oaphi: Perfect, thanks! I'll do a quick check a bit later, but should be good to go otherwise. Re:I18n - that's fine, I don't think those are in scope for this PR, we can tackle them separately. Re:500 - yeah, still needs fixing - if I have enough time to spare, will look into it as well

cellio: Thanks.  I added some info to https://github.com/codidact/qpixel/issues/1353 about the 500.  I thought maybe I could fix that too, but no, at least not with my current level of understanding.

Oaphi (approved):
LGTM

PR #1469, opened 2024-12-06
pagination breaks when there are no tags

cellio:
Fixes https://github.com/codidact/qpixel/issues/1353 .  The problem is that we were checking nil but not a count of 0, and apparently tags pagination requires there actually be something there.

I do not fully understand the problem, because it appears that other places with pagination don't have these kinds of checks, but with this check in place, I can't break tag search in my local dev environment, so I propose to try it.

### DIFF
--- a/app/views/tags/category.html.erb
+++ b/app/views/tags/category.html.erb
@@ -53,7 +53,9 @@
   <% end %>
 <% end %>
 
+<% if @tags&.size > 0 %>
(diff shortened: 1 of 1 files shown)


Oaphi (approved):
LGTM - there is a [follow up discussion](https://discord.com/channels/634104110131445811/676843271762214932/1314655211687510066) in Discord about further improvements, but it shouldn't preclude merging if we take too long. Not merging for now, but works as expected.

PR #1470, opened 2024-12-07
Fixes for some minor positioning annoyances

Oaphi:
This is a minor PR addressing positioning annoyances that drive me mad every time I see them.

Filters "delete" button before the change:

And after:

Search OAuth apps button before the change:
qpixel-search-oauth-btn-before
And after (the icon is changed from textual to icon-based to align with every other search input):
qpixel-search-oauth-btn-after

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -17,14 +17,6 @@ select.form-element {
 
   @media screen and (min-width: $screen-md) {
     flex-direction: row;
-
-    & > :first-child {
-      margin: 0 0.5em 0 0 !important;
-    }
-
-    & > :last-child {
-      margin: 0 0 0 0.5em !important;
-    }
   }
 
   & > .form-group {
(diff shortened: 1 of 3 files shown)


Oaphi:
https://github.com/codidact/qpixel/pull/1470#issuecomment-2525366004

@cellio from what I understand, they are unlinked, at least for now (may be mistaken), but can be navigated to via a direct URL (just substitute `micro_auth` with `oauth` in routes definitions). Even if they are unliked for jow, we will need them for a proper API, so it amde sense to me to fix them while I am at it.

cellio:
I totally agree on fixing them; I was just trying to figure out why I'd never seen that. :-)

app/assets/stylesheets/forms.scss
```diff
@@ -17,14 +17,6 @@ select.form-element {
 
   @media screen and (min-width: $screen-md) {
     flex-direction: row;
-
```
cellio, line 20: What is the effect of removing these?

Oaphi: @cellio no effect that I managed to notice while checking other than the positioning of the "delete" button being fixed. All the other instances have CSS that make these rules redundant.

ArtOfCode- (approved):

PR #1487, opened 2024-12-15
Add post history for help center and policy posts

### LINKED ISSUE #807: Make edit history available on help posts
Help topics can be edited by moderators, and when editing you can enter a revision comment (like for regular posts).  This implies that a history is kept in the database, but there's no way to see it through the UI.  This request is to expose edit history for help topics.  The history should be visible to anybody who can see the topic -- usually everybody, but there are a few mod-only topics.

Oaphi:
closes #807

This PR enables post history views for the Help Center and policy posts:

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -464,13 +464,9 @@ class PostsController < ApplicationController
   end
 
   def document
-    @post = Post.unscoped.where(doc_slug: params[:slug], community_id: [RequestContext.community_id, nil]).first
-    not_found && return if @post.nil?
+    @post = Post.by_slug(params[:slug], current_user)
 
-    if @post&.help_category == '$Disabled'
-      not_found
-    end
-    if @post&.help_category == '$Moderator' && !current_user&.is_moderator
+    if @post.nil?
       not_found
     end
 
--- a/app/views/post_history/post.html.erb
+++ b/app/views/post_history/post.html.erb
@@ -1,8 +1,12 @@
+<% @show_content = !!defined?(show_content) ? show_content : true %>
+
 <h1>Post History</h1>
 
+<% if @show_content %>
 <div class="item-list">
   <%= render 'posts/type_agnostic', post: @post, show_category_tag: true, show_type_tag: true, last_activity: false %>
 </div>
+<% end %>
 
 <% @history.each.with_index do |event, index| %>
   <details class="history-event" id="<%= @history.size - index %>">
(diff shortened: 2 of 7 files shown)


Oaphi:
Blocked by the system's user community users not being seeded, resulting in `/users/-1` link returning 404, as well as in some layout shenanigans (alt text for the missing user's avatar image):

Screenshot from 2024-12-15 19-37-26-1

cellio:
Thanks for working on this!  Do I understand correctly that clicking on the history button marked in red cause the in-page view also marked in red?

Oaphi:
#1487 should unblock the PR

app/controllers/posts_controller.rb
```diff
@@ -464,13 +464,9 @@ def restore
   end
 
   def document
-    @post = Post.unscoped.where(doc_slug: params[:slug], community_id: [RequestContext.community_id, nil]).first
```
cellio, line 467: Moved to post.rb -> self.by_slug, no change other than taking parameter

app/views/post_history/post.html.erb
```diff
@@ -1,8 +1,12 @@
+<% @show_content = !!defined?(show_content) ? show_content : true %>
```
cellio, line 1: not sure what `!!defined?` does (double negation?) but I assume that's my failing -- just mentioning for other reviewers to double-check this line

test/controllers/posts/help_test.rb
```diff
@@ -25,20 +25,20 @@ class PostsControllerTest < ActionController::TestCase
   test 'moderator help requires authentication' do
     get :document, params: { slug: posts(:mod_help_article).doc_slug }
     assert_response 404
-    assert_not_nil assigns(:post)
```
cellio, line 28: I'm curious about the flip from not nil to nil in these tests without the names of the tests changing.  Were they backwards before?

app/views/post_history/post.html.erb
Oaphi: This is a safe check for a variable being provided at all (when calling `render` with the template that uses the variable) and assigning a default otherwise, we use the same pattern in many partials. `defined?` prevents a runtime error.

test/controllers/posts/help_test.rb
Oaphi: We've already [discussed this](https://discord.com/channels/634104110131445811/1306699834262032476/1317861103434469407) briefly with @ArtOfCode- in chat. Short version: the access check _used to_ happen _after_ the variable `@post` has been assigned, now it doesn't even happen if the user doesn't have access (as the check itself has been abstracted away), hence if the route responds with `404`, it is guaranteed to not assign the variable as well.

ArtOfCode- (approved):

PR #1490, opened 2024-12-16
Add seeding of the system user's community users

### LINKED ISSUE #1489: System user's community users are not seeded
We have a special user `-1` that performs a bunch of system-related operations. We also have a seed for it, however, that seed does not create the user's community users, leading to it being only available if one goes and manually creates those users on all communities on their installation.

Oaphi:
closes #1489 

blocks https://github.com/codidact/qpixel/pull/1487

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -21,6 +21,16 @@ sorted = files.zip(types).to_h.sort do |a, b|
   (priority.index(a.second) || 999) <=> (priority.index(b.second) || 999)
 end.to_h
 
+def expand_communities(type, seed)
+  if type.column_names.include?('community_id') && !seed.include?('community_id')
+    # if model includes a community_id, create the seed for every community
+    Community.all.map { |c| seed.deep_symbolize_keys.merge(community_id: c.id) }
+  else
+    # otherwise, no need to worry, just create it
+    [seed]
+  end
+end
+
 sorted.each do |f, type|
   begin
     processed = ERB.new(File.read(f)).result(binding)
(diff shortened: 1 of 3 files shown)


Oaphi:
Note: I opted for `trust level` equal to `4`, but not sure what's the trust level for the user's community users on production communities - happy to adjust if the assumption is incorrect.

cellio:
> Note: I opted for `trust level` equal to `4`, but not sure what's the trust level for the user's community users on production communities - happy to adjust if the assumption is incorrect.

In production "trust level" is NULL, but mine and Art's are both 5.  I'm not sure what we use this for today other than category read access (which System wouldn't care about).  I'd guess we should go with NULL, and I'll defer if another reviewer says something different.

db/seeds.rb
```diff
@@ -16,11 +16,75 @@
...
+
+  system_users.each do |su|
+    abilities = Ability.unscoped
+      .where(internal_id: ['everyone', 'mod', 'unrestricted'])
+      .where(community_id: su.community_id)
+
+    user_abilities = UserAbility.unscoped.where(community_user_id: su.id)
+
+    abilities.each do |ab|
+      unless user_abilities.any? { |ua| ua.ability_id == ab.id }
```
cellio, line 78: I'm having trouble understanding this line of code.  Dropping this marker here so I can ask you about it in chat.  (I'm assuming this is a "me" problem and not a code clarity problem.)

Oaphi: Perfect, thanks! I'll change it to `nil` then - level 4 was just a placeholder value until someone with production access could verify.

Oaphi:
> > Note: I opted for `trust level` equal to `4`, but not sure what's the trust level for the user's community users on production communities - happy to adjust if the assumption is incorrect.
> 
> In production "trust level" is NULL, but mine and Art's are both 5. I'm not sure what we use this for today other than category read access (which System wouldn't care about). I'd guess we should go with NULL, and I'll defer if another reviewer says something different.

Perfect, thanks! I'll change it to nil then - level 4 was just a placeholder value until someone with production access could verify.

db/seeds.rb
Oaphi: This part skips any user ability that is already granted - there might be a more elegant way of doing this, but I couldn't think of anything better (yet)

cellio:
LGTM.  Needs review by a better Rubyist.

ArtOfCode- (approved):

PR #1491, opened 2024-12-16
Fix server error on user profile vote summary tab if there are orphaned votes

### LINKED ISSUE #1488: Trying to get votes summary for a user might result in server error if one of the posts are somehow destroyed
Noticed live on [Writing Codidact](https://writing.codidact.com/users/-1/vote-summary). Trying to open the vote summary tab of the system user results in server error 500. This is most likely (repro'd locally) is due to a missing `post_id` on one (or more) of the retained votes (the column is nullable). While it's expected behavior, the vote list doesn't have checks for missing posts, hence the error.

Oaphi:
Closes #1488

This PR fixes the issue with orphan votes (with `post_id` but no actual post) by:
- making the vote_summary view resilient to entries with missing posts;
- making the UsersController#vote_summary filter out votes without associated posts;
- adding a `CleanupVotes` daily job that removes orphaned votes (those that do not correspond to any post but still have a `post_id` assigned to them);
- making Vote model more resilient to `post` not being present;


app/jobs/cleanup_votes_job.rb
```diff
@@ -0,0 +1,17 @@
+class CleanupVotesJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    Community.all.each do |c|
+      RequestContext.community = c
+      orphan_votes = Vote.all.reject { |v| v.post.present? }
+      puts "[#{c.name}] destroying #{orphan_votes.length} #{'orphan vote'.pluralize(orphan_votes.length)}"
```
cellio, line 8: Just as a safety net, in case we ever accidentally destroy a valid vote (maybe I'm being paranoid): can we record the rows we're nuking somewhere (output, log file, whatever)?  This seems very unlikely, but also, a user saying "wait where'd all my votes go???" would be bad.

cellio:
LGTM based on inspection (one paranoid request).  I appreciate the extra resilience in the view and controller.

Someone with more knowledge of the scripts/daily jobs should also review this.

How do we test it?  Can we inject some bad data on the dev server and test there before going to production?

app/jobs/cleanup_votes_job.rb
Oaphi: Sure, I can add a log to the successful case of `destroy` as well, won't hurt (even though we definitely should have backups to restore from even if something extra gets dropped).

cellio: We do have backups, yes.  I was thinking vaguely that it might be helpful to be able to assess the scale of the problem (like how many users were affected).

Oaphi:
> How do we test it?

There are several ways of doing so. A simple prerequisite is voting on something, then intentionally breaking the database integrity by updating the record in the `votes` table to contain a non-existent `post_id`. If done right, you will see an entry in the target user's vote summary tab that says "Post not found".

After that's done, it's just a matter of executing the job. The simplest approach is to change the associated script file to call `perform_now` instead of `perform_later` and simply run it as a script via `bin/rails runner <path_to_script>` (in fact, this is basically what whenever does under the hood). There is no reason to test the schedule as it's managed by whenever (it sets up a cron job), although that can be done too

app/controllers/users_controller.rb
```diff
@@ -574,18 +574,21 @@ def my_vote_summary
   end
 
   def vote_summary
-    @votes = Vote.where(recv_user: @user) \
-                 .includes(:post).group(:date_of, :post_id, :vote_type)
-    @votes = @votes.select(:post_id, :vote_type) \
-                   .select('count(*) as vote_count') \
-                   .select('date(created_at) as date_of')
+    @votes = Vote.where(recv_user: @user)
+                 .joins(:post)
```
ArtOfCode-, line 578: As far as I can see these two diffs are identical (excepting blank space) except this line - changed from an `includes` to a `joins`. Unless Rails has changed something since I last knew about it, `joins` _doesn't_ preload the referenced table, which is what `includes` does. What's the reasoning behind this change?

config/schedule.rb
```diff
@@ -18,6 +18,10 @@
   runner 'scripts/cleanup_drafts.rb'
 end
 
+every 1.day, ay: '02.25' do
```
ArtOfCode-, line 21: Typo alert!

```suggestion
every 1.day, at: '02:25' do
```

ArtOfCode- (changes requested):

config/schedule.rb
Oaphi: aye, d'oh, fixed!

app/controllers/users_controller.rb
Oaphi: Fixed as per discussion - I feared that'll require a rewrite of the query, but after looking into it, I realized that the changed version does not require `joins` and can make do with `includes` without sacrificing the fix.

ArtOfCode- (approved):

Issue #1497, opened 2024-12-27
unsafe ID reference causes 500 on post


cellio:
meta:292507

Trying to access https://writing.codidact.com/posts/20382 without mod/admin privileges produces a 500 error, but the page loads fine for me as an admin.  This is an imported post with, among things, a pending suggested edit from a now-deleted user.  The error log (UUID 6d9d002a-a045-4363-9799-255e6ad7bf2e) says:

```
    NoMethodError: undefined method `id' for nil:NilClass

/var/apps/qpixel/app/views/posts/_expanded.html.erb:158:in `_app_views_posts__expanded_html_erb__3248471424673514599_73080'
/var/apps/qpixel/app/views/posts/show.html.erb:40:in `block in _app_views_posts_show_html_erb__3660260357249465092_73020'
/var/apps/qpixel/app/views/posts/show.html.erb:39:in `_app_views_posts_show_html_erb__3660260357249465092_73020'
```

Line 158 is:
```
<% elsif post.pending_suggested_edit.user.id == current_user&.id %>
```

And I speculate that this ID access is our problem.

PR #1498, opened 2024-12-27
made user check nil-safe

cellio:
Fixes https://github.com/codidact/qpixel/issues/1497 .

I can't reproduce the problem in my dev environment.  Possibly this is a data problem left over from old-style deletions, which were more thorough.  I saw no change in dev before/after this change, so it seems to do no harm, and based on code inspection, it seems to fix a correctness error.  Looking for a reviewer to sanity-check that.


Oaphi:
LGTM

Oaphi (approved):

Issue #1500, opened 2024-12-29
Don't render links in profile until a user earns Participate Everywhere 


cellio:
There's been a lot of discussion about spam of various sorts.  Sometimes a user creates an account but never posts (so no posts violate spam policies), *but* the profile itself is a spam pitch with links.

Profiles exist so that users can talk about themselves and things they care about, so "self-promotion" isn't a reason to restrict the profile content.  However, we don't need to provide "SEO juice" to such links.  Thus, I propose that we not *render* links in profiles where the user has not yet gained the Participate Everywhere ability, which is the catch-all for dropping new-user restrictions.

When doing this, let's add a notice on the "edit profile" page somewhere that links won't be rendered until that ability is gained.  We don't want to surprise new users with this.

Similarly, we would show the "website" link as text but not yet as a link.  I don't see a reason to muck with the Twitter link.

Related (might share some underlying code): https://github.com/codidact/qpixel/issues/1381

PR #1501, opened 2024-12-30
in user profiles, restrict visibility of links for new users

cellio:
Fixes https://github.com/codidact/qpixel/issues/1500.

`sanitize` only accepts an allow list, so I couldn't express "allow anything we normally do, except `a`".  I instead blocked all attributes, so the `href` doesn't work and thus the `a` doesn't render.  There might be a more elegant way to do this; writing a custom sanitizer seemed like overkill and I didn't figure out anything else from Rails doc.

This strips links if the viewer is not logged in and the profile user does not have "participate everywhere".   Note for testing: in a new dev environment users come in with that ability by default, so remember to suspend it for your test user.

Screenshots:

1. Logged-out view of account in good standing:

Screenshot

2. Logged-out view of restricted account:

Screenshot

3. Logged-in view of restricted account (account in good standing is unchanged):

Screenshot

### DIFF
--- a/app/views/users/edit_profile.html.erb
+++ b/app/views/users/edit_profile.html.erb
@@ -41,6 +41,7 @@
   <%= render 'shared/body_field', f: f, field_name: :profile_markdown, field_label: 'Profile', post: current_user,
             cur_length: current_user.profile_markdown&.length, min_length: 0 %>
 
+  <p>Note: Links are not shown publicly until you have earned the Participate Everywhere ability.</p>
   <div class="post-preview"></div>
 
   <div class="grid">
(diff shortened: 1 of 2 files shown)


cellio:
I think this is ready now.  After the initial commit, I applied the same criteria to the web site link (if present) -- we just don't show the field at all in this case, rather than trying to do something to indicate that it's there but hidden.  I also added a not-very-pretty notice to the edit page, and would welcome improvements from people who grok CSS better than I do.

ArtOfCode- (approved):
This is a really smart way of doing this. Admittedly I haven't sat for hours thinking about this, but when the topic has come up before now I haven't come up with an easy way to go about it - this is ideal! Thanks!

Issue #1362, opened 2024-07-05
Inconsistent display for a user with the Edit ability suspended


trichoplax:
**Describe the bug**
A user with the Edit ability, who has had that ability suspended, shows inconsistent indicators.

**To Reproduce**
1. As a moderator, suspend a user's Edit ability.
2. As that user, view a post that is not your own and see a "Suggest edit" button (as expected).
3. As a third user, view the other user's profile page and see "Edit Posts" listed in the right hand panel under "Earned Abilities".
4. As the third user, click on "Abilities" and see a table showing "no" in the "Earned" column for "Edit Posts".

**Expected behavior**
As the user profile page and the Abilities page are both visible to a third user, they should show consistent information.

**Screenshots**
![Screenshot 2024-07-05 at 14-23-07 DEV QPixel - Test post by a moderator for testing](https://github.com/codidact/qpixel/assets/6706597/4a51133f-69b3-4884-8eab-8010397e9d03)
![Screenshot 2024-07-05 at 14-22-18 DEV QPixel - User user](https://github.com/codidact/qpixel/assets/6706597/1f88eef4-5d51-426d-8985-9bf4628d45d7)
![Screenshot 2024-07-05 at 14-21-49 DEV QPixel](https://github.com/codidact/qpixel/assets/6706597/1032bc10-c960-43c2-96ff-f226d6e9326a)

cellio:
This is a more general issue, not specific to suggested edits: suspended abilities are still listed as earned, and you have to click through to see otherwise.  We should exclude suspended abilities from the list shown on the profile.  (Technically they were *earned*, but they aren't currently *active*.)

trichoplax:
Should this be displayed the same way for the user affected by the ability suspension and other users? Alternatively the affected user could have an indication of the time remaining in their ability suspension (without showing this to other users).

PR #1502, opened 2025-01-02
show only active abilities on profile

cellio:
Fixes https://github.com/codidact/qpixel/issues/1362.

I opted for the simpler approach of just restricting the list, instead of showing abilities as earned but suspended.  A user with a suspended ability sees more info about that when clicking through to the full list, and I'd rather not try to jam more information into a small space here too.


cellio:
Got some accidental branch pollution here.  Withdrawing so I can do this right.

Issue #1480, opened 2024-12-10
Large gap between name and text on profile page


cellio:
After we fixed https://github.com/codidact/qpixel/issues/1440, I started noticing that the gap on the profile page between the user name and text that follows is larger than I expected:

![screenshot](https://github.com/user-attachments/assets/5896d4c0-9311-4647-9add-74a0c7ae5279)

I thought this might be a simple matter of tweaking `app/views/users/show.html.erb`, but the existing divs and classes are throwing me.  I suspect the problem is somewhere in this block:

```
<div class="grid <%= deleted_user?(@user) ? 'deleted-content' : '' %>">
  <div class="grid--cell is-9-lg is-12">
    <div class="h-p-6 h-p-t-2">
      <div class="profile-text">
      <p>...
```

The first grid, I think, is setting up the two columns and probably not part of the problem here [1].  I'm not sure whether I should be trying to change the second grid, the divs, the CSS behind one of the six classes involved, or something else.

[1] The profile picture, the top of the right column, is a little higher than the first text below the name in the main part of the page.  This makes me think the problem is in or below the second grid.

2S00B5:
I want to work on this issue

Oaphi:
Feel free to - just remember to follow the [contribution guidelines](https://github.com/codidact/qpixel?tab=readme-ov-file#contributing)

PR #1512, opened 2025-01-12
Reduce gap on profile page

2S00B5:
I have tried to reduce the gap between the user name and the succeeding text. Hope it's what you're looking for.  And I have tested the changes in the inspect element on the webpage.

Fixes https://github.com/codidact/qpixel/issues/1480.


ArtOfCode- (changes requested):
Has this been tested? I'm not sure this will work as it currently stands - the element that you've added that tries to reduce the margin is before the element that the margin is on.

Regardless, attempting to fix this by adding more markup/CSS isn't ideal - this can and should be fixed using the existing styles available. I'd suggest looking at [the design framework](https://design.codidact.org) to see what's available. In this case, adding a `h-m-b-0` to the `h1` and changing `h-p-t-2` to `h-p-t-0` on the profile container should resolve this without any additions.

2S00B5:
Thank you for your feedback, I'll take a look at it again asap.

Oaphi:
Do note that the gap is not caused by the header, but by the content that follows it (see other tabs for reference).

Oaphi:
While we are at it, can you please take a look at filtering out the empty paragraph from from the profile description if the user doesn't have neither website, Twitter, or Discord link added (not an issue with your PR, but we might as well address it here)? This results in a an empty `<p>` element - while it doesn't cause a layout shift, it's best we remove it.

https://github.com/codidact/qpixel/blob/91c09029d7e40abe3491693ee6e5774d1fed6a56/app/views/users/show.html.erb#L27-L48

Oaphi:
More on point, even with the changes the misalignment is still present, some further tweaking is needed - it's not as simple of a fix as just reducing the top padding (we should probably remove the padding on the left of the activity column as well while we are tweaking the layout as it's causes an additional visual shift).

For reference, compare the updated version form the new PR (might look better with profile links filled in, but many users do not have those):

Screenshot from 2025-01-13 16-57-46

To how the header looks like on the "Vote Summary" tab:

Screenshot from 2025-01-13 16-57-49

2S00B5:
Yes while editing I thought of removing the side padding but then thought it might be unnecessary! Sure I will implement the changes. Please let me know if I need to look into any other parts. Thank you so much.

Oaphi:
@2S00B5 doesn't seem like there's anything else for now, thanks! Left padding is optional, probably will need more thought put to it (don't recall the reason why it's there)

app/views/users/show.html.erb
```diff
@@ -22,30 +23,36 @@
 
 <div class="grid <%= deleted_user?(@user) ? 'deleted-content' : '' %>">
   <div class="grid--cell is-9-lg is-12">
+<<<<<<< Updated upstream
     <div class="h-p-6 h-p-t-2">
+=======
+    <div class="h-p-0 h-p-t-0">
+>>>>>>> Stashed changes
```
Oaphi, line 30: Note that you've committed with merge conflict marker active

app/assets/stylesheets/users.scss
```diff
@@ -43,6 +43,11 @@
   }
 }
 
+.reducegap {
+  font-weight: bold;
+  margin-bottom:-20px;
+}
+
```
Oaphi, line 50: That (as well as its usage) seems like an artifact of updating the branch - didn't @ArtOfCode- already review this and asked not to use the override?

app/views/users/show.html.erb
2S00B5: I apologize, this is my first time contributing to an open source project, so I followed the instructions for merging available online. Must not have worked out fine.

2S00B5:
I have implemented the changes in the individual files manually for the gap removal.

app/assets/stylesheets/dist/users.css
ArtOfCode-, line 1: Not sure how this has ended up here - there should be no `dist` folder here at all. The following git commands should amend your commits to remove it:

1. Checkout a new temporary branch from this one. Make sure you're on this branch, then `git checkout -b temp-branch`
2. Switch back to this branch: `git checkout 2S00B5/1480/large-textheader-gap`
3. Reset back to the commit that added this file: `git reset --hard b65c61d`
4. Remove the folder: `rm -r app/assets/stylesheets/dist`
5. Add the change: `git add app/assets/stylesheets/dist`
6. Amend the commit: `git commit -m "Implemented review" --amend`
7. Cherry-pick the subsequent commits back to this branch: `git cherry-pick 28562e2^..4be3438`

You can then delete the temporary branch and force-push to this one, which should solve this.

ArtOfCode- (changes requested):

app/views/users/show.html.erb
cellio: Sorry for the hassles.  Git can be a little finicky with merges sometimes.  I needed some help from more experienced devs when I came here too.  Please don't be discouraged; unfortunately it's kind of normal.

app/assets/stylesheets/dist/users.css
2S00B5: I believe the file has been removed now. Please let me know.

app/views/users/show.html.erb
2S00B5: I really appreciate your immense patience and support while I learn all of this from scratch!!

ArtOfCode- (approved):

ArtOfCode-:
Looks good - merged. Congrats on your first contribution to open-source!

2S00B5:
Thank you so much, It was an enriching experience working on this project.

cellio:
Thanks @2S00B5 !  I hope you'll continue to contribute (to Codidact and to open-source software in general).

Issue #1509, opened 2025-01-09
Tracking issue for improving test coverage for search


Oaphi:
#1507 highlighted that our testing coverage for search-related functionality is a bit lacking. This issue exists to track improvements to it to prevent such issues going forward.

- [ ] add tests for the `SearchHelper` helper;
- [ ] add integration / system tests for search;

PR #1513, opened 2025-01-12
Testing improvements

Oaphi:
followup for #1509 

Also introduces several GitHub actions that should take us off CircleCI:

- Rubocop (see rubocop.yml);
- Rails tests (see test.yml);
- CodeCov coverage uploads as part of Rails tests;
- screenshot upload as part of Rails system tests;
- dev server deployment on success;

And fixes a vulnerability related to search:

- categories with trust level higher than the requesting user no longer show up in search;


app/helpers/search_helper.rb
```diff
@@ -1,12 +1,12 @@
 module SearchHelper
-  def check_posts_permissions
-    (current_user&.is_moderator || current_user&.is_admin ? Post : Post.undeleted)
+  # @param user [User] user to check
+  def get_accessible_posts(user)
```
ArtOfCode-, line 3: Ruby convention is not to use `get_` or `set_` - a getter method is named after what it returns, and a setter uses `!` after the name. I'd suggest `accessible_posts_for(user)` here.

Oaphi: Yeah, old non-Ruby habits, @ArtOfCode-, will do

app/helpers/search_helper.rb
```diff
@@ -1,12 +1,12 @@
 module SearchHelper
-  def check_posts_permissions
-    (current_user&.is_moderator || current_user&.is_admin ? Post : Post.undeleted)
+  # @param user [User] user to check
+  def accessible_posts_for(user)
+    (user&.is_moderator || user&.is_admin ? Post : Post.undeleted)
       .qa_only.list_includes
   end
```
ArtOfCode-, line 6: Should this also check category access settings here? We've got an open security report that posts in private categories sometimes show in search; this might be an ideal place to check that.

Oaphi: Yup, looks like just the place to do so - will update once I am done with basic test coverage for the helper (want to make sure there are no other surprises first)

Oaphi: Or actually sooner than that as properly testing the `:category` qualifier requires the check to work correctly.

Oaphi: After reading through the [security report](https://github.com/codidact/qpixel/security/advisories/GHSA-pv74-hcg9-65r4) itself, I realized that it doesn't seem to be related to search, @ArtOfCode-, am I missing something? I abstracted the check for category access into `accessible_categories_for` just in case, but the tests show that search handles category access correctly.

ArtOfCode-: That's a different report - there's another one open in the support tickets that hasn't made it to a GH advisory yet.

Oaphi: Ah, got it, there is another one. Hmm, looks like that one might crop up when there is no filter by category. Will take a look

Oaphi:
Forth time's the charm, I guess. Rubcop GitHub Action is functional

app/models/user.rb
```diff
@@ -50,6 +50,12 @@ def self.search(term)
     where('username LIKE ?', "%#{sanitize_sql_like(term)}%")
   end
 
+  # Safely gets the user's trust level even if they don't have a community user
+  # @return [Integer] user's trust level
+  def trust_level
+    community_user&.trust_level || 0
```
Oaphi, line 56: Had to restore the `trust_level` method on user to avoid having to explicitly check that the community user exists

app/helpers/search_helper.rb
Oaphi: Deferring the second vulnerability fix (for the publicly reported one) to another PR - this now takes us over the finish line with switching pipelines, so it's best not to stall

Oaphi:
Just added a status badge to display in the README, no meaningful changes

cellio (commented):
LGTM.  I'm not qualified to evaluate the workflow part but I did read through all the tests.

Oaphi:
> LGTM. I'm not qualified to evaluate the workflow part but I did read through all the tests.

If you have time to spare, please do check whether everything works as expected in regards to the vulnerability fix (see the PR's description for details), that's the most important part (provided you haven't already, of course)

cellio:
Wait I am on the wrong branch. 🤦

cellio:
This turned out to be a data error (discussed in chat).  It's now working for me and my moderator user gets posts from a mod-restricted category in search results, but normal users do not.

PR #1517, opened 2025-01-14
Fix global site settings cache

ArtOfCode-:
When updating site settings, although we nominally [bust the cache](https://github.com/codidact/qpixel/blob/develop/app/controllers/site_settings_controller.rb#L52) and it would include nil as the community ID, our cache setup automatically [includes the current community ID](https://github.com/codidact/qpixel/blob/develop/lib/namespaced_env_cache.rb#L76) as a hidden part of the key. This is particularly evident with global site settings, where changes to the value enacted on one community don't propagate to others because there's a separate cache for each community.


Oaphi (approved):
LGTM with the fix to `include_community` option (would like someone else to take a look due to that just in case).

PR #1519, opened 2025-01-17
Notify user on ability grant

### LINKED ISSUE #1419: Send inbox notification when an ability is earned/granted
Currently, abilities are granted when the criteria are met, via a daily script, but the changes are silent -- you have to notice that you now have an ability.  New abilities are a big deal; let's at least send the user an inbox notification.  (Broader publicity should be considered as part of https://github.com/codidact/qpixel/issues/1360.)

This raises the question of whether, for transparency, we should also send a notification if an ability is revoked or suspended.  I think we should, though we currently have no way to send an accompanying message as part of the suspension/revocation.  (Mods can of course send a separate mod message.)  We don't want to force people to ask publicly on meta about what should be private matters, but we also don't want to surprise them when they try to do something they thought they could do but can't any more.  The new mod tools will give us private comment threads, so maybe we should wait for that to do this part?

ArtOfCode-:
Closes #1419 

Adds a notification to the user on granting an ability. Also some related code cleanup.

### DIFF
--- a/app/models/community_user.rb
+++ b/app/models/community_user.rb
@@ -77,11 +77,21 @@ class CommunityUser < ApplicationRecord
     UserAbility.joins(:ability).where(community_user_id: id, abilities: { internal_id: internal_id }).first
   end
 
-  def grant_privilege(internal_id)
+  ##
+  # Grant a specified ability to this CommunityUser.
+  # @param internal_id [String] The +internal_id+ of the ability to grant.
+  def grant_privilege!(internal_id)
     priv = Ability.where(internal_id: internal_id).first
     UserAbility.create community_user_id: id, ability: priv
+    user.create_notification("You've earned the #{priv.name} ability! Learn more.", ability_url(priv))
   end
 
+  ##
+  # Recalculate a specified ability for this CommunityUser. Will not revoke abilities that have already been granted.
+  # @param internal_id [String] The +internal_id+ of the ability to be recalculated.
+  # @param sandbox [Boolean] Whether to run in sandbox mode - if sandboxed, the ability will not be granted but the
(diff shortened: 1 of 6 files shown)


app/models/community_user.rb
```diff
@@ -77,11 +77,26 @@ def privilege(internal_id)
...
+  # Grant a specified ability to this CommunityUser.
+  # @param internal_id [String] The +internal_id+ of the ability to grant.
+  # @param notify [Boolean] Whether to send a notification to the user.
+  def grant_privilege!(internal_id, notify: true)
     priv = Ability.where(internal_id: internal_id).first
     UserAbility.create community_user_id: id, ability: priv
+    if notify
+      community_host = priv.community.host
+      user.create_notification("You've earned the #{priv.name} ability! Learn more.",
+                               url_helpers.ability_url(priv, host: community_host))
```
Oaphi, line 90: `ability_url` needs to use the `internal_id` of the privilege - otherwise, clicking on the notification results in navigating to `/abilities/<numric_id>`, causing 404 (alternatively, what do you think about making this more robust by allowing navigation to the ability page by numeric id as well [i.e., `/abilities/2`]?).

Oaphi (changes requested):
RC'ing just in case we accidentally get it merged without solving the 404 issue as otherwise - LGTM.

app/models/community_user.rb
ArtOfCode-: Huh, I read the route and saw `:id` and assumed numeric ID rather than `internal_id`... whoops.

Oaphi (approved):
Perfect, no 404 after the fix - LGTM

Issue #1381, opened 2024-08-16
Nerf posts that have been flagged as spam


cellio:
Related to meta:291174

The linked meta post requests a spam reaction to alert readers, but that risks people reacting but not flagging (and then the post never gets addressed).  Also, reactions are not anonymous and maybe this should be.  An [answer](https://meta.codidact.com/posts/291174/291177#answer-291177) suggests disabling links if a post has a spam flag.  Let's build on that.

Proposal: if a post has a threshold number of spam flags, disable links.  Because that could be confusing, also add a notice along the lines of "links have been removed pending human review" to the bottom of the post.  Spam isn't the only problem case here (there's also malware), so I'd rather not say "spam" in the notice, but we can certainly wordsmith this.

As for the threshold: a single flag seems consistent with close votes and our current small size, but I don't know if that will always be true, so let's code it in a way that's easy to change later, or maybe even make it a site setting.

ArtOfCode-:
The easy way to do this would be a JS flag that sets links' `href` to `javascript:void(0)` if the post has been flagged as spam. That wouldn't neutralise links for users without JS, though.

The harder but proper way to do it would be to trigger an automatic edit (including post history event) when a spam flag is created to edit links out of the post.

cellio:
The easy way sounds reasonable to me.  Plus, if we do it the harder but proper way, we also need to catch if those flags are declined and update the edit history again.  A runtime check for "is there currently a spam flag?" seems more resilient.  (_Also_, we wouldn't want to attribute those edits, so that's more complexity.)

Oaphi:
Can't we just derender links for posts with spam flags as it's trivial to fet that info via a query? Will require a couple of changes to the views and the controller, but worth it in the long run, imo.

This way we also have the benefit of, if users are fast enough, to prevent those links from being forever embedded into scraper sites and the IA.

ws909:
> Can't we just derender links for posts with spam flags as it's trivial to get that info via a query? Will require a couple of changes to the views and the controller, but worth it in the long run, imo.
> 
> This way we also have the benefit of, if users are fast enough, preventing those links from being forever embedded into scraper sites and the IA.

[...]

PR #1520, opened 2025-01-20
guardrails for posts with pending spam flags

cellio:
If a post has pending spam flags, warn logged-in users and suppress links for anonymous users.  There are currently no exceptions for the status of the target user; should there be?  (Does this allow someone to harass curators by decorating their posts with warnings until the flags get declined?)

Fixes https://github.com/codidact/qpixel/issues/1381.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -202,6 +202,14 @@ class Post < ApplicationRecord
     false
   end
 
+  # The test here is for flags that are pending (no status). A spam flag
+  # could be marked helpful but the post wouldn't be deleted, and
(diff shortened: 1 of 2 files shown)


cellio:
I am aware that this adds to the code that depends on a name that came from a YAML (configuration) file, which is inherently a little fragile.  If system code depends on a specific flag type, it would be better to bake that in.  But reseeding that in prod now also seems a little scary, so since this is not the first case of this (we were already doing that), I don't want to try to do that as part of this PR.

ArtOfCode- (approved):

Issue #1522, opened 2025-01-23
Notice about links not being shown in profiles until Participate Everywhere ability is earned, even when it is already earned


MoshiKoi:
**Describe the bug**

When editing one's profile, a short note is shown which says

> Note: Links are not shown publicly until you have earned the Participate Everywhere ability.

This to me feels like it's saying that I haven't yet earned the Participate Everywhere ability, and that is why links are not shown. In fact, I did have the Participate Everywhere ability, and so I was able to add links to my profile.

**To Reproduce**

1. Have the Participate Everywhere ability
2. Edit your profile

**Expected behavior**

The notice is not shown, or it has a clarification that it doesn't apply to the current user.

cellio:
Oh yeah, making that context-sensitive rather than static probably would have been a better idea.  Sorry for the confusion.

PR #1523, opened 2025-01-23
only show the warning about links if it applies to you

cellio:
Fixes https://github.com/codidact/qpixel/issues/1522.

### DIFF
--- a/app/views/users/edit_profile.html.erb
+++ b/app/views/users/edit_profile.html.erb
@@ -41,7 +41,7 @@
   <%= render 'shared/body_field', f: f, field_name: :profile_markdown, field_label: 'Profile', post: current_user,
             cur_length: current_user.profile_markdown&.length, min_length: 0 %>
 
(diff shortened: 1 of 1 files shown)


cellio:
It's the ND SSL test failure again.  This view change seems pretty unrelated to me.

app/views/users/edit_profile.html.erb
```diff
@@ -41,7 +41,9 @@
   <%= render 'shared/body_field', f: f, field_name: :profile_markdown, field_label: 'Profile', post: current_user,
             cur_length: current_user.profile_markdown&.length, min_length: 0 %>
 
-  <p>Note: Links are not shown publicly until you have earned the Participate Everywhere ability.</p>
+  <% if !current_user.community_user.privilege?('unrestricted') %>
```
ArtOfCode-, line 44: For "if not" without any else clauses, use `unless`

ArtOfCode- (changes requested):

ArtOfCode- (approved):

Issue #1248, opened 2023-11-26
Generalize external links on user profile


cellio:
meta:288948

Currently, the user profile allows adding three "external" connections: web site, Discord ID, Twitter handle.  We should keep Discord until we have a better chat solution, but otherwise, let's generalize this.  I see two approaches we could take:

1. The simpler one to lay out: three (two?) unlabelled links, put whatever you want there.  Instead of seeing the link icon and the Twitter bird, you'd see up to three link icons with their links, plus Discord would be there.
2. The "nicer" one, but requires more layout adjustment: the Mastodon approach of some number of editable pairs, label + value, which you can populate however you want.  These won't fit in a single line like the current links do.  

I am _not_ proposing any sort of validation of these links a la Mastodon.

I'm proposing three slots because I'd like to encourage GitHub links on our production system, so that would allow you personal web site, GitHub, and some social-media site, plus Discord.

We'll need to add some explanatory text in the "Edit Profile" page for these fields, either way, so we can suggest your personal site, GitHub, or wherever else you can be contacted.

This change should include converting existing Twitter links to use one of the new slots, so no information is lost.

PR #1524, opened 2025-01-26
generalize profile links

cellio:
Fixes https://github.com/codidact/qpixel/issues/1248.

Adds user-controlled extra fields to the profile to replace the baked-in website and twitter slots.  You can put whatever you want here; proper URLs get linkified and everything else is text, like so:

Edit profile section:

editor UI

Main profile view:

profile rendering

The same link-nerfing rules that applied before apply to this new section.

Added a [maintenance task](https://github.com/Shopify/maintenance_tasks) to initialize new table rows for all current users.  After discussion, we decided this would be better than constructing new rows on demand (and creating the "add" UI in the editor).  When the task is run, all users will get three entries in user_websites, with the first two being filled with the website and twitter links if they exist.   This allows the profile editor to present a form with three label/URL pairs for editing.  For new users, we add these rows after creation.

Out of scope here (could be considered in the future):
- After this safely deploys without issues, delete the old website/twitter values after migrating.  (TBD: maintenance task to remove them or schema change to drop the columns.)  Remove from authenticated_user_object and configure_permitted_parameters as well (mentioned in chat by Oaphi).
- Make the number of websites configurable in a global site setting.  Handle the potential data loss if the limit is lowered (the main reason we're not doing this now).  My recommendation would be to preserve the data and limit the number that are shown.  (In the editor, we'd show the extras in some grayed-out-like state, so that users could still recover that data if they wanted to move it up.)
- Better integration w/third-party platforms, for example generating the icon and link like we did for twitter before this change.  (I'm concerned about maintenance burden and complexity here, but acknowledge that entering a proper link vs. just a username could be a stumbling block for some users.)

**Deployment note:** After deploying, run the following from the command line:
```
bundle exec maintenance_tasks perform Maintenance::InitializeUserWebsitesTask
```

With a large number of users we might need to do that more than once, so check the output.  (I believe the task times out after 5 minutes.)

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -359,7 +359,10 @@ class UsersController < ApplicationController
   def validated_profile_websites(profile_params)
     sites = profile_params[:user_websites_attributes]
 
-    websites = sites.transform_values { |w| w.merge({ url: ensure_protocol(w[:url]) }) }
+    websites = sites.select { |_, v| v[:url].present? &&  v[:url] != ''}
+                 .transform_values { |w| w.merge({ url: ensure_protocol(w[:url]) }) }    
+
+#    websites = sites.transform_values { |w| w.merge({ url: ensure_protocol(w[:url]) }) }    
 
     invalid_keys = websites.select { |_, w| w[:url].nil? }.keys
 
@@ -395,9 +398,9 @@ class UsersController < ApplicationController
                                                   user_websites_attributes: [:id, :label, :url])
 
     # Ensure that all user-supplied URLs are valid (strip ones that aren't).
-    if profile_params[:user_websites_attributes].present?
-      profile_params[:user_websites_attributes] = validated_profile_websites(profile_params)
-    end
+#    if profile_params[:user_websites_attributes].present?
+#      profile_params[:user_websites_attributes] = validated_profile_websites(profile_params)
+#    end
 
     @user = current_user
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -366,18 +366,6 @@ class UsersController < ApplicationController
     end
   end
 
-  # Ensure that a "naked" URL like example.com gets a protocol.
-  def ensure_protocol(uri)
-    # URI::Generic indicates the user didn't include a protocol
-    if URI.parse(uri).instance_of?(URI::Generic)
-      "https://#{uri}"
-    else
-      uri
-    end
-  rescue URI::InvalidURIError
-    nil
-  end
-
   def update_profile
     profile_params = params.require(:user).permit(:username,
                                                   :profile_markdown,
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -370,9 +370,7 @@ class UsersController < ApplicationController
 
   def validate_profile_websites(profile_params)
     sites = profile_params[:user_websites]
-    for ws in sites do
-      ensure_protocol(ws.url) || flash[:danger] = 'Invalid external link' + ws.url
-    end
+    ws.url.all? { |u| ensure_protocol(u) }
   end
 
   def ensure_protocol(uri)
@@ -391,7 +389,7 @@ class UsersController < ApplicationController
     profile_params = params.require(:user).permit(:username, :profile_markdown, :website, :discord, :user_websites)
 
     if profile_params[:user_websites].present?
-      validate_profile_websites(profile_params)
+      validate_profile_websites(profile_params) || flash[:danger] = "Invalid external link"
     end
     
     @user = current_user
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -355,28 +355,6 @@ class UsersController < ApplicationController
     render layout: 'without_sidebar'
   end
 
-  # Return the user websites that pass validation (only).
-  def validated_profile_websites(profile_params)
-    sites = profile_params[:user_websites_attributes]
-    
-    websites = sites.select { |_, v| v[:url].present? &&  v[:url] != ''}
-                 .transform_values { |w| w.merge({ url: ensure_protocol(w[:url]) }) }    
-
-#    websites = sites.transform_values { |w| w.merge({ url: ensure_protocol(w[:url]) }) }    
-
-    invalid_keys = websites.select { |_, w| w[:url].nil? }.keys
-
-    errors = invalid_keys.map do |key|
-      "\"#{sites[key][:url]}\""
-    end
-
-    unless errors.empty?
-      flash[:danger] = "Some of the external link URLs are invalid: #{errors.join(', ')}"
-    end
-
-    websites.reject { |_, w| w[:url].nil? }
-  end
-
   def cleaned_profile_websites(profile_params)
     sites = profile_params[:user_websites_attributes]
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -358,7 +358,7 @@ class UsersController < ApplicationController
   # to be deleted (when I'm done debugging with it)
   def validate_profile_website(profile_params)
     uri = profile_params[:website]
-    
+
     if URI.parse(uri).instance_of?(URI::Generic)
       # URI::Generic indicates the user didn't include a protocol, so we'll add one now so that it can be
       # parsed correctly in the view later on.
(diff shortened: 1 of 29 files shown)


app/models/user_website.rb
```diff
@@ -0,0 +1,5 @@
+class UserWebsite < ApplicationRecord
+
+  belongs_to :user
+
```
ArtOfCode-, line 4: trim: don't need the extra blank lines here

app/views/users/edit_profile.html.erb
```diff
@@ -44,6 +44,24 @@
   <p>Note: Links are not shown publicly until you have earned the Participate Everywhere ability.</p>
   <div class="post-preview"></div>
 
+  <div>
+    <p>External links (your web site, blog, social media, GitHub, etc)</p>
```
ArtOfCode-, line 48: Is it worth mentioning that only 3 will be displayed?

app/views/users/show.html.erb
```diff
@@ -57,6 +57,20 @@
...
+      <div>
+	<p><strong>External links</strong></p>
+	<table class="table is-with-hover">
+	  <% @user.user_websites.first(3).each do |w| %>
+	  <tr>
+	    <td><%= w.label %></td>
+	    <td><%= link_to w.url, w.url, rel: 'nofollow' %></td>
+	  </tr>
+	  <% end %>
+	</table>
```
ArtOfCode-, line 70: Some weird indentation here - looks like you might be using tab stops instead of spaces for indentation.

ArtOfCode- (changes requested):

app/controllers/users_controller.rb
```diff
@@ -368,14 +369,30 @@ def validate_profile_website(profile_params)
     flash[:danger] = 'Invalid profile website link.'
   end
 
-  def update_profile
-    profile_params = params.require(:user).permit(:username, :profile_markdown, :website, :twitter, :discord)
-    profile_params[:twitter] = profile_params[:twitter].delete('@')
+  def validate_profile_websites(profile_params)
+    sites = profile_params[:user_websites]
+    sites.all? { |u| u = ensure_protocol(u.url) }
```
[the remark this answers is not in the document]
cellio: Since a single update on the edit page can carry multiple websites now, my aim was to only accept the edit if all of the URLs are ok and otherwise complain.  But we also had the pre-existing code to fix missing protocols, so we both need to fix otherwise-ok URLs that need to add the protocol and (still) reject edits that aren't ok even with that.  I'm pretty confused about how to structure this.  The more I look at this code the more convinced I am that this way is wrong.  I think your map suggestion (where we strip out the bad ones and otherwise proceed) is a better idea; I'll try that.

app/views/users/edit_profile.html.erb
cellio: I think we're only going to give you three label/url textbox pairs to edit.  (Oleg suggested prefilling these rows for all users so that we don't have to figure out how to get new ones in there in the view code.)  There's no "add site" button.

app/controllers/users_controller.rb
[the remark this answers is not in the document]
ArtOfCode-: This suggestion would save _only_ the valid URLs and also complain about the invalid URLs without saving them. That would be my choice of approach, but rejecting the whole edit if any are invalid is also possible.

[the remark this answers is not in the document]
cellio: We need to ultimately save the labels and the valid websites.  We have a user_websites relation from a user, and each user_website has a position, label, and URL.  We're not letting users edit positions (move the labels/urls instead if you need to change the order), but they can change labels and urls in a single edit.

app/controllers/users_controller.rb
```diff
@@ -368,14 +369,30 @@ def validate_profile_website(profile_params)
...
+    if URI.parse(uri).instance_of?(URI::Generic)
+      # URI::Generic indicates the user didn't include a protocol, so we'll add one now so that it can be
+      # parsed correctly in the view later on.
+      return "https://#{uri}"
+    else
+      return uri
     end
+  rescue URI::InvalidURIError
+    return nil
+  end
```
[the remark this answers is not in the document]
cellio: Just checking my understanding: the `validate_profile_website` (singular) method was already here in the controller, but it should have been in the model and I should move its replacement there?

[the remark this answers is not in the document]
Oaphi: No longer applicable as we dropped validation as per discussion in Discord.

app/controllers/users_controller.rb
```diff
@@ -355,25 +355,27 @@ def edit_profile
...
+    sites = profile_params[:user_websites_attributes]
 
-    if URI.parse(uri).instance_of?(URI::Generic)
-      # URI::Generic indicates the user didn't include a protocol, so we'll add one now so that it can be
-      # parsed correctly in the view later on.
-      profile_params[:website] = "https://#{uri}"
+    sites.transform_values do |w|
+      w.merge({
+                label: w[:label].present? ? w[:label] : nil,
+                url: w[:url].present? ? w[:url] : nil
```
ArtOfCode-, line 364:
```suggestion
                label: w[:label].presence,
                url: w[:url].presence
```

ArtOfCode-: `presence` is a handy little helper that does exactly this.

app/views/users/show.html.erb
```diff
@@ -57,6 +35,36 @@
...
+          <% @user.valid_websites_for.each do |w| %>
+          <tr>
+            <td><%= w.label %></td>
+            <td>
+	      <% if w.url[0,4] == 'http' %>
+	        <%= link_to w.url, w.url, rel: 'nofollow' %>
+	      <% else %>
+	      <%= w.url %>
+	      <% end %>
+	    </td>
```
ArtOfCode-, line 52: A few more tab stops slipped past here

app/models/user.rb
```diff
@@ -61,6 +63,12 @@ def trust_level
     community_user.trust_level
   end
 
+  # Checks whether this user is the same as a given user
+  # @param [User] user user to compare with
+  def same?(user)
```
ArtOfCode-, line 68: I'd rename this `same_as?` - it reads slightly more naturally when the method is used. Not exactly a major issue, though.

test/controllers/users_controller_test.rb
```diff
@@ -115,8 +115,13 @@ class UsersControllerTest < ActionController::TestCase
...
     sign_in users(:standard_user)
-    patch :update_profile, params: { user: { profile_markdown: 'ABCDEF GHIJKL', website: 'https://example.com/user',
-                                             twitter: '@standard_user' } }
+    patch :update_profile, params: {
+      user: {
+        profile_markdown: 'ABCDEF GHIJKL',
+        website: 'https://example.com/user',
+        twitter: 'standard_user'
+      }
+    }
```
ArtOfCode-, line 124: Is this correct now that we've changed to generic links, or does this need updating to use the new format?

ArtOfCode- (changes requested):

app/models/user.rb
Oaphi: Agreed, sounds better - done

app/views/users/show.html.erb
Oaphi: Fixed those new ones to speed things up (on a tangent, we seem to have 9 other files with tab stops)

app/controllers/users_controller.rb
Oaphi: Oh, nice, that one is on me - forgot `presence` exists, done

test/controllers/users_controller_test.rb
Oaphi: No, it's definitely not correct anymore, I only fixed it up temporarily by dropping the Twitter handle validation check, the test needs to be updated (cc @cellio)

app/controllers/users_controller.rb
cellio: New to me too -- thanks for the clue!

test/controllers/users_controller_test.rb
cellio: I can remove website and twitter and add discord (which is missing).  I might need a little consultation on how to test for the new user_websites table.

cellio: Added test for the new values.

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  mount MaintenanceTasks::Engine, at: "/maintenance_tasks"
```
cellio, line 2: We need to restrict access to this path to admins.  @Oaphi do you have any idea how to do that?  (Pinging you since you suggested the package, so hoping you know more about it.)

ArtOfCode- (approved):

PR #1525, opened 2025-01-26
Followup fix for cache & global site setttings handling

Oaphi:
This PR follows up #1517 and #1516 ensuring that:

- global site settings clear cache for _all_ communities, not just the current one;
- adds proper methods for SiteSettingsController's actions (`access?`, `audit_update`, `clear_cache`, `do_create` methods);
- adds `global?` method on the SiteSetting model;
- fixes overflow of site setting values when the `<details>` element is opened (it's a crude fix for now via `overflow-wrap` - we should make HTML-valued settings use code blocks instead);

Site setting value overflow before the fix:

Screenshot from 2025-01-26 22-13-56

And lack thereof after:

Screenshot from 2025-01-26 22-23-42

### DIFF
--- a/app/assets/stylesheets/site_settings.scss
+++ b/app/assets/stylesheets/site_settings.scss
@@ -1,4 +1,5 @@
 .site-setting--value {
   min-height: 1em;
   min-width: 2em;
+  overflow-wrap: anywhere;
 }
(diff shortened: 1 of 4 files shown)


cellio (commented):
<s>I'm not sure if this is expected behavior, but I just tested this and here's what I see:

- On community A, edited the setting.  New value showed up immediately.
- On community B, old value shows up.
- On community B, opened the setting and it shows the new value (from A).
- If I then click in and then click update, it does.

So it seems like the *community* knows about the new value (it's shown in the settings), but the display on the main page doesn't.  Is there something else I needed to do before doing this test?</s>

Never mind - pilot error.  (Needed to clear the bad cache state first.)

cellio:
LGTM - tested and it works. I defer to Art for code review.

ArtOfCode- (approved):

PR #1526, opened 2025-01-29
Support only Ruby 3.1 onwards

trichoplax:
Currently we test Ruby 2.7 and Ruby 3.1. Ruby 3.1 will reach end of life on 2025-03-31, and Ruby 2.7 has already passed end of life.

This pull request stops testing Ruby 2.7 and starts testing Ruby 3.2. The intention is that we will continue to test the oldest minor version that is still supported (has not reached end of life) plus the minor version after that, so that the tests ensure we are always ready to stop using a version when it reaches end of life.

For example, on 2025-04-01 when Ruby 3.1 is no longer supported, we can switch from testing 3.1 and 3.2 to testing 3.2 and 3.3.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,7 +198,7 @@ jobs:
             bundle exec rails test
       - store_test_results:
           path: "~/qpixel/test/reports"
-  system-test-ruby31:
+  system-test-ruby32:
     docker:
       - image: cimg/ruby:3.2-browsers
       - image: cimg/mysql:8.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,137 +1,5 @@
 version: 2.1
 jobs:
-  test-ruby27:
-    docker:
-      - image: cimg/ruby:2.7-node
-      - image: cimg/mysql:8.0
-        command: [--default-authentication-plugin=mysql_native_password]
-        environment:
-          MYSQL_ROOT_HOST: '%'
-          MYSQL_ROOT_PASSWORD: 'root'
-          MYSQL_DATABASE: 'qpixel_test'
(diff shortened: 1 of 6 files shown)


Oaphi:
TODO list:

- [ ] update documentation to mention that we switched to 3.1 / 3.2 (or, better yet, add badges);
- [ ] update Docker configuration to use Ruby 3.2;

Oaphi:
> Would the badges be 2 test outcome badges, one for 3.1 and one for 3.2?

Don't have a strong opinion on that, but 2 badges, one for each target, sounds reasonable to me

> One possibility I was imagining was keeping this at 3.1 until a second pull request is merged on 2025-04-01

d'oh, I haven't paid attention to the PR's description... Well, this item is mostly to ensure I don't forget to do so as promised, we can merge without it if I end up not being available to make good on that promise in a timely fashion.

trichoplax:
Having separate badges for results of testing Ruby 3.1 and 3.2 seems impossible with CircleCI without ongoing manual workarounds with separate branches, so I've left just the single badge for now. Splitting into 2 might be worth considering once we move away from CircleCI.

Previously the Ruby versions supported were mentioned only in `INSTALLATION.md`, and not below the badge in `README.md`. Unless anyone needs them to also be mentioned below the badge, I'll leave it this way to reduce the number of places that need updating each year when we move to the next minimum supported version.

If you're happy to leave the Docker configuration using 3.1 until we update the minimum supported version to 3.2 on 2025-04-01, is this pull request ready to merge now?

Oaphi:
LGTM - I've pushed a minor update to the example output when starting Rails (to reflect the new Ruby version).

ArtOfCode- (approved):

PR #1527, opened 2025-02-06
alert mods if a user has past warnings/suspensions

cellio:
Small mod assist: on the user page, if there have been warning or suspension messages, show the number and how recent to mods.  This information is already available if you go into the mod tools, but some mods have found it helpful to have a heads-up that there's something to see.

Totally open to better ways to present this!

Suspended user (to show interaction with suspension notice):

screenshot: notice appears after profile text/buttons and just before posts

Non-suspended user (only warnings or inactive suspensions here):

screenshot

### DIFF
--- a/app/models/community_user.rb
+++ b/app/models/community_user.rb
@@ -30,6 +30,10 @@ class CommunityUser < ApplicationRecord
     false
   end
 
+  def latest_warning
+    mod_warnings&.order(created_at: 'desc').first.created_at
+  end
+
   # Calculation functions for privilege scores
   # These are quite expensive, so we'll cache them for a while
   def post_score
--- a/app/models/community_user.rb
+++ b/app/models/community_user.rb
@@ -31,7 +31,7 @@ class CommunityUser < ApplicationRecord
   end
 
   def latest_warning
-    mod_warnings&.order(created_at: 'desc').first.created_at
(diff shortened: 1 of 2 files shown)


ArtOfCode-:
I'm not totally sold that it needs to be this prominent - the majority of the time you visit a user profile it's probably _not_ for mod activity. What if we just tacked a little counter into the mod tools button instead?

image

cellio:
> I'm not totally sold that it needs to be this prominent - the majority of the time you visit a user profile it's probably _not_ for mod activity. What if we just tacked a little counter into the mod tools button instead?

That's a good idea!  With timestamp of the most recent in a tooltip if we can do that on that button, maybe?

ArtOfCode-:
Yep, we can do that

cellio:
I put the count in the button (if there's anything to say), and the date in the menu item itself.  However, to reduce clutter, if the user is *currently* suspended, I just show the "includes lifting suspension" part that was already there without jamming in a date too.  If there's no active suspension, most recent contact is shown.  I figure if the user is suspended, you already know there are issues and will take a look if you need to.

Suspended user:
Screenshot

Contacted not-suspended user:
Screenshot

Unremarkable user:
Screenshot

ArtOfCode- (approved):

PR #1530, opened 2025-02-14
Update maintenance_tasks

ArtOfCode-:
Updates the maintenance_tasks gem. Also restores the routes and protects them to developer access only, with an integration test to confirm.


Oaphi (approved):
LGTM, tested out locally

PR #1548, opened 2025-03-03
Remove special case notifying author of threads

### LINKED ISSUE #1025: On your own posts, show that you're following new threads (don't show "follow new" when you already are)

trichoplax:
This pull request continues from #1037 that was from MoshiKoi's fork. It's now a branch on the QPixel repo to allow any QPixel developer to work on it. See the discussion on that previous pull request.

Fixes #1025.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -41,12 +41,9 @@ class CommentsController < ApplicationController
 
     if success
       notification = "New comment thread on #{@comment.root.title}: #{@comment_thread.title}"
-      unless @comment.post.user == current_user
-        @comment.post.user.create_notification(notification, helpers.comment_link(@comment))
-      end
 
       ThreadFollower.where(post: @post).each do |tf|
-        unless tf.user == current_user || tf.user == @comment.post.user
+        unless tf.user == current_user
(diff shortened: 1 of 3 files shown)


trichoplax:
Note that there is a suggested migration approach in a [comment on the original version of this pull request](https://github.com/codidact/qpixel/pull/1037#issuecomment-1807167400), which will need to be added to this new pull request before merging. I haven't implemented this migration yet because it needs to come after the migration in #1920 so I'll wait until any changes requested during its review to ensure this one is dated later than that one.

Issue #1540, opened 2025-02-24
All communities: link posts, reputation


cellio:
meta:292842

The "all communities" tab on the user profile shows the number of posts and the reputation on each community.  This request is to make those links to the "all posts" page and the "vote summary" page on the target profile, respectively.

PR #1549, opened 2025-03-05
linkify entries for posts, reputation on "all communities" page

cellio:
Fixes https://github.com/codidact/qpixel/issues/1540.

### DIFF
--- a/app/views/users/_network.html.erb
+++ b/app/views/users/_network.html.erb
@@ -9,16 +9,18 @@
     <% if @user.has_profile_on(c) %>
     <tr>
       <td><%= user_link @user, { host: c.host}, anchortext: c.name %>
-	<% if @user.is_moderator_on(c) %>
-	(moderator)
-	<% end %>
+        <% if @user.is_moderator_on(c) %>
+          (moderator)
+        <% end %>
       </td>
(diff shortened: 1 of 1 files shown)


app/views/users/_network.html.erb
```diff
@@ -13,8 +13,12 @@
...
 	<% end %>
       </td>
-      <td><%= @user.post_count_on(c) %> </td>
-      <td><%= @user.reputation_on(c) %> </td>
+      <td><%= link_to user_posts_url(@user, host: c.host) do %>	
+	<%= @user.post_count_on(c) %>
+	<% end %> </td>
+      <td><%= link_to vote_summary_url(@user, host: c.host) do %>
+	<%= @user.reputation_on(c) %>
+	<% end %> </td>
```
ArtOfCode-, line 21: Looks like there's some tab stops in here, which is making the indentation funny.

ArtOfCode- (changes requested):

app/views/users/_network.html.erb
cellio: Whoops.

ArtOfCode- (approved):

PR #1558, opened 2025-03-09
Add `overflow-wrap`s to several elements currently overflowing

### LINKED ISSUE #802: Odd horizontal scrollbar on paragraphs with math rendering
See [this answer](https://math.codidact.com/posts/286166/286174#answer-286174) on the Math community:

a horizontal scrollbar appears

A horizontal scrollbar appears instead of it being word-wrapped.

What differentiates this from other [long equations with horizontal scroll](https://math.codidact.com/posts/286292) is that the equations themselves aren't especially wide; instead, it is an issue of them being improperly wrapped.

### LINKED ISSUE #1249: Code block in a <summary> overflows the post bounds
meta:288974

If someone uses a code block as the summary text in a `<summary>`/`<details>` block, the text can overflow the bounds of the post body:

screenshot

See the linked post for more examples.

We don't need to support such uses of these HTML tags, but we should prevent the overflow.  I don't care if we use horizontal scrolling, truncate with "...", or something else I haven't thought of, but let's enforce the boundary somehow.

### LINKED ISSUE #1478: Unbreakable lines in comments can overflow their bounds
meta:292198

A comment with a very long line in code markup (which doesn't line-wrap) causes the line to overflow the bounds of the comment box.  I imagine the same would happen with a very long "word" even not in code markup, like SomeReallyLongAndAnnoyingJavaExceptionNameException, which can happen.

I don't know what the solution is -- scrollbar on comments (like on code blocks)?  Something else?

LunarWatcher:
Various instances of fixed overflow:
* Tables, particularly the user activity table when the edit summary overflows. The fix here is `overflow-wrap: anywhere;`, because `overflow-wrap: break-word` did nothing - not sure if this has any unintended consequences, but it doesn't appear to have. 
* Titles (post listing, question page)
* Body (post listing, post preview)
* Tag wikis (summary, body)
* Revisions, specifically the diff
* ... and possibly more places affected by the modified/added CSS selectors

Comments, which were the initial problem with #1478, appear to have been fixed. Since no source was linked on the meta post, I couldn't see the original, but a quick test setup with inline code showed that it did indeed wrap within width under 306284d0bf671eff8d2d9dcda0548705a5a3dc52. Comment content for the test was just a long series of letters without spaces wrapped in an inline codeblock, because that was the text that overflowed.

No title was used for this comment, however, so the auto-generated title _did_ overflow. That has also been fixed, so comment thread titles now also wrap properly.

Various unhandled edge-cases:
* The "about me" currently handles it by adding horizontal scrollbars. Didn't touch this for now, what to do about it (do what everything else does, or accept it on account of the "about me" being special) needs to be handled separately 
* Though not described in any current issues, zalgo does overflow into other elements. It's pretty easy to fix in pure CSS (`overflow: clip`), but it requires adding it to quite a few more CSS classes, and just adding it everywhere feels unnecessary

On that note, I'm not sure if it would be better to remove some of the added `overflow-wrap`s and throw then in a separate CSS class (`has-overflow-break-word` or something), but this would require the classes[^1] added to a _lot_ of elements for it to work properly.

---

Closes #802
Closes #1249
Closes #1478

[^1]: `break-word` is the most common use-case, but `anywhere` has a few applications as well - though nowhere near as many. Some also need an explicit width for `break-word` to be properly respected, which may or may not be  defined already for any given element.


Oaphi:
To check (if somebody gets to any of these items before I do, feel free to mark as complete):

- [ ] Tables, particularly the user activity table;
- [ ] Titles (post listing, question page);
- [ ] Body (post listing, post preview);
- [ ] Tag wikis (summary, body);
- [ ] Revisions, specifically the diff;

Oaphi:
@LunarWatcher fixed another instance where we have an overflow:
Screenshot from 2025-03-10 11-44-42

I've added a `wrap-anywhere` utility CSS class, but don't take it as wanting to use it for the rest of the elements: as you know, I hate "utility classes" with passion. We'll make a proper CSS component for such titles at a later point (hopefully)

Oaphi:
Note to self (I think we have it already logged, but just in case) - tags with very long names can be created but cannot be used: 
Screenshot from 2025-03-10 11-59-12

trichoplax:
> To check (if somebody gets to any of these items before I do, feel free to mark as complete)

Should we include #1249 in this issue as part of this list of things to check?

cellio:
> Should we include #1249 in this issue as part of this list of things to check?

If it's easy to do here, let's.  If we encounter any complications, I think it's ok to defer because we already have an issue tracking that (so it won't be lost).

On the long-tag-name case, a clip with ellipsis seems better than wrapping (doesn't have to be here), *but*: if we properly validate the length, does the case arise?  How long can a tag name be without causing problems (remember smaller tablets)?  Tag length can be configured (35 is the default, not baked in), but if any reasonable length would work, then we could fix the bug that allows long names to be created and not worry about how to wrap too-long tags.  At least until somebody sees a use case for 100-character tags or something...

LunarWatcher:
@Oaphi:
> To check (if somebody gets to any of these items before I do, feel free to mark as complete): 

For the sake of bookkeeping, there's also comment thread titles, but it wasn't a separate bullet because reasons 

@trichoplax:
> Should we include #1249 in this issue as part of this list of things to check?

I'll check when I get around to it - did look for other issues, but didn't find that one. There's also #802, but I've been unable to reproduce it, so it might be fixed already

cellio:
> There's also #802, but I've been unable to reproduce it, so it might be fixed already

I see the bug in the Math post linked in the issue.  From discussion there, it seems to show up at some viewport sizes/zoom levels and not others.

LunarWatcher:
Oh, one of those bugs. I tried to reproduce it at several zoom levels and screen sizes, but couldn't. I'll try again later with more pixel-perfect resizing then

cellio:
> #802

FWIW, my browser reports that it is currently 1160px wide and at 100% zoom I see the bug on that Math answer (Brave, Chrome).  I haven't done more exhaustive testing, but that's a datum, anyway.

LunarWatcher:
d8322f4 probably needs a bit more manual testing than bbacd33, since the latter is limited to mathjax elements in any context. I don't _think_ this should have unintended consequences, but I don't have that many test cases in my dev instance yet. Checked with the obvious troublemaker (tables), and it rendered fine:

Screenshot showing a table with 5 columns and one non-header row, where each cell just consists of a long string of random characters, with no table overflow

LunarWatcher:
... bbacd33 might've broken long block equations. Checking now

cellio:
> Update: It did not, I checked out `develop`, and it's still borked. The test post is https://math.codidact.com/posts/286292, and the live version renders correctly, but not on develop. Weirdly, the master branch is also borked on my machine, so maybe there's custom CSS on prod?

I can repro -- live Math post uses horizontal scroll and does not overflow; a copy of that same post in my dev environment sometimers overflows and sometimes uses h-scroll.  In this screenshot I'm seeing different behavior for consecutive math blocks; don't know yet if that's related:

Screenshot

LunarWatcher:
@cellio See update 2 in that comment, I figured out why it's being weird

cellio:
Oh good; I was working on assembling the list of mathjax for each of four equations of interest, but looks like I don't need to.

cellio:
Where does this PR stand?  Are we waiting for review, or are reviewers waiting for changes?

LunarWatcher:
@cellio Awaiting either review, or confirmation of a performed review. No changes have been requested yet, unless I missed something on Discord or something

ArtOfCode- (approved):

Oaphi (approved):
@LunarWatcher I've added one more fix while we are at it (post history, initial revision) - with a scroll on overflow (preserves your fixes + ensures diffs display reasonably well when we can't wrap). CircleCI went into maintenance, so let's wait until it's up again before merging (just in case), and once it is - LGTM to me too.

Before:
Screenshot from 2025-03-28 02-40-27

After:
Screenshot from 2025-03-28 02-38-42

P.S. Screenshots do not start from the same position, the width should not change

Oaphi:
And one other small fix for tags to ensure they don't overflow the container even in the worst cases.

Before:
Screenshot from 2025-03-28 02-58-43

After:
Screenshot from 2025-03-28 03-00-08

Oaphi:
Last one, I swear: tags container in the posts list (once again - not a flaw with the PR, another extra thing I noticed).

Before:
Screenshot from 2025-03-28 03-06-52

After:
Screenshot from 2025-03-28 03-06-55

Oaphi:
There's one other place we might want to address, but I am out of time to spare, unfortunately (and what's in the PR is certainly much more than enough as it is): tag merge titles:

image

cellio:
> There's one other place we might want to address, but I am out of time to spare, unfortunately (and what's in the PR is certainly much more than enough as it is) - tag merge titles:

That can wait.  But what happened to the tests?  Is that the maintenance you mentioned?  Will we need to manually kick them off later?

Oaphi:
> > There's one other place we might want to address, but I am out of time to spare, unfortunately (and what's in the PR is certainly much more than enough as it is) - tag merge titles:
> 
> That can wait. But what happened to the tests? Is that the maintenance you mentioned? Will we need to manually kick them off later?

Yup, all workflows are failing right now due to CircleCI's maintenance - if there will be no new commits after they wrap up, we will have to rerun the jobs manually (or they'll run automatically upon merge)

PR #1571, opened 2025-03-24
Fix for seeds failing on clean test database

Oaphi:


Oaphi:
I also don't think `Rails.cache.delete 'network/post_types/rep_changes'` does anything as we only cache `network/category_post_types/rep_changes` but not sure investigating this is in scope for the PR.

trichoplax:
I've raised a separate issue so we don't forget to check if anything needs fixing later.

PR #1552, opened 2025-03-07
Fix Nokogiri dependency issue due to bump to 1.8.3

Oaphi:
Dependabot PR responsible for the issue: #1532

Relevant change happened in [Nokogiri 1.8.0](https://github.com/sparklemotion/nokogiri/blob/main/CHANGELOG.md#v1180--2024-12-25). Should fix #1551

Issue #1551, opened 2025-03-06
docker based installation fails: missing nokogiri gem


edgimar:
**Describe the bug**
When running `docker compose build` followed by `docker compose up`, the following error messages are printed to the console:

```
uwsgi-1  | /usr/local/bundle/gems/bundler-2.4.13/lib/bundler/definition.rb:524:in `materialize': Could not find nokogiri-1.18.3-x86_64-linux in locally installed gems (Bundler::GemNotFound)
uwsgi-1  | 	from /usr/local/bundle/gems/bundler-2.4.13/lib/bundler/definition.rb:197:in `specs'
uwsgi-1  | 	from /usr/local/bundle/gems/bundler-2.4.13/lib/bundler/definition.rb:254:in `specs_for'
uwsgi-1  | 	from /usr/local/bundle/gems/bundler-2.4.13/lib/bundler/runtime.rb:18:in `setup'
uwsgi-1  | 	from /usr/local/bundle/gems/bundler-2.4.13/lib/bundler.rb:162:in `setup'
uwsgi-1  | 	from /usr/local/bundle/gems/bundler-2.4.13/lib/bundler/setup.rb:23:in `block in <top (required)>'
uwsgi-1  | 	from /usr/local/bundle/gems/bundler-2.4.13/lib/bundler/ui/shell.rb:159:in `with_level'
uwsgi-1  | 	from /usr/local/bundle/gems/bundler-2.4.13/lib/bundler/ui/shell.rb:111:in `silence'
uwsgi-1  | 	from /usr/local/bundle/gems/bundler-2.4.13/lib/bundler/setup.rb:23:in `<top (required)>'
uwsgi-1  | 	from <internal:/usr/local/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
uwsgi-1  | 	from <internal:/usr/local/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
uwsgi-1  | 	from /code/config/boot.rb:3:in `<top (required)>'
uwsgi-1  | 	from /code/bin/rails:8:in `require_relative'
uwsgi-1  | 	from /code/bin/rails:8:in `<top (required)>'
uwsgi-1  | 	from /usr/local/bundle/gems/spring-4.0.0/lib/spring/client/rails.rb:30:in `load'
[...]

PR #1573, opened 2025-03-26
cleaning up after dependabot's breaking of nokogiri

Oaphi:
This is basically a copy of #1552 because the issue resurfaced due to #1564 switching generic `nokogiri` back to `x86_64-linux-gnu` (which is incorrect and breaks Docker setup, see #1551 for details).

Related issues (I am not sure what we need to do to prevent Dependabot from doing this yet):

- https://github.com/dependabot/dependabot-core/issues/3299;
- https://github.com/dependabot/dependabot-core/issues/5917;

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -203,6 +203,7 @@ GEM
     mime-types-data (3.2022.0105)
     mini_magick (4.11.0)
     mini_mime (1.1.5)
+    mini_portile2 (2.8.8)
     minitest (5.16.3)
     minitest-ci (3.4.0)
       minitest (>= 5.0.6)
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
I think it is possible to prevent Dependabot from updating a specific dependency, but nokogiri is one of our high-risk ones that has fairly frequent issues and fixes, so I'd rather not - we might just have to keep an eye out for nokogiri PRs and update them manually.

ArtOfCode- (approved):

Oaphi:
> I think it is possible to prevent Dependabot from updating a specific dependency, but nokogiri is one of our high-risk ones that has fairly frequent issues and fixes, so I'd rather not - we might just have to keep an eye out for nokogiri PRs and update them manually.

Yeah, my first thought was to add nokogiri to the ignore list, but it isn't really viable. I *think* it can be solved by specifying all platforms and all overrides as deepndencies instead of generic `ruby` and generic nokogiri, but the reports indicate that there's also a bundler problem, so... idk, need to tesr things out, and doing so is quite an annoyance 😔

Issue #1577, opened 2025-03-30
Misleading false failure message when trying to post a comment on an answer to your own post


trichoplax:
meta:293281

When a user without the "Participate Everywhere" ability tries to add a comment to an answer to their own question (which they should be able to do), they receive an error page (not an HTML page, just a raw JSON error page) saying:

> status: "failed"
> message: "New users can only comment on their own posts and on answers to them."

> ![JSON error summary showing instead of a web page](https://meta.codidact.com/uploads/wbu8ygvtbf49l8i6411i2vzaky0v)

Note that this screenshot is not the developer window, this is what is shown directly to the user.

In addition to causing confusion because the user should be able to comment, there is further confusion because the comment does save successfully, but the user is unaware of this due to the misleading error page.

This has since also been raised independently by a different user on Software Development Codidact's Meta: [I comment on an answer on my question, and get this error: "New users can only comment on their own posts and on answers to them" ](https://software.codidact.com/posts/293642)

cellio:
Commenting for Participate seems to be more generally broken: https://github.com/codidact/qpixel/issues/1538 .  :-(

PR #1578, opened 2025-03-30
New user comments

### LINKED ISSUE #1538: User without Participate Everywhere can comment freely
The Participate ability allows you to comment on your own posts and answers to your own questions, but not more broadly (according to the documentation).  Participate Everywhere allows commenting everywhere.  However, this restriction isn't working -- a newly-created user with only Participate was able to comment on an existing comment thread, and I just tested in a dev environment and found that such a user can also create new comment threads on other users' posts.  It seems we are not enforcing this restriction, and spammers are able to exploit it.  (I found out it wasn't working because of a spam flag.)

ArtOfCode-:
Update the comment rate limiting code so the logic is easier to follow. Adds tests for the rate limit helper to ensure it's returning the right thing. Also changes from a `render json` to a `redirect` with flash so that the raw JSON error message isn't displayed.

I haven't been able to run and test this myself, so I'm relying on the automated testing. If anyone is able to test manually when reviewing that would be useful.

Closes #1538, #1577

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -32,7 +32,12 @@ class CommentsController < ApplicationController
 
     pings = check_for_pings @comment_thread, body
 
-    return if comment_rate_limited
+    rate_limited, limit_message = helpers.comment_rate_limited?(current_user, @post)
+    if rate_limited
+      flash[:danger] = limit_message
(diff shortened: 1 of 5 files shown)


Oaphi:
Tested manually locally - LGTM

Oaphi (approved):

PR #1581, opened 2025-03-31
Fix OpenGraph tags

### LINKED ISSUE #1391: OpenGraph HTML meta tags are broken or missing
I copied an answer of mine from SO to SD-CD, so I checked the HTML sources of the webpages on both websites to compare them, and found several issues with the CD page, some of which concern OpenGraph.

- [Software Development Codidact Q/A](https://software.codidact.com/posts/292128)
- [Stack Overflow Q/A](https://stackoverflow.com/questions/33902080/avoid-a-class-to-be-extend-anonymously-in-java)

The problems I found in the HTML source for CD's page:

- `<meta property="og:type" content="website"/>` is missing
- `<meta property="og:image" .../>`'s URL is broken
- `<meta property="og:site_name" .../>`  is missing

ArtOfCode-:
Closes #1391.

Adds missing `og` tags and fixes the broken `og:image` tags.

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -67,5 +67,5 @@
   <meta property="og:url" content="<%= request.original_url %>" />
   <meta property="og:title" content="<%= page_title %>" />
   <meta property="og:description" content="<%= content_for?(:description) ? content_for(:description) : desc %>" />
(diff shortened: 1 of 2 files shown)


cellio (approved):
LGTM.

PR #1582, opened 2025-03-31
Null-safe usernames

ArtOfCode-:
Found a few more instances of `rtl_safe_username` being used directly instead of the safer helper - this switches them over.


cellio (approved):
LGTM (but I am not an expert here). I see the code-coverage complaint but am not concerned about it.  We already didn't have coverage here; we don't need to block a fix for a potentially common server error on that.

I'll approve and you can decide whether we need more eyes on it before merging.

PR #1583, opened 2025-03-31
Use numeric inputs for two-factor codes

### LINKED ISSUE #1575: Two-factor code entry field does not use numeric keypad
meta:293614

I thought it did, but shows what I know!

We should set attributes on the code entry field so that it triggers the numeric keypad on mobile devices.

ArtOfCode-:
Closes #1575.

Changes the two-factor code inputs to autocomplete OTC, autofocus, numeric input mode, and input type "number". This should trigger mobile devices to use numeric keypads.


Oaphi (approved):
LGTM

PR #1584, opened 2025-03-31
Remove "create new comment thread" if it is not possible to add a comment in the first place (+ huge refactor of access control-related methods)

### LINKED ISSUE #1579: Indicate to users that can't create new comment threads on a given post they can't do so before they spend time writing it up
This is a followup issue based on the rate limiting fix in #1578. Currently, we only indicate that the user cannot comment (or create a new thread) only after submission - which means they can go through the whole flow and only learn that they failed after spending time writing the comment body & title.

https://software.codidact.com/posts/293642

Oaphi:
closes #1579

~~Also introduces a couple of QoL methods on the User model: `is_standard` for checking if the current user is neither an admin nor a moderator, as well as its inverse `is_privileged` (we should rename all of the `is_` checks to follow the `?` convention, but this is outside the scope of this PR).~~

Also nearly fully refactors access control methods to disambiguate their usage & reduce duplication (such as "is moderator or admin" and related checks).

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -240,7 +240,7 @@ class CommentsController < ApplicationController
     when 'delete'
       return not_found unless current_user.privilege?('flag_curate') && @comment_thread.deleted?
 
-      if @comment_thread.deleted_by.is_moderator && !current_user.is_moderator
+      if @comment_thread.deleted_by.at_least_moderator? && !current_user.at_least_moderator?
         render json: { status: 'error',
                        message: 'Threads deleted by a moderator can only be undeleted by a moderator.' }
         return
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -258,7 +258,7 @@ class CommentsController < ApplicationController
 
   def post
     @post = Post.find(params[:post_id])
-    @comment_threads = if helpers.at_least_moderator? || current_user&.has_post_privilege?('flag_curate', @post)
+    @comment_threads = if current_user&.at_least_moderator? || current_user&.has_post_privilege?('flag_curate', @post)
                          CommentThread
                        else
                          CommentThread.undeleted
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -240,7 +240,7 @@ class CommentsController < ApplicationController
     when 'delete'
       return not_found unless current_user.privilege?('flag_curate') && @comment_thread.deleted?
 
-      if @comment_thread.deleted_by.at_least_moderator? && !current_user.at_least_moderator?
+      if @comment_thread.deleted_by.is_moderator && !current_user.is_moderator
         render json: { status: 'error',
                        message: 'Threads deleted by a moderator can only be undeleted by a moderator.' }
         return
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -187,7 +187,7 @@ class CommentsController < ApplicationController
   end
 
   def thread_rename
-    if @comment_thread.read_only? && !current_user.is_moderator
+    if @comment_thread.read_only? && !current_user.at_least_moderator?
       flash[:danger] = 'This thread has been locked.'
       redirect_to comment_thread_path(@comment_thread.id)
       return
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -240,7 +240,7 @@ class CommentsController < ApplicationController
     when 'delete'
       return not_found unless current_user.privilege?('flag_curate') && @comment_thread.deleted?
 
-      if @comment_thread.deleted_by.is_moderator && !current_user.at_least_moderator?
+      if @comment_thread.deleted_by.at_least_moderator? && !current_user.at_least_moderator?
         render json: { status: 'error',
                        message: 'Threads deleted by a moderator can only be undeleted by a moderator.' }
         return
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -258,7 +258,7 @@ class CommentsController < ApplicationController
 
   def post
     @post = Post.find(params[:post_id])
-    @comment_threads = if helpers.moderator? || current_user&.has_post_privilege?('flag_curate', @post)
+    @comment_threads = if helpers.at_least_moderator? || current_user&.has_post_privilege?('flag_curate', @post)
                          CommentThread
                        else
                          CommentThread.undeleted
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -240,7 +240,7 @@ class CommentsController < ApplicationController
     when 'delete'
       return not_found unless current_user.privilege?('flag_curate') && @comment_thread.deleted?
 
-      if @comment_thread.deleted_by.is_moderator && !current_user.is_moderator
+      if @comment_thread.deleted_by.is_moderator && !current_user.at_least_moderator?
         render json: { status: 'error',
                        message: 'Threads deleted by a moderator can only be undeleted by a moderator.' }
         return
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -535,16 +535,16 @@ class PostsController < ApplicationController
   end
 
   def lock
-    return not_found unless current_user.privilege? 'flag_curate'
+    return not_found unless current_user&.privilege? 'flag_curate'
     return not_found if @post.locked?
 
     length = params[:length].present? ? params[:length].to_i : nil
     if length
-      if !current_user.is_moderator && length > 30
+      if !current_user&.at_least_moderator? && length > 30
         length = 30
       end
       end_date = length.days.from_now
-    elsif current_user.is_moderator
+    elsif current_user&.at_least_moderator?
       end_date = nil
     else
       end_date = 7.days.from_now
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -438,7 +438,7 @@ class PostsController < ApplicationController
       return
     end
 
-    if @post.deleted_by.is_moderator && !current_user&.at_least_moderator?
+    if @post.deleted_by.at_least_moderator? && !current_user&.at_least_moderator?
       flash[:danger] = helpers.i18ns('posts.cant_restore_deleted_by_moderator')
       redirect_to post_path(@post)
       return
@@ -559,7 +559,7 @@ class PostsController < ApplicationController
     return not_found(errors: ['no_privilege']) unless current_user&.privilege? 'flag_curate'
     return not_found(errors: ['not_locked']) unless @post.locked?
 
-    if @post.locked_by.is_moderator && !current_user&.at_least_moderator?
+    if @post.locked_by.at_least_moderator? && !current_user&.at_least_moderator?
       return not_found(errors: ['locked_by_mod'])
     end
 
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -569,7 +569,7 @@ class PostsController < ApplicationController
   end
 
   def feature
-    return not_found(errors: ['no_privilege']) unless current_user.is_moderator
+    return not_found(errors: ['no_privilege']) unless current_user&.at_least_moderator?
 
     data = {
       label: @post.parent.nil? ? @post.title : @post.parent.title,
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -673,7 +673,7 @@ class PostsController < ApplicationController
       redirect_back fallback_location: root_path
     end
 
-    if !@post_type.is_public_editable && !(@post.user == current_user || current_user.is_moderator)
+    if !@post_type.is_public_editable && !(@post.user == current_user || current_user&.at_least_moderator?)
       flash[:danger] = helpers.i18ns('posts.not_public_editable')
       redirect_back fallback_location: root_path
     end
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -387,13 +387,13 @@ class PostsController < ApplicationController
       return
     end
 
-    if @post.post_type.is_freely_editable && !current_user&.is_moderator
+    if @post.post_type.is_freely_editable && !current_user&.at_least_moderator?
       flash[:danger] = helpers.i18ns('posts.cant_delete_community')
       redirect_to post_path(@post)
       return
     end
 
-    if @post.children.any? { |a| !a.deleted? && a.score >= 0.5 } && !current_user&.is_moderator
+    if @post.children.any? { |a| !a.deleted? && a.score >= 0.5 } && !current_user&.at_least_moderator?
       flash[:danger] = helpers.i18ns('posts.cant_delete_responded')
       redirect_to post_path(@post)
       return
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -556,9 +556,10 @@ class PostsController < ApplicationController
   end
 
   def unlock
-    return not_found(errors: ['no_privilege']) unless current_user.privilege? 'flag_curate'
+    return not_found(errors: ['no_privilege']) unless current_user&.privilege? 'flag_curate'
     return not_found(errors: ['not_locked']) unless @post.locked?
-    if @post.locked_by.is_moderator && !current_user.is_moderator
+
+    if @post.locked_by.is_moderator && !current_user&.at_least_moderator?
       return not_found(errors: ['locked_by_mod'])
     end
 
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -438,7 +438,7 @@ class PostsController < ApplicationController
       return
     end
 
-    if @post.deleted_by.is_moderator && !current_user.is_moderator
+    if @post.deleted_by.is_moderator && !current_user&.at_least_moderator?
       flash[:danger] = helpers.i18ns('posts.cant_restore_deleted_by_moderator')
       redirect_to post_path(@post)
       return
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -10,7 +10,7 @@ class ReactionsController < ApplicationController
     comment = nil
 
     if !comment_text.blank?
-      if @post.comments_disabled && !current_user.is_moderator && !current_user.is_admin
+      if @post.comments_disabled && !current_user.at_least_moderator?
         render json: { status: 'failed', message: 'Comments have been disabled on this post.' }, status: :forbidden
         return
       end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -51,7 +51,16 @@ class UsersController < ApplicationController
         redirect_to user_path(@user)
       end
       format.json do
-        data = [:id, :username, :is_moderator, :is_admin, :is_global_moderator, :is_global_admin, :trust_level,
+        data = [:id,
+                :username,
+                :at_least_moderator?,
+                :at_least_global_moderator?,
+                :admin?,
+                :moderator?,
+                :standard?,
+                :is_global_moderator,
+                :is_global_admin,
+                :trust_level,
                 :se_acct_id].to_h { |a| [a, @user.send(a)] }
         render json: data
       end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -654,7 +654,7 @@ class UsersController < ApplicationController
   end
 
   def user_scope
-    if helpers.moderator?
+    if helpers.at_least_moderator?
       User.all
     else
       User.active
@@ -662,7 +662,7 @@ class UsersController < ApplicationController
   end
 
   def check_deleted
-    if (@user.deleted? || @user.community_user.deleted?) && (!helpers.moderator? || params[:deleted_screen].present?)
+    if (@user.deleted? || @user.community_user.deleted?) && (!helpers.at_least_moderator? || params[:deleted_screen].present?)
       render :deleted_user, layout: 'without_sidebar', status: 404
     end
   end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -663,7 +663,7 @@ class UsersController < ApplicationController
   end
 
   def user_scope
-    if helpers.at_least_moderator?
+    if current_user&.at_least_moderator?
       User.all
     else
       User.active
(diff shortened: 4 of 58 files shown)


Oaphi:
> Are we also intercepting attempts to leave a new comment in an existing thread?

Actually, we already do, we were only missing a similar check in the inline view:

Screenshot from 2025-04-01 03-21-46
Screenshot from 2025-04-01 03-21-29

app/models/user.rb
```diff
@@ -151,14 +151,34 @@ def ensure_websites
...
+  # Checks if the user is neither a moderator not an admin (global or on the current community)
+  # @return [Boolean] check result
+  def is_standard
+    !is_moderator && !is_admin
+  end
+
+  # Checks if the user is either a moderator or an admin (global or on the current community).
+  # It's the inverse of the is_standard check
+  # @return [Boolean] check result
+  def is_privileged
```
ArtOfCode-, line 178: Given that we use "privilege" a lot elsewhere in the code to talk about abilities, I wonder if it would be better to rename this something like `is_superuser`

ArtOfCode- (approved):
One comment for consideration but otherwise LGTM

trichoplax (changes requested):
Various places in the code have had redundancy due to the somewhat misleadingly named `is_moderator` actually meaning "is a moderator or an admin" (since admins can do everything moderators can do). Over time I've tried to remove that redundancy by removing `|| is_admin` from places I've worked on (since it's already covered by `is_moderator`), but I can now see that it is still potentially confusing so perhaps we should make the naming more clear.

I've added comments to specific places where I'd expect confusion to occur in future.

I don't know the best way to rename the functions but I think it's worth a discussion before committing to something new. The `is_moderator` function effectively means "has moderator powers" but I'm not sure that's a better name.

app/models/user.rb
```diff
@@ -151,14 +151,34 @@ def ensure_websites
...
+  # Checks if the user is either a global admin or an admin on the current community
+  # @return [Boolean] check result
   def is_admin
     is_global_admin || community_user&.is_admin || false
   end
 
+  # Checks if the user is neither a moderator not an admin (global or on the current community)
+  # @return [Boolean] check result
+  def is_standard
+    !is_moderator && !is_admin
```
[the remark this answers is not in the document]
Oaphi: Agreed on renaming, that definitely needs to happen as `is_moderator` doesn't actually mean what it's name implies (well, more like doesn't _only_ mean that). That said, I don't think "just the negation of another function" is a problem - it's a pattern Ruby/Rails (see, for example, [`present?`](https://apidock.com/rails/v7.1.3.2/Object/present%3F), [`disjoint?`](https://apidock.com/ruby/v2_5_5/Set/disjoint%3F)) + it's easier to extend than explicit negation (plus it fits with `unless` better).

app/models/user.rb
Oaphi: Not applicable anymore due to extensive refactoring of access control methods

app/models/user.rb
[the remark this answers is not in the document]
Oaphi: Keeping it unresolved for now until we settle on the exact name

app/controllers/users_controller.rb
```diff
@@ -654,15 +663,18 @@ def set_user
   end
 
   def user_scope
-    if helpers.moderator?
+    if helpers.at_least_moderator?
```
ArtOfCode-, line 666: Same comment re helpers

ArtOfCode- (approved):
This is great work, thank you! LGTM with a couple of minor comments for consideration.

cellio:
This is approved (with some suggestions still pending), but per discussion in chat, do not merge yet.  (The author would like to participate in the merge.)

app/controllers/comments_controller.rb
```diff
@@ -258,7 +258,7 @@ def thread_unrestrict
 
   def post
     @post = Post.find(params[:post_id])
-    @comment_threads = if helpers.moderator? || current_user&.has_post_privilege?('flag_curate', @post)
+    @comment_threads = if helpers.at_least_moderator? || current_user&.has_post_privilege?('flag_curate', @post)
```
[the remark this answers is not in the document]
Oaphi: Agreed, good idea - this one's done

app/controllers/users_controller.rb
Oaphi: This one is done as well - seems safe enough to switch

app/models/user.rb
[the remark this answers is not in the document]
Oaphi: Resolving this as it doesn't seem like there's been any objection to using `at_least_moderator?`, `admin?`, and `moderator?` so far

Oaphi:
Should be good to go now, I believe I've addressed everything I wanted to

app/controllers/comments_controller.rb
```diff
@@ -187,7 +187,7 @@ def thread_followers
   end
 
   def thread_rename
-    if @comment_thread.read_only? && !current_user.is_moderator
+    if @comment_thread.read_only? && !current_user.at_least_moderator?
```
trichoplax, line 190: Untested suggestion:

```suggestion
    if @comment_thread.read_only? && current_user.standard?
```

app/controllers/comments_controller.rb
```diff
@@ -240,7 +240,7 @@ def thread_unrestrict
     when 'delete'
       return not_found unless current_user.privilege?('flag_curate') && @comment_thread.deleted?
 
-      if @comment_thread.deleted_by.is_moderator && !current_user.is_moderator
+      if @comment_thread.deleted_by.at_least_moderator? && !current_user.at_least_moderator?
```
trichoplax, line 243: Untested suggestion:

```suggestion
      if @comment_thread.deleted_by.at_least_moderator? && current_user.standard?
```

app/controllers/posts_controller.rb
```diff
@@ -387,13 +387,13 @@ def delete
       return
     end
 
-    if @post.post_type.is_freely_editable && !current_user&.is_moderator
+    if @post.post_type.is_freely_editable && !current_user&.at_least_moderator?
```
trichoplax, line 390: Untested suggestion:

```suggestion
    if @post.post_type.is_freely_editable && current_user&.standard?
```

app/controllers/posts_controller.rb
```diff
@@ -387,13 +387,13 @@ def delete
...
       flash[:danger] = helpers.i18ns('posts.cant_delete_community')
       redirect_to post_path(@post)
       return
     end
 
-    if @post.children.any? { |a| !a.deleted? && a.score >= 0.5 } && !current_user&.is_moderator
+    if @post.children.any? { |a| !a.deleted? && a.score >= 0.5 } && !current_user&.at_least_moderator?
```
trichoplax, line 396: Untested suggestion:

```suggestion
    if @post.children.any? { |a| !a.deleted? && a.score >= 0.5 } && current_user&.standard?
```

app/controllers/posts_controller.rb
```diff
@@ -438,7 +438,7 @@ def restore
       return
     end
 
-    if @post.deleted_by.is_moderator && !current_user.is_moderator
+    if @post.deleted_by.at_least_moderator? && !current_user&.at_least_moderator?
```
trichoplax, line 441: Untested suggestion:

```suggestion
    if @post.deleted_by.at_least_moderator? && current_user&.standard?
```

app/controllers/posts_controller.rb
```diff
@@ -535,16 +535,16 @@ def toggle_comments
...
 
   def lock
-    return not_found unless current_user.privilege? 'flag_curate'
+    return not_found unless current_user&.privilege? 'flag_curate'
     return not_found if @post.locked?
 
     length = params[:length].present? ? params[:length].to_i : nil
     if length
-      if !current_user.is_moderator && length > 30
+      if !current_user&.at_least_moderator? && length > 30
```
trichoplax, line 543: Untested suggestion:

```suggestion
      if current_user&.standard? && length > 30
```

app/controllers/posts_controller.rb
```diff
@@ -568,7 +569,7 @@ def unlock
   end
 
   def feature
-    return not_found(errors: ['no_privilege']) unless current_user.is_moderator
+    return not_found(errors: ['no_privilege']) unless current_user&.at_least_moderator?
```
trichoplax, line 572: Untested suggestion:

```suggestion
    return not_found(errors: ['no_privilege']) if current_user&.standard?
```

(This one seems very much personal taste - suggesting in case it's of interest.)

app/controllers/posts_controller.rb
```diff
@@ -672,7 +673,7 @@ def edit_checks
       redirect_back fallback_location: root_path
     end
 
-    if !@post_type.is_public_editable && !(@post.user == current_user || current_user.is_moderator)
+    if !@post_type.is_public_editable && !(@post.user == current_user || current_user&.at_least_moderator?)
```
trichoplax, line 676: Untested suggestion:

```suggestion
    if !@post_type.is_public_editable && !@post.user == current_user && current_user&.standard?
```

Without testing, I think this change follows from [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws), but which is more readable is probably personal taste. Personally I find the suggestion easier to parse quickly because it doesn't require parentheses for operator precedence.

app/controllers/reactions_controller.rb
```diff
@@ -10,7 +10,7 @@ def add
     comment = nil
 
     if !comment_text.blank?
-      if @post.comments_disabled && !current_user.is_moderator && !current_user.is_admin
+      if @post.comments_disabled && !current_user.at_least_moderator?
```
trichoplax, line 13: Untested suggestion:

```suggestion
      if @post.comments_disabled && current_user.standard?
```

app/controllers/users_controller.rb
```diff
@@ -654,15 +662,18 @@ def set_user
...
       User.all
     else
       User.active
     end.joins(:community_user).includes(:community_user, :avatar_attachment)
   end
 
   def check_deleted
-    if (@user.deleted? || @user.community_user.deleted?) && (!helpers.moderator? || params[:deleted_screen].present?)
+    deleted = @user.deleted? || @user.community_user.deleted?
+    go_to_not_found = !current_user&.at_least_moderator? || params[:deleted_screen].present?
```
trichoplax, line 674: Untested suggestion:

```suggestion
    go_to_not_found = current_user&.standard? || params[:deleted_screen].present?
```

app/helpers/users_helper.rb
```diff
@@ -113,7 +113,7 @@ def rtl_safe_username(user)
   def user_link(user, url_opts = {}, **link_opts)
     anchortext = link_opts[:anchortext]
     link_opts_reduced = { dir: 'ltr' }.merge(link_opts).except(:anchortext)
-    if user.nil? || (deleted_user?(user) && !moderator?)
+    if user.nil? || (deleted_user?(user) && !at_least_moderator?)
```
trichoplax, line 116: Untested suggestion:

```suggestion
    if user.nil? || (deleted_user?(user) && standard?)
```

app/models/community_user.rb
```diff
@@ -152,10 +155,28 @@ def trust_level
     attributes['trust_level'] || recalc_trust_level
   end
 
+  # Checks if the community user is an admin (global or on the current community)
+  # @return [Boolean] check result
+  def admin?
+    is_admin || user&.is_global_admin || false
```
trichoplax, line 161: Untested suggestion:

```suggestion
    is_admin || user&.global_admin? || false
```

app/models/community_user.rb
```diff
@@ -152,10 +155,28 @@ def trust_level
...
+  end
+
+  # Checks if the community user is a moderator (global or on the current community)
+  # @return [Boolean] check result
+  def moderator?
+    is_moderator || user&.is_global_moderator || false
```
trichoplax, line 167: Untested suggestion:

```suggestion
    is_moderator || user&.global_moderator? || false
```

app/models/post.rb
```diff
@@ -73,7 +73,7 @@ def self.by_slug(slug, user)
       return nil
     end
 
-    if post&.help_category == '$Moderator' && !user&.is_moderator
+    if post&.help_category == '$Moderator' && !user&.at_least_moderator?
```
trichoplax, line 76: Untested suggestion:

```suggestion
    if post&.help_category == '$Moderator' && user&.standard?
```

app/views/comments/thread.html.erb
```diff
@@ -78,7 +78,7 @@
   <% end %>
   <div role="list">
     <% comments.each do |comment| %>
-      <% if comment.deleted && !(current_user&.is_moderator && params[:show_deleted_comments] == "1") %>
+      <% if comment.deleted && !(current_user&.at_least_moderator? && params[:show_deleted_comments] == "1") %>
```
trichoplax, line 81: Untested suggestion:

```suggestion
      <% if comment.deleted && (current_user&.standard? || params[:show_deleted_comments] != "1") %>
```

I really can't decide which is more readable, so this is just a suggestion for comparison.

trichoplax (commented):
As someone who has worked on some of the affected code in the past and struggled, this feels much more comfortable to read now. Thank you!

There are a few places where `user.standard?` has been used to replace `!user.at_least_moderator?`. For consistency, would it make sense to use this in more places? I've added a few suggested places, but please note I don't have a dev environment at present so I haven't tested anything.

test/controllers/flags_controller_test.rb
```diff
@@ -68,6 +68,18 @@ class FlagsControllerTest < ActionController::TestCase
     assert_response(404)
   end
 
+  test 'should not allow non-moderator users to resolve flags on themselves' do
```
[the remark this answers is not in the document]
trichoplax: Following [discussion in Discord](https://discord.com/channels/634104110131445811/668890340199235613/1364795715527901216), this isn't decided yet so no further test is needed in this pull request.

trichoplax (approved):
Marking approved to avoid holding this up. Note that I've only read the code, not being able to test locally at present.

app/models/community_user.rb
Oaphi: Same as with `admin?`, done

app/models/post.rb
Oaphi: Unresolving this - reverted for now, for some reason I am not sure yet, it causes a test failure despite the `standard?` method being just a negation of `at_least_moderator?`

Oaphi: Let's not apply these ones as the actual switch is `!user || user.standard?` which is not what you hoped for, I believe

trichoplax: Oh I see. Well spotted. Yes that's more work to read than the original. There are a few suggested changes that don't involve the safe navigation operator, which I'm guessing would still work? I don't know what's best for consistency though.

test/controllers/comments_controller_test.rb
```diff
@@ -37,6 +37,32 @@ class CommentsControllerTest < ActionController::TestCase
...
+    threads = JSON.parse(response.body)
+    assert_equal threads.any? { |t| t['deleted'] }, true
+
+    sign_in users(:moderator)
+    get :post, params: { post_id: posts(:question_one).id }, format: :json
+    threads = JSON.parse(response.body)
+    assert_equal threads.any? { |t| t['deleted'] }, true
+  end
+
+  test 'users should see deleted threads on their own posts even' do
```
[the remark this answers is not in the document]
Oaphi: Yup, looks like I forgot to add the "if those threads are deleted" part that was supposed to explain what the test is for in the first place 😅

app/controllers/posts_controller.rb
Oaphi: As per our discussion, let's leave this one be as well, as `!current_user&.at_least_moderator?` is actually equivalent to `!current_user || current_user.standard?` (safe navigation operator)

app/controllers/posts_controller.rb
Oaphi: Same thing here - resolving as we decided to avoid `!current_user || current_user.standard?`, which is needed due to the safe navigation operator.

app/controllers/posts_controller.rb
Oaphi: Same as with other inversions of `at_least_moderator?` + safe navigation operator + negation - not as simple as inverting to `standard?`

app/controllers/reactions_controller.rb
Oaphi: This might be one that's fine as there is no safe navigation - need to check whether that's not an oversight in the first place, though.

app/views/comments/thread.html.erb
Oaphi: This one is a bit trickier, but, unfortunately, suffers from the same issue of negation + safe navigation + `at_least_moderator` not being equivalent to `&.standard?` ( due to `!(A & B)` -> `(!A | !B)` )

app/helpers/users_helper.rb
Oaphi: As I mentioned in another comment, this one seems safe to me due to `user.nil? || ...` - let's see

app/controllers/posts_controller.rb
Oaphi: Resolving due to `!user&.at_least_moderator?` and `user&.standard?` not being equal

app/models/user.rb
```diff
@@ -151,12 +151,49 @@ def ensure_websites
...
+  # Checks if the user is a global moderator (ensures consistent return type & naming scheme)
+  # @return [Boolean] check result
+  def global_moderator?
+    is_global_moderator || false
+  end
+
+  # Checks if the user is either a global admin or an admin on the current community
+  # @return [Boolean] check result
+  def admin?
+    global_admin? || community_user&.is_admin || false
```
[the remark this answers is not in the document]
Oaphi: This looks fine to me (putting aside the issue I have the check itself being a ticking bomb for a circular dependency between the `User` and `CommunityUser` models just waiting to happen), let's check.

app/views/posts/_help_center_posts.html.erb
```diff
@@ -1,6 +1,6 @@
 <% posts.to_a.in_groups_of(3).map(&:compact).each do |row| %>
   <% row.each do |row_data| %>
-    <% next if row_data[0] == '$Moderator' && !current_user&.is_moderator %>
+    <% next if row_data[0] == '$Moderator' && !current_user&.at_least_moderator? %>
```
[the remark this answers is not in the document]
Oaphi: Resolving this too as, unfrotunately, `!current_user&.at_least_moderator?` is not the same as `current_user&.standard?` without an additional `user.nil?`/`!user` check.

app/controllers/comments_controller.rb
Oaphi: I believe `current_user` is safe to switch, but let's avoid relying on its implementation too much just in case.

app/controllers/comments_controller.rb
Oaphi: Same here as with the other checks relying on `current_user` implementation - while it might be safe, let's not invite issues should that change.

app/views/users/_common_card.html.erb
```diff
@@ -9,7 +9,7 @@
 <% ckb ||= false %>
 <% small ||= false %>
 
-<% if user.nil? || (deleted_user?(user) && !moderator?) %>
+<% if user.nil? || (deleted_user?(user) && !at_least_moderator?) %>
```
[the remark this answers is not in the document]
Oaphi: This one should be safe to apply due to `user.nil? || ...`

app/controllers/posts_controller.rb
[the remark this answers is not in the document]
Oaphi: Given that `deleted_by` refers to a user model via association, it should use the current status, not the status at the time of deletion. I don't think that's desired, though, we should probably store the status (something along the lines of `deleted_by_moderator`) as well. That said, agreed that it doesn't seem in scope for the PR - let's make a separate issue, maybe?

app/views/posts/_post_tools.html.erb
```diff
@@ -5,7 +5,7 @@
       Moderator Tools
     </div>
     <div class="modal--body">
-      <% unless post.locked? && !moderator? %>
+      <% unless post.locked? && !at_least_moderator? %>
```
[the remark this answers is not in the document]
Oaphi: > It's unclear to me why there was already a moderator check inside the Moderator Tools modal

Defensive programming, I suppose - never hurts to make sure in case the partial gets accidentally added to an unprotected context.

As for the change, I opted to keep it as is - the helper relies on `user_signed_in?` for the equivalence to work - I don't want to touch that unless necessary (at least for now).

app/controllers/reactions_controller.rb
Oaphi: As I mentioned in my other comments, I've decided to not touch the `current_user` ones for now even though I think it should be safe from the looks of it. We should probably revisit those at some point, though.

app/views/posts/_expanded.html.erb
```diff
@@ -501,7 +501,7 @@
             <button class="toc--header" data-toggle="#toc-toggle" data-toggle-property="class" data-toggle-value="is-active">Table of Contents</button>
             <% sorted_answers = post.children.sort_by { |answer| answer.score }.reverse! %>
             <% sorted_answers.each do |answer| %>
-              <% next if answer.deleted? && !moderator? %>
+              <% next if answer.deleted? && !at_least_moderator? %>
```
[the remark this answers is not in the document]
Oaphi: Same as with other helpers relying on `user_signed_in` for the equivalency to work - decided to not do that for now (we should address that at some point, though).

app/views/comments/thread.html.erb
```diff
@@ -101,18 +101,18 @@
...
-          <% end %>
+          <%= render 'shared/lock_notice', level: 'info', text: 'Comments are disabled on deleted posts.' %>
+        <% elsif @post.comments_disabled && standard? %>
+          <%= render 'shared/lock_notice', text: 'Comments have been disabled on the post.' %>
+        <% elsif @post.locked? %>
+          <%= render 'shared/lock_notice', text: 'Comments are disabled on locked posts, but as a moderator you are exempt from that block.' %>
+        <% elsif @post.comments_disabled %>
+            <%= render 'shared/lock_notice', text: 'Comments have been disabled on this post, but as a moderator you are exempt from that block.' %>
+        <% end %>
+        <% if @post.comments_allowed? || !standard? %>
```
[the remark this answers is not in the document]
Oaphi: Same here - decided not to apply `at_least_moderator?` -> `!standard?` inversions to avoid locking in to the `user_signed_in?`'s internal implementation.

app/controllers/posts_controller.rb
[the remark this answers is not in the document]
trichoplax: Yes definitely not looking to add it to this pull request - I just wanted a second opinion before raising it. Thanks for confirming. I've now raised #1607.

PR #1585, opened 2025-03-31
Link to deleted users for moderators

### LINKED ISSUE #1416: How viewable should deleted users be by moderators?
**Describe the bug**
The user profile page of a deleted user shows no information to a standard user (as expected), but to a moderator shows the profile information accompanied by a message:

> This profile is deleted. You have access to this page as a moderator, but this is not public information.

Moderators therefore have access to this information, but no obvious way to find the page. When a user is deleted, links to the user profile page are removed for standard users (as expected), but they are also removed for moderators. For example, all of the following places do not let a moderator navigate to the deleted user's profile page:
- The list of users (deleted users do not show up, with no option for a moderator to view them).
- The "Suggested by" section of a suggested edit (the user shows as "deleted user" and the link goes back to the same page rather than the profile page).
- The username of a comment (the user shows as "deleted user" and the link goes back to the same page).
- The user card below a post (this shows as "deleted user" with no link).

This means that a moderator can find out which user was the author in all of these cases, but only by an inconvenient process of changing URLs. Would it be better to be consistent, and do one of the following:
- Deny the moderator the information altogether (by making the profile page unviewable).
- Make it easy for the moderator to see information that they are allowed to see, by making the links to the profile still work for moderators after deletion.

What access to deleted user information should each of the following have:
- User
- Curator
- Moderator
- Admin
- Developer

ArtOfCode-:
Closes #1416.

Makes links to deleted users work for moderators. User cards retain the 'deleted' style but will show the sanitised username and profile image as normal.

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -113,7 +113,7 @@ module UsersHelper
   def user_link(user, url_opts = {}, **link_opts)
     anchortext = link_opts[:anchortext]
     link_opts_reduced = { dir: 'ltr' }.merge(link_opts).except(:anchortext)
-    if deleted_user?(user) || user.nil?
+    if user.nil? || (deleted_user?(user) && !moderator?)
       link_to 'deleted user', '#', link_opts_reduced
     elsif !anchortext.nil?
       link_to anchortext, user_url(user, **url_opts), { dir: 'ltr' }.merge(link_opts)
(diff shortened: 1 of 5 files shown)


ArtOfCode-:
Not concerned about the CI failures here - the system test is a flaky test that passes when re-run and the coverage is previously uncovered code.

Oaphi:
Apologies for a bit of a mess, I was trying to figure out what's causing the coverage to stubbornly refuse to increase in this PR. Turns out it's a combination of environment & our Spring usage (see #1588 for reference)

Oaphi (approved):
LGTM otherwise, checked locally. Merging in - want to give the deploy a shove as well

PR #1587, opened 2025-03-31
fixed max-width override taking over droppanel's max-width in posts

### LINKED ISSUE #1586: Recent fix for wrapping issues clashed with drop panel styling resulting in its width balooning to the root element's width
meta:293652

Oaphi:
Closes #1586

Thank god for specificity (also damn it as it's also the main reason behind the issue in the first place)

### DIFF
--- a/app/assets/stylesheets/posts.scss
+++ b/app/assets/stylesheets/posts.scss
@@ -48,6 +48,7 @@ h1 .badge.is-tag.is-master-tag {
 .post-list--title {
   font-size: 19px;
   width: 100%;
+
   a {
     display: inline-block;
     width: 100%;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/posts.scss
```diff
@@ -217,10 +219,16 @@ h1 .badge.is-tag.is-master-tag {
   font-weight: lighter;
 }
 
+.post--actions {
```
cellio, line 222: Are these classes already in use and we only needed to add this CSS property?  (I expected to find an invocation in the PR.)

Oaphi: Yup, nothing new, we just don't have any styling overrides for `post--actions` (well, _now_ we do)

cellio (approved):

PR #1588, opened 2025-04-01
RAILS_ENV should be set to test not only during prep when running CircleCI jobs

Oaphi:
While we set up our tests with `RAILS_ENV` set to `true`, we run the actual tests without the env set. This causes Spring to be used and, what's more important, for SimpleCov to be initialized after it is run. This makes our Coverage reports incorrect (see, for example, lack of coverage increase for helpers and a significant bump in coverage in this exact PR).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,8 @@ jobs:
             git rev-parse $(git rev-parse --abbrev-ref HEAD)
       - run:
           name: Test
+          environment:
+            RAILS_ENV: test
           command: |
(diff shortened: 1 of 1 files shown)


ArtOfCode- (approved):

cellio (approved):
LGTM with my limited knowledge of this area of tooling.

Oaphi:
O.K., let's go with this for now - if we'll need Spring back (which wasn't intentional in the first place), we can enable it later

Issue #1591, opened 2025-04-01
Post content doesn't save properly


ArtOfCode-:
We've had two separate reports of this now, same issue both times: meta:293657 and meta:293622.

I've checked the logs and the `__html` attribute that should contain the HTML rendered by our JS was sent with no content in it. I haven't been able to work out why.

One workaround may be to check on the server side that the generated HTML is longer than the Markdown sent (I can't think of any cases where the Markdown would be longer than the HTML), and render the Markdown server-side if it's not.

cellio:
Markdown is pretty compact; I can't think of a case where it would be longer than the HTML either.  We do want to find the root cause, but that sounds like a reasonable workaround in the meantime.  (Let's be sure to leave a code comment so future-us doesn't wonder why we're doing this extra check and remove it...)

trichoplax:
Would a check that the HTML is not empty be sufficient - were both cases completely empty?

trichoplax:
Unusual but possible case where Markdown could be longer than its resulting HTML:

Markdown (231 characters):

```text
Fairly Long Title | Measurement Before (with notes) | Measurement After (with notes)
----------------- | ------------------------------- | ------------------------------
Content           | Content                         | Content
```

Resulting HTML (223 characters):

```html
<thead>
<tr>
<th>Fairly Long Title</th>
<th>Measurement Before (with notes)</th>
<th>Measurement After (with notes)</th>
</tr>
</thead>
<tbody>
<tr>
<td>Content</td>
<td>Content</td>
<td>Content</td>
</tr>
</tbody>
</table>
```

This only happens because the `--- | --- | ---` separating table headings from table body can optionally be extended to make the Markdown readable as a table even when not rendered as HTML. This means arbitrarily many Markdown characters can be added without adding anything to the HTML.

Not everyone aligns their tables nicely in Markdown like this (it only benefits future editors, not readers), so some look more like this:

Markdown (124 characters):

```text
Fairly Long Title | Measurement Before (with notes) | Measurement After (with notes)
---|---|---
Content | Content | Content
```

However, I've seen at least some aligned Markdown tables (and I sometimes do this myself).

For a single row table like this, the HTML adds 99 characters which is easily matched by adding hyphens and spaces to align the columns in the Markdown. Adding more rows increases the 99 characters cost of converting to HTML, but also introduces more opportunity for extra spaces in the extra rows in the Markdown, so a fairly natural looking table with arbitrarily many rows could easily have longer Markdown than HTML.

[...]

PR #1592, opened 2025-04-01
Workaround for #1591: expand server-side rendering

ArtOfCode-:
See #1591.

Checks if the client side has actually rendered any HTML and not just sent an empty string, and renders server-side if it hasn't.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -46,7 +46,7 @@ class PostsController < ApplicationController
                     @parent.category
                   end
                 end || nil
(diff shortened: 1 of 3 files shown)


trichoplax (approved):
Works well. On `develop` adding a post with just a single image (so the body field is never directly interacted with) results in an empty saved post. On this branch, it results in a correctly displayed image. While I don't know how to reproduce the specific browser problem causing this to happen for longer posts, this seems very likely to mitigate it.

I've added a couple of comments about potential wording changes, but I'm uncertain what's best so please view these as optional - I'm approving either way.

PR #1593, opened 2025-04-01
Fix numeric stats column wrapping in user sidebar

### LINKED ISSUE #1590: Visual bug: user profile stats tables are broken by overflow wrapping
Image

This is caused by the `overflow-wrap: anywhere` rule on `td` elements. I don't know what that rule was added to address so I'm slightly wary of just removing it. However, there's a test post with wide-content table cells [here](https://a.dev.codidact.org/posts/706/707#answer-707) and removing the rule doesn't seem to break that table.

Is anyone able to shed more light on what that rule was for, or suggest any alternative fixes? I wondered briefly about `overflow: scroll` instead of wrapping, but again without knowing what the rule was for it's hard to tell if that would work.

Oaphi:
Closes #1590 

I opted for a combination of using:
- fixed table layout;
- disabled wrapping for the stats column (always the right one);
- clipping the content of the stats column on overflow (with ellipsis);
- showing a toolitp with the full value on hover;

The above should be enough to address the issue and it also has an extra benefit of improving layout alignment across the sidebar:

Screenshot from 2025-04-02 01-03-17

The PR also comes with a bunch of other minor improvements (general view cleanup, several methods & helpers moving logic where it belongs - to models & helpers)

### DIFF
--- a/app/assets/stylesheets/tables.scss
+++ b/app/assets/stylesheets/tables.scss
@@ -1,0 +1,5 @@
+.table {
+  &.fixed {
+    table-layout: fixed;
+  }
+}
--- a/app/assets/stylesheets/utilities.scss
+++ b/app/assets/stylesheets/utilities.scss
@@ -345,6 +345,12 @@ span.spoiler {
   overflow: hidden;
 }
 
+.overflow-ellipsis {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
 .transparent {
   opacity: 0;
 }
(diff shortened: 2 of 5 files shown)


cellio (approved):
LGTM.  Thanks for moving SQL out of the view and into the model.

cellio:
Can we get a review from someone who's more code-fluent than I am?  It's not large and it would be nice to either get this fixed or identify anything that still needs to be done here.  Thanks!

ArtOfCode- (approved):

PR #1595, opened 2025-04-05
Add additional post history events

### LINKED ISSUE #475: Post history doesn't include category changes
The comments and history on this post seems confusing to those who don't know about the missing step: https://software.codidact.com/posts/281308

When a post is moved to a different category, could we log that in the history?  Right now the history implies that a Codidact admin overrode the community by reopening a question, but it was closed on Q&A and reopened on Meta after being moved (and "for reasons" the move required an admin assist).

ArtOfCode-:
Closes #475.

Adds additional post history events for locks, unlocks, and category changes.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -555,7 +555,7 @@ class PostsController < ApplicationController
     end
 
     ApplicationRecord.transaction do
-      @post.update locked: true, locked_by: current_user,
+      @post.update! locked: true, locked_by: current_user,
                    locked_at: DateTime.now, locked_until: end_date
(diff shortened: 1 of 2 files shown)


ArtOfCode-:
Not 100% why rubocop is failing here - it's passing locally and the error it's picking up doesn't exist locally.

cellio:
What's the status of this PR?  I see the test failure and the out-of-date notice; would addressing the latter possibly address the former?

ArtOfCode-:
Just needs review - other than that good to go

Oaphi:
> What's the status of this PR? I see the test failure and the out-of-date notice; would addressing the latter possibly address the former?

The failure seems to be our good old flaky system tests, so everything else should be good to go (I'll review a bit later - updated the branch for now)

PR #1598, opened 2025-04-09
Fix .comment-meta width issue

LunarWatcher:
#1558 introduced a [`max-width` condition](https://github.com/codidact/qpixel/pull/1558/files#diff-ab4ea0c322b2990cbb02c30bdc9f8739c62a39c617b8aa33647aa36c8e05200aR224-R226) that, for reasons I don't understand and likely never will, caused comment boxes to start underflowing:

Screenshot showing a comment with the meta header not taking up the full available width

Fix verified in:
* `/thread/{id}`
    * Selected thread + when embedded on that page
* `/posts/{id}`

This PR fixes this bug. AFAIK, it has not been reported anywhere (yet).

### DIFF
--- a/app/assets/stylesheets/comments.scss
+++ b/app/assets/stylesheets/comments.scss
@@ -74,6 +74,10 @@
 
 .post--comments-container {
   margin-bottom: 1rem;
+  .comment-meta {
+      min-width: 100%;
+      box-sizing: content-box;
+  }
 }
 
 .post--comments-thread.is-inline {


LunarWatcher:
Is the test failure one of the flaky tests addressed in #1569? I don't see how this tiny CSS change would cause that to fail otherwise

Oaphi:
> Is the test failure one of the flaky tests addressed in #1569? I don't see how this tiny CSS change would cause that to fail on one single ruby version otherwise

Yeah, judging by the error message, it's the flaky system tests, I'll rerun the pipeline shortly

Oaphi:
Oh, I just realized I've noticed the issue too and fixed it in my PR with fixes related to the button: https://github.com/codidact/qpixel/pull/1584/commits/2e7578c0c503d45dc7c8a8dbb8ef79eebe821b2a. Not sure how to proceed, though

LunarWatcher:
Maybe it's worth reconsidering the problematic rule instead of doing this fix? I don't think the entire subtree needs a max-width, so maybe
```diff
-   * {
+   > * {
      max-width: 100%;
    }
```
Or something like that? I have not tested this rule, nor have I checked if that backfires, but I'm pretty sure the problem it was trying to fix was immediate descendants of the post, and not the entire subtree. Might've been an either inline mathjax or mathjax block fix as well, but that could be done separately

ArtOfCode-:
Superseded by #1584

PR #1600, opened 2025-04-13
Add TS type check action

ArtOfCode-:
Adds a GH action to perform JS type checking based on TypeScript types.

We still write pure JS, this just adds a layer of type safety.

### DIFF
--- a/.github/workflows/typescript.yml
+++ b/.github/workflows/typescript.yml
@@ -1,0 +1,26 @@
+name: TypeScript check
+
+on:
+  push:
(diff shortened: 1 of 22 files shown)


ArtOfCode-:
Before anyone asks, that was _meant_ to fail...

Oaphi:
@ArtOfCode- done, fixed the remaining type check errors, passing on the reigns back to you

ArtOfCode-:
@Oaphi thanks for the assist!

PR #1606, opened 2025-04-24
user search should match any substring, not just initial substring

cellio:
Made the `LIKE` behavior for user search align with that for tag search, so you can match a substring from the middle of the name and not just the initial substring.  One-character fix.


ArtOfCode- (approved):

Oaphi (approved):
LGTM, just added a quick test to increase coverage

Issue #1609, opened 2025-04-30
Automated tests for the community assets


trichoplax:
**Is your feature request related to a problem? Please describe.**
Recently the automated leaderboard for the Code Golf community developed a bug - [Missing usernames on leaderboard](https://codegolf.codidact.com/posts/293883). The JavaScript that produces this leaderboard can be found in [public/assets/community](https://github.com/codidact/qpixel/tree/develop/public/assets/community). There do not appear to be any automated tests for the code in this folder. It is outside the `app` folder and it is not clear to me where such tests would belong.

**Describe the solution you'd like**
Could we have tests for the community-specific code in this folder? Where should such tests live? Will it be possible to include them in the CI so that we are automatically alerted of failures?

What approach would also work for other instances of QPixel, that do not have Code Golf and Judaism, but may have their own custom CSS and JS for some of their communities?

Oaphi:
On a related note, I'd move our community assets to separate repositories and start publishing them as packages - I don't think they belong to the "core" for a couple of reasons:

- QPixel is supposed to be an engine anyone can use without being a part of our network;
- it tightly couples them to our deploy cycle, meaning that any, even trivial change to those has to wait till we can deploy;

That said (and part of the reason why I am bringing it up here), such a change would make it harder to add them to our CI pipeline. It doesn't make it _impossible_, though, we'll just have to be a bit more creative with setting it up.

trichoplax:
Sounds good. Having a separate repository per community (and presumably environment variables pointing to them from QPixel?) would mean we could keep the tests for each community in that community's repo, so the tests for QPixel are only those relevant to other instances. Then a community is free to write many/long/inefficient tests without having any effect on QPixel development.

I have no idea how to pull QPixel in for running the community repo tests, but if it's possible I'd like to see that at some point.

cellio:
I'm not sure it needs a repo per community.  A way to separate our network's add-ons from the core code base would seem to meet the need and be less complex?  (In other words, think instance, not community within instance.  This might also be easier for other instances to emulate.)

Oaphi:
Splitting between "core" and "community" is also na option, yes - I don't have a strong preference either way as long as we keep a clear distinction between thhe engine and the stuff that is instance-specific. Having a repo for *all* comminity-related assets doesn't affect our ability publish them as packages as well (although it's a bit more complex to set up) - it can be a monorepo with multiple packages (it's common practice as well, see, for example, DefinitelyTyped that provides typings for various npm packages)

PR #1610, opened 2025-05-01
Fix missing usernames on Code Golf leaderboard

trichoplax:
Fixes [Missing usernames on leaderboard](https://codegolf.codidact.com/posts/293883) which I've only just posted - it doesn't have a GitHub issue yet.

We recently made the user avatar in the user card under a post into a link to the user profile page. This means there are now 2 elements with the class `.user-card--link` per user card, and the leaderboard code was trying to find a username in the avatar instead of the text link.

This pull request makes the selector more specific to avoid this problem.

I'm not on my development machine so I've only tested this in the browser developer window on the live site, not in development. Making a pull request now to remind me to test this when I can, and to allow anyone else to test if you wish.

### DIFF
--- a/public/assets/community/codegolf.js
+++ b/public/assets/community/codegolf.js
@@ -3,6 +3,22 @@
  *  License: AGPLv3
  */
 
+/**
+ * @typedef {{
+ *   answerID: string
+ *   answerURL?: string
+ *   page: number
+ *   username: string
+ *   userid: string
+ *   full_language?: string
+ *   language?: string
+ *   variant?: string
+ *   extensions?: string
+ *   code?: string
+ *   score?: string
+ * }} ChallengeEntry
+ */
+
 (() => {
   const dom_parser = new DOMParser();
   const match = location.pathname.match(/(?<=posts\/)\d+/);
@@ -59,6 +75,9 @@
     }
   };
 
+  /**
+   * @param {string} id
+   */
   async function getLeaderboard(id) {
     const response = await fetch(`/posts/${id}`);
     const text = await response.text();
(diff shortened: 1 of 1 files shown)


Oaphi (approved):
LGTM, tested locally and made a couple of improvements on top:

1. Added some JSDoc for type checking in TS-aware IDEs;
2. Ensured `getLeaderboard` doesn't cause a runtime error if `header` is not found in some of the entries;
3. Added a bit of padding to the right of the now restrored usernames;

cellio:
I apologize.  I made the change that caused this and it never *occurred* to me that there was a dependency in the leaderboard.

Oaphi:
@cellio that's fine, the issue is more that we neither provide a hook there for scripts to use (in other places, we mark elements that are used by our JS assets with classes prefixed with `js-*` [which is common practice]) nor do we have an API _yet_ to avoid coupling to specific markup structure.

trichoplax:
@cellio echoing @Oaphi there's nothing to apologise for. Every change is seen by at least 2 people (commiter and reviewer) and I wouldn't expect either of them to spot something that no amount of manual testing can find without going through every existing community (please don't start doing that!).

I've raised #1609 to discuss ways of automatically testing the communities that have custom JS. On top of that, as pointed out, once we have an API this sensitivity of the leaderboard to HTML changes will go away.

And if you insist on apologising, then I'm sorry too - I both requested the change and looked over the pull request before it was merged and I didn't spot this either.

trichoplax:
@Oaphi thanks for the extra improvements - those all look good too. Does this mean we need a third person to review now...?

Oaphi:
> @Oaphi thanks for the extra improvements - those all look good too. Does this mean we need a third person to review now...?

No need to, I've merged the PR - a cross-review we've done should be enough in this case, imo

PR #1619, opened 2025-05-11
Move bulleted/ordered list and tables editor buttons to the first level

Oaphi:
Re: meta:293936 (doesn't close it, but related) and meta:293940 (partially addresses it)

~~Unfortunately, it competes with #1618 and will make it obsolete if merged.~~ (merged in as per discussion)

# Rationale

It's very unusual for editor tools to not show list management buttons at first level, so I think we should not reinvent the wheel. Examples:

GitHub editor tools:

Stack Overflow editor tools:

Well-known CKEditor:

And countless others.

# Changes

The change makes the tools panel look like this for a sufficiently wide viewport (note that list and table insertion buttons are no longer wrapped under the "tools" modal):

And like this for a narrow one (we should improve button positioning on narrow viewports in general, but that's outside the scope of the PR):

### DIFF
--- a/app/helpers/markdown_tools_helper.rb
+++ b/app/helpers/markdown_tools_helper.rb
@@ -22,6 +22,7 @@ module MarkdownToolsHelper
     attribs.merge! href: 'javascript:void(0)',
                    class: "#{attribs[:class] || ''} button is-muted is-outlined js-markdown-tool",
                    data_action: action,
+                   draggable: false,
                    aria_label: label,
                    title: label,
                    role: 'button'
@@ -42,6 +43,7 @@ module MarkdownToolsHelper
     attribs.merge! href: 'javascript:void(0)',
                    class: "#{attribs[:class] || ''}js-markdown-tool",
                    data_action: action,
+                   draggable: false,
                    aria_label: label,
                    title: label,
                    role: 'button'
--- a/app/views/shared/_markdown_tools.html.erb
+++ b/app/views/shared/_markdown_tools.html.erb
@@ -58,11 +58,14 @@
     <%= md_button label: 'Insert image', class: 'is-icon-only', data_modal: '#markdown-image-upload' do %>
       <i class="fas fa-fw fa-images"></i>
     <% end %>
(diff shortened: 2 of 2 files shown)


cellio:
I'm happy to defer to this PR and close mine.  Would you be willing to expand the tooltip here to reflect what's still in there?  In particular, while lines and headings are kind of uninteresting, tables should probably be more findable than they are.

cellio (commented):
LGTM other than the tooltip question.

Oaphi:
> In particular, while lines and headings are kind of uninteresting, tables should probably be more findable than they are.

@cellio what do you think about moving the table button to the first level too (I think it's best suited with the other "insert" buttons)? This is common practice to do so as well

Oaphi:
> LGTM other than the tooltip question.

Sounds good to me, I'll merge your PR into mine then so as we get the best of both worlds (after the pipeline wraps up)

cellio:
> Sounds good to me, I'll merge your PR into mine then (after the CI pipeline finishes running that is) so as we get the best of both worlds.

Works for me.  I agree with promoting tables, at which point we're down to two tools right now.  You'll need to adjust the wording on the tooltip.

Oaphi:
> Works for me. I agree with promoting tables, at which point we're down to two tools right now

@cellio agreed then, will do shortly - and we can also partially address another request that's just came in: meta:293940

app/helpers/markdown_tools_helper.rb
```diff
@@ -22,6 +22,7 @@ def md_button(name = nil, action: nil, label: nil, **attribs, &block)
     attribs.merge! href: 'javascript:void(0)',
                    class: "#{attribs[:class] || ''} button is-muted is-outlined js-markdown-tool",
                    data_action: action,
+                   draggable: false,
```
Oaphi, line 25: Forgot to mention it, but this change solves a minor issue of being able to drag & drop markdown buttons themselves into the editor (resulting in `javascript:void(0)` being inserted)

cellio (approved):
LGTM.

Issue #1418, opened 2024-09-27
Broken granting and revoking of mod and admin abilities


trichoplax:
**Describe the bug**

---
**Although I'm testing this on the `develop` branch, I have database changes locally, related to the `luap42/moderator_tools_improvements` branch, so it would be useful if someone who has not tested that branch could confirm my observations on the `develop` branch. I'm not expecting that to be related but it would be good to have independent confirmation.**

---

As a network admin, you can grant a user roles, including making them a community moderator or admin, or a network moderator or admin. You can also revoke each of these roles.

This process appears to be broken. Some abilities trigger changes in other abilities, but not the expected abilities. Some clicks on buttons make background changes that do not show up until you press another button afterwards. It's like playing a puzzle game.

- It might be expected that granting network moderator would also grant community moderator. It does. However, it also grants community admin.
- It might be expected that granting network admin would also grant community admin, and possibly network moderator. It grants neither of these, but does grant community moderator.
- It might be expected that granting ***community admin*** would also grant community moderator. It does. However, it does not grant ***community admin*** (that's not a typo - it does not grant itself). That is, the button being pressed (grant community admin) grants only community moderator, and not community admin. Repeatedly clicking on community admin has no effect on that button, but does switch on and off community moderator. Clicking on grant or revoke community moderator only works if the community admin button has not been used, otherwise the community moderator button has no effect until the community admin button is pressed again, at which point the community moderator button goes back to working as expected.

**To Reproduce**
- Go to the user profile page.
- Click on "Moderator Tools".
- Click on "privileges".
- Scroll to the bottom of the page and try various combinations of the buttons under "Roles".

**Expected behavior**
To be discussed. Clearly what I describe above is not the correct behaviour, but there are several potential ways to make this work. Which is most intuitive is probably subjective. Whichever approach is chosen, my concern is that it be made clear and unambiguous.

When revoking network admin, should this also revoke community admin on every community? Intuitively, some might expect the answer to be "yes" if community admin was only a side effect of network admin, but "no" on any communities where community admin was already granted before network admin was granted. I can also imagine cases where someone would want to remove community admin on all communities at once, even if granted before network admin. Ideally the interface would cater to both situations. Similarly for revoking network moderator.

[...]

PR #1622, opened 2025-05-12
Fix role_toggle breaking due to access control methods rename

### LINKED ISSUE #1621: Granting / revoking roles manually is broken
**Describe the bug**

Mea culpa, I _did_ manage to miss a case during the rename to access control methods in #1584. As a result, trying to grant / revoke privileges manually results in a server error as the methods no longer exist (I assumed those were column names and missed the metaprogramming part that these are method names the user model is supposed to respond to [that also means we need better coverage]).

I'll handle the fix myself, of course, but logging it just in case.

**To Reproduce**

Just try to manually grant a user a privilege (for example, by making them a moderator).

Oaphi:
closes #1621

This PR also fixes a couple of issues present before the rename, specifically (hereinafter, "the user" refers to the user being granted/revoked a role, not the granting/revoking user):

- it's no longer impossible to revoke the `admin` role as the check for whether the user is a mod used to check for "at least a moderator", resulting in us taking the incorrect path;
- the check for whether the user is an admin no longer incorrectly passes if the user is a global moderator;
- `role_toggle` no longer fails when trying to revoke the `mod` role if it is granted only via `is_moderator` (but not as an ability);
- `role_toggle` now has coverage for community-specific & global roles;


cellio (approved):
LGTM (someone more fluent should also review the code changes).  I pulled the branch and tested manually, and granting/revoking abilities and mod status works.

Oaphi:
@cellio thanks! Quick note: I've added a couple more tests to cover the global roles as well, no code changes

cellio:
Also tested admin, global mod, and global admin; they toggle as expected.

Oaphi:
@trichoplax adding you as I believe your #1418 is at least partially addressed by this PR (specifically, incorrect behavior regarding granting / revoking community admin role).

ArtOfCode- (approved):

PR #1626, opened 2025-05-14
Fix for CodeGolf's leaderboard sorting when grouped by language

Oaphi:
This is a followup PR based on improvements made in #1616. While we _have_ fixed ungrouped sorting, we still have an issue with language groups, specifically, we don't have a stable sort applied to groups themselves. This PR makes it so that language groups are sorted alphanumerically.

As a result, ATTOW, the sample leaderboard will look like this when grouped by language with this PR's changes (note that the invalid submission is _correctly_ not sorted to the bottom as it's actually _partially valid_ [language name is successfully parsed, even if it's "ESCR - not a golfing submission"]):

This PR also further improves JSDoc and, subsequently, type inference for the asset for TS-aware IDEs and ensures that `score` is a number, so as we don't end up with alphanumeric sorting.

### DIFF
--- a/public/assets/community/codegolf.js
+++ b/public/assets/community/codegolf.js
@@ -15,8 +15,10 @@
  *   variant?: string
  *   extensions?: string
  *   code?: string
- *   score?: string
+ *   score?: number
  * }} ChallengeEntry
+ * 
+ * @typedef {(a: ChallengeEntry, b: ChallengeEntry) => number} SortComparator
  */
 
 (() => {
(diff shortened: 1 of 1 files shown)


ArtOfCode- (approved):

Oaphi:
I also believe we have another bug here in that even though the sample challenge is not tagged `lowest-score`, we still sort the leaderboard as if the lower the score, the higher the position. Which seems kind of backwards to me, but not sure since I don't know the community's expectations.

cellio (approved):
I'm assuming that the other approver did a code review.  Behavior LGTM.

PR #1630, opened 2025-05-16
Don't show answer sorting buttons if there are none

### LINKED ISSUE #1350: On question pages, don't show answer sorting buttons until there are answers
meta:290806

The answer sorting buttons (Score, Active, Age) serve no function until there are two or more answers (that the current user can see -- this could include deleted answers).  Let's not show the sorting buttons until there's something to sort.

I believe that every action that could lead to a new answer showing up involves a page refresh, so we aren't risking an answer "winking in" while the buttons are gone.  Some sites work like that, but as far as I can tell, qpixel doesn't -- you won't see new (or undeleted) answers until you refresh the page, at which point we'd recheck whether to show the buttons.

Oaphi:
closes #1350 

Additionally, this PR:

- removes the "<N> Answers" header if there are no answers visible to the user;
- ensures the `N` in the abovementioned header accounts for deleted answers that the user can access;  
- removes the line from the top of the title line of first-level posts (see [Should a question page have a line above the title?](https://meta.codidact.com/posts/293691) - it made sense for me to make the change while I am at it);

Before the change:

Screenshot from 2025-05-16 17-16-11

After the change:

image

Same view for a user that can see deleted answers (the example post has 1 undeleted + 2 deleted answers -> 3 total):

image

The PR also replaces (as per discussion below) the "your answer" static header with a dynamic one. If there are no answers (that are visible to the user), "Be the first to answer" will be shown:

image

If there are, "Add an answer" will be shown instead:

Screenshot from 2025-05-17 12-59-26

---

Regarding removal of the line from the top of the title line, here is how a first-level post will look like with the change:

image

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -16,12 +16,18 @@ class UsersController < ApplicationController
 
   def index
     sort_param = { reputation: :reputation, age: :created_at }[params[:sort]&.to_sym] || :reputation
+
     @users = if params[:search].present?
                user_scope.search(params[:search])
              else
-               user_scope.order(sort_param => :desc)
-             end.where.not(deleted: true).where.not(community_users: { deleted: true })
-                .paginate(page: params[:page], per_page: 48) # rubocop:disable Layout/MultilineMethodCallIndentation
+               user_scope
+             end
+
+    @users = @users.where.not(deleted: true)
+                   .where.not(community_users: { deleted: true })
+                   .order(sort_param => :desc)
+                   .paginate(page: params[:page], per_page: 48)
+
     @post_counts = Post.where(user_id: @users.pluck(:id).uniq).group(:user_id).count
   end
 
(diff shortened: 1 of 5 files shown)


cellio:
I agree that the spacing looks weird without the intervening buttons.  More space would be good, but can we do even better by combining the headings in the 0-answer case?  "Be the first to answer", "0 answers - add yours", something like that?

Oaphi:
> but can we do even better by combining the headings in the 0-answer case?

Yeah, I like the idea, no need to waste space in this case. That said, do note that I also opted to remove the buttons if there's only one answer - meaning that we should be careful with when we merge the headers.

cellio:
> > but can we do even better by combining the headings in the 0-answer case?
> 
> Yeah, I like the idea, no need to waste space in this case. That said, do note that I opted to remove the buttons if there's only one answer as well - meaning that we should be careful with when we merge the headers.

Oh, I missed the single-answer case.  Yeah, that changes things a little.  The *spacing* is only a problem when there are 0 answers and those two headings are right next to each other.  So I guess the logic I'm proposing (if this is too complicated we should punt) is:

- "N answers" heading: if 0 then omit, else show as we do now
- Heading above answer textbox: if 0 answers then customize text, else "your answer" like we do now

app/views/posts/show.html.erb
```diff
@@ -19,20 +19,25 @@
 
 <%= render 'posts/expanded', post: @post, float_notice: false %>
 
+<% num_answers = @children.count %>
```
cellio, line 22: Info: does this count all answers or only answers the viewing user can see?  (The PR shows an example for people who can see deleted answers; just want to make sure we're doing the right thing for people who can't.)

Oaphi: Specifically only those that the user can see, yes - `@children` is a local variable set by the controller that does the check (the same one is currently used to display the list of answers)

PR #1634, opened 2025-05-20
show who reviewed an edit in the 'decided' list

cellio:
Small change suggested in chat recently: in the list of handled suggested edits, show who handled it.  The information is shown on the individual item if you click through, but this makes it a little easier to find a specific review.

screenshot: adds "handled by (name)" after the editor name and time


Oaphi:
Not a blocker, but I'd move the info to the next line - the line gets a bit unreadable given a sufficiently long username - like this (on a tangent, I also think that the line should say "decided by", not "reviewed by" to be future-proof):

image

cellio:
> Not a blocker, but I'd move the info to the next line - the line gets a bit unreadable given a sufficiently long username - like this (on a tangent, I also think that the line should say "decided by", not "reviewed by" to be future-proof):

Good points, and if we move to the next line, including when it was reviewed should work for most usernames (and not be horrible for long ones).

cellio:
Revised screenshot:

Decided by (name), (time) - on new line

Oaphi (approved):

PR #1637, opened 2025-05-24
fixed parsing of partial CG scores

### LINKED ISSUE #1636: Code Golf leaderboard incorrectly interprets fractional scores
**Describe the bug**
The leaderboard takes the last contiguous string of digits in the heading to be the score, which works fine for integers but fails for fractional scores like 6.25, which is interpreted as 25.

**To Reproduce**
1. Go to [a fractionally scored code golf answer](https://codegolf.codidact.com/posts/279820/294008#answer-294008).
2. See that the score in the heading has to be rounded up to the nearest byte to work with the current leaderboard.

**Expected behavior**
Since there is [consensus that fractional scores are acceptable](https://codegolf.codidact.com/posts/288547) (for example, for languages that can have their code measured in bits rather than bytes), and since other challenge types based on arbitrary score calculations may have arbitrary fractional scores, the leaderboard should interpret the last number in the heading as the score, including any decimal point.

I've marked this as a bug rather than a feature request, because even before the consensus on allowing fractional code golf scores, fractional scores for other challenge types were always a possibility.

Oaphi:
closes #1636

### DIFF
--- a/public/assets/community/codegolf.js
+++ b/public/assets/community/codegolf.js
@@ -120,7 +120,8 @@
           ".user-card--content .user-card--link",
         );
 
-        const matchedScore = header?.innerText.match(/\d+/g)?.pop();
+        // https://regex101.com/r/BjIjk5/2
+        const matchedScore = header?.innerText.match(/\d+(?:\.\d+)?/g)?.pop();
 
         /** @type {ChallengeEntry} */
         const entry = {


trichoplax:
Since none of the changed files are covered by tests, I'm assuming that the test failures after merging in `develop` are sporadic. I'm therefore merging this into `develop` regardless.

PR #1642, opened 2025-05-28
Restrict users from directly editing posts in categories they are restricted from participating

### LINKED ISSUE #1605: Edit Posts ability allows editing in a restricted category
**Describe the bug**
In a restricted category (such as Blog on Meta), only users with access can add new posts. These users should also be the only ones able to edit posts, but a user who gains the Edit Posts ability for that community can then edit posts in the restricted category.

**Expected behavior**
1. After gaining the Edit Posts ability for a community, a user should still see "Suggest edit" rather than "Edit" below a post in a restricted category, so their posts will still need to be reviewed by a user with access.
2. After gaining the Edit Posts ability for a community, a user should still not be able to approve suggested edits in a restricted category.

**Screenshots**
Notice the Edit button (even though I don't have access to post in the Blog category).
A post in the restricted Blog category, showing "Edit" below the post

Oaphi:
closes #1605

With this change, even users with the ability to edit the post will only be able to submit suggested edits unless they meet category requirements (if it's present) [both screenshots show a post with a pending suggested edit made by the user lacking access]:

image

And this is how the same post looks like for a user with appropriate access:

image

This PR also adds a couple of partials to make it easier to work with our rather complex view logic when it comes to editing.

Additionally, the PR fixes broken error notice when attempting to reject a suggested edit without access ("Failure: undefined" due to what is likely just a typo).

Finally, it ensures user trust level is recalculated when manually granting / revoking user roles (currently, **it is not updated**).

On a tangent: we should review whether we still need the `check_your_post_privilege` helper (it's legacy and significantly differs from server-side checks at this point)

### DIFF
--- a/test/controllers/suggested_edit_controller_test.rb
+++ b/test/controllers/suggested_edit_controller_test.rb
@@ -45,7 +45,7 @@ class SuggestedEditControllerTest < ActionController::TestCase
     suggested_edit.update(active: true, accepted: false)
 
     post :approve, params: { id: suggested_edit.id }
-    assert_response(400)
+    assert_response(:forbidden)
   end
 
   test 'signed-out shouldn\'t be able to reject' do
@@ -53,7 +53,7 @@ class SuggestedEditControllerTest < ActionController::TestCase
     suggested_edit.update(active: true, accepted: false)
 
     post :reject, params: { id: suggested_edit.id, rejection_comment: 'WHY NOT?' }
-    assert_response(400)
+    assert_response(:forbidden)
   end
 
   test 'users without the ability to edit posts shouldn\'t be able to approve' do
@@ -62,7 +62,7 @@ class SuggestedEditControllerTest < ActionController::TestCase
     edit = suggested_edits(:pending_high_trust)
 
     post :approve, params: { id: edit.id, format: 'json' }
-    assert_response(:bad_request)
+    assert_response(:forbidden)
 
     assert_nothing_raised do
       JSON.parse(response.body)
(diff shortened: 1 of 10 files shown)


cellio (commented):
Looks good so far, based on inspection.  (I won't be able to test for a few days.)  I see that this is marked as a draft so I assume more changes are coming (hence "so far" and not hitting the approval button yet).

Oaphi:
> I assume more changes are coming (hence "so far" and not hitting the approval button yet).

Yup, at least one more change is coming - I closed off the ability to self-review suggested edits in restricted categories client-side, but not server-side yet, which is more important - will switch out of draft mode afterwards.

Oaphi:
I'll leave the missing coverage line be as it's on the code path that normally should not even possible to reach.

test/controllers/suggested_edit_controller_test.rb
```diff
@@ -56,6 +56,40 @@ class SuggestedEditControllerTest < ActionController::TestCase
     assert_response(400)
   end
 
+  test 'users without the ability to edit posts shouldn\'t be able to approve' do
+    sign_in users(:moderator)
+
+    edit = suggested_edits(:pending_high_trust)
+
+    post :approve, params: { id: edit.id, format: 'json' }
+    assert_response(:bad_request)
```
cellio, line 65: Q: why is this a bad request? Would a comment help or is it just my inexperience?

cellio: I agree with not doing that here. That's a separate issue and might affect other tests too. Thanks for explaining.

This PR looks good to me by inspection. I'm not currently in a position to test; approving with that caveat and I see Art has also reviewed.

cellio (approved):

ArtOfCode- (approved):

PR #1647, opened 2025-06-08
General improvements to comment threads

### LINKED ISSUE #544: Let people flag comments from the in-page preview
When looking at a comment thread, somebody (with the right abilities) can see the following actions for each individual comment: link, edit, delete, flag.  When a thread is expanded on a post page, though, the only available action is link.

Flagging seems like something we should make it easy to do; if there's something wrong in a visible comment, let's let people tell us that instead of making them load the thread page (which some won't do).

Arguments can be made both ways about editing and deleting, but I think flagging is pretty clear.  Can we support that?

### LINKED ISSUE #1579: Indicate to users that can't create new comment threads on a given post they can't do so before they spend time writing it up

### LINKED ISSUE #1670: Avatar size varies in thread follower lists
Minor cosmetic bug: in the list of followers for a thread, the avatar is rendered at different sizes apparently based on the length of the username.  Very few people can see this list and it's not *broken* just *inconsistent*, so unless this is a symptom of a bigger problem, this is low priority.

screenshot

Oaphi:
Actually closes #1579, likely closes #544, definitely closes #1670

This PR builds on the work done in #1584 by preventing the buttons for creating a new comment thread or replying in an existent one from showing up if the user can't comment (not only when it's *impossible* to comment [specifically, when the user is not a mod or an admin & comments on the post are disabled, or it is locked or deleted]).

Most importantly, it adds 2 new site settings for rate-limiting comments:

- `RL_NewUserCommentsOwnPosts` for new user comments on their own posts;
- `RL_CommentsOwnPosts` for unrestricted user comments on their own posts;

Combined with the existing settings, users will now be limited to the following number of comments per day (by default):

- new users: 0 comments on posts of others, 30 on their own posts & answers to them;
- unrestricted users: 50 comments on posts of others, 50 on their own posts & answers to them;

Open to discussion on whether to adjust the number of comments users can make on their own posts per day.

---

The PR also ensures that the add thread / reply buttons are always shown to users but disabled if they can't comment for one reason or the other. The exact reason as to why a user can't comment is moved to the button's title (instead of littering the page with huge notices):

https://github.com/user-attachments/assets/9b6c576e-3e45-464b-af9b-89b6d032399e

It also makes possible to take actions on inline comment threads (comment editing & deleting included):

https://github.com/user-attachments/assets/7b314983-34aa-4522-a945-c2e8ca378455

Plus, locked threads in post comment threads lists now show the lock icon (as do other "restricted" states):

We also no longer display the "Showing X of Y comments. See the whole thread." widget if X is equal to Y (it simply wastes space in such cases):

---

It also adds several scopes (and subsequently increases coverage across the project to make sure the new scopes don't break anything important):

**model scopes**

- `AuditLog.of_type(name)` (only include logs of a specific type by `name`);
- `Comment.by(user)`  (only include ones made by the `user`);
- `Flag.by(user)` (same as `Comment.by`);
- `Flag.declined` (only include those that are handled as declined);
- `Flag.helpful` (only include those that are handled as helpful);
- `ModWarning.active` (only include warnings that are currently active);
- `ModWarning.to(user)` (only include warnings issued to the `user`);
- `Post.bad` (only include posts with score < 0.5); 
- `Post.by(user)` (same as `Comment.by`);
- `Post.good` (only include posts with score > 0.5); 
- `Post.in(category)` (only include posts made in the `category`);
- `Post.on(community)` (only include posts made on the `community`);
- `Post.parent_by(user)` (only include posts that have a parent made by the `user`);
- `Post.problematic` (only include posts with score < 0.25 or that are deleted);
- `PostHistory.by` (same as `Comment.by`);
- `PostHistory.of_type(name)` (same as `AuditLog.of_type`);
- `PostHistory.on_undeleted` (only include events made on undeleted posts);
- `SuggestedEdit.approved` (only include those that are decided as approved);
- `SuggestedEdit.by(user)` (same as `Comment.by`);
- `SuggestedEdit.rejected` (only include those that are decided as rejected);
- `Vote.by(user)` (same as `Comment.by`);
- `Vote.for(user)` (only include those where the target is the `user`);

**concern (shared) scopes**

- `Lockable.locked` (only include ones that are locked);
- `Lockable.unlocked` (the inverse of `Lockable.locked`);
- `SoftDeletable.deleted` (only include ones that are marked as deleted);
- `SortDeletable.undeleted` (the inverse of `SoftDeletable.deleted`);
- `Timestamped.newest_first` (order by created_at DESC);
- `Timestamped.oldest_first` (order by created_at ASC);
- `Timestamped.recent` (only include ones made in the last 24 hours);

as well as removes some existing scopes:

- `CommunityUser.active` (replaced by `CommunityUser.undeleted` to align with other soft-deletable models);
- `User.active` (replaced by `User.undeleted` to align with other soft-deletable models);

It also removes `check_edits_limit!` as it's been unused since 412b49096da9292eb0ed366df3be89a85ff8a3a8 (the check has been done inline for at least 4 years - we might want to abstract it into a similar helper).

Additionally, it adds a QoL `[]=` class method to the `SiteSetting` model for us to be able to update site settings simply via `SiteSetting[name] = value` without having to explicitly handle value type conversion & cache updates.

Tangentially, it also adds two new methods:
- `max_votes_per_day` that functions identically to `max_comments_per_day` (proper logic encapsulation);
- `recent_votes_count` that functions similarly to `recent_comments_count` (unsure why we exclude self votes on parents but not votes on own posts there);

as well as renames some existing ones:

- `User#can_push_to_network` to `can_push_to_network?` (to align with the naming convention for boolean checks);
- `User#can_update` to `can_update?` (same as with `can_push_to_network`);
- `User#can_see_deleted?` to `can_see_deleted_posts?` (the check only applies to posts);
- `User#has_ability_on` to `ability_on?` (same as `can_push_to_network`);
- `User#has_profile_on` to `profile_on?` (same as `can_push_to_network`);
- `User#is_moderator_on` to `moderator_on?` (same as `can_push_to_network`);
- `ModeratorController#verify_can_see_deleted` to `verify_can_see_deleted_posts` (same as with `can_see_deleted?`);

and a couple of shared concerns:

- `Inspectable` that provides a generic `inspect` method that can be used on any model;
- `Lockable` for logic shared by all models representing records that can be locked;
- `SoftDeletable` for logic shared by all models representing soft-deletable records;
- `Timestamped` for logic shared by all models representing timestamped records;
- `UserRateLimits` that houses methods related to various rate limits (f.e., comments & votes per day);
- `UsernameValidations` that encapsulates username-related validations (user model is too large);

Some extras:

- `Layout/SpaceInsideArrayLiteralBrackets` is now enabled to match our style guide for Ruby;
- `assert_json_response_message(expected)` test helper method for asserting that response body contains a given message;
- `assert_redirected_to_sign_in` test helper method for asserting that response has the `:found` HTTP status code & redirects the requestor to sign in;

Related meta issue: meta:294088

### DIFF
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -139,7 +139,7 @@ module CommentsHelper
     comments_limit = user.max_comments_per_day(post)
     is_rate_limited = comments_count >= comments_limit
 
-    unless is_rate_limited
+    unless is_rate_limited && user.standard?
       return [false, nil]
     end
 
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -74,14 +74,14 @@ module CommentsHelper
   # @return [String] error message
   def comments_post_error_msg(post)
     if post.locked?
-      'Comments are disabled on locked posts.'
+      I18n.t('comments.errors.disabled_on_locked_posts')
     elsif post.deleted?
-      'Comments are disabled on deleted posts.'
+      I18n.t('comments.errors.disabled_on_deleted_posts')
     elsif post.comments_disabled
-      'Comments on this post are disabled.'
+      I18n.t('comments.errors.disabled_on_post_specific')
     else
-      'This post cannot be commented on.' # just in case
-    end
+      I18n.t('comments.errors.disabled_on_post_generic')
+    end.strip
   end
 
   # Gets a standard comments error message for a given thread
@@ -89,14 +89,14 @@ module CommentsHelper
   # @return [String] error message
   def comments_thread_error_msg(thread)
     if thread.locked?
-      'Locked threads cannot be replied to.'
+      I18n.t('comments.errors.disabled_on_locked_threads')
     elsif thread.deleted
-      'Deleted threads cannot be replied to.'
+      I18n.t('comments.errors.disabled_on_deleted_threads')
     elsif thread.archived
-      'Archived threads cannot be replied to.'
+      I18n.t('comments.errors.disabled_on_archived_threads')
     else
-      'This thread cannot be replied to.' # just in case
-    end
+      I18n.t('comments.errors.disabled_on_thread_generic')
+    end.strip
   end
 
   ##
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -99,6 +99,14 @@ module CommentsHelper
     end.strip
   end
 
+  # Gets a standard comments rate limit error message for a given user & post
+  # @param user [User] user to get the comments count for
+  # @param post [Post] post to get the comments count for
+  def rate_limited_error_msg(user, post)
+    comments_count = user.recent_comments_count(post)
+    I18n.t('comments.errors.rate_limited', count: comments_count)
+  end
+
   ##
   # Get a list of user IDs who should be pingable in a specified comment thread. This combines the post author, answer
   # authors, recent history event authors, recent comment authors on the post (in any thread), and all thread followers.
@@ -144,18 +152,17 @@ module CommentsHelper
     end
 
     if user.new? && !user.owns_post_or_parent?(post) && comments_limit.zero?
-      message = 'As a new user, you can only comment on your own posts and on answers to them.'
+      message = I18n.t('comments.errors.new_user_rate_limited')
 
       if create_audit_log
         AuditLog.rate_limit_log(event_type: 'comment', related: post, user: user,
                                 comment: "'unrestricted' ability required to comment on non-owned posts")
       end
     else
-      message = "You have used your daily limit of #{comments_count} comments. Come back tomorrow to continue."
+      message = rate_limited_error_msg(user, post)
 
       if create_audit_log
-        AuditLog.rate_limit_log(event_type: 'comment', related: post, user: user,
-                                comment: "limit: #{comments_limit}")
+        AuditLog.rate_limit_log(event_type: 'comment', related: post, user: user, comment: "limit: #{comments_limit}")
       end
     end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -43,8 +43,6 @@ class User < ApplicationRecord
 
   delegate :trust_level, :reputation, :reputation=, :privilege?, :privilege, to: :community_user
 
-  after_create :send_welcome_tour_message, :ensure_websites
-
   def self.list_includes
     includes(:posts, :avatar_attachment)
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -272,6 +272,26 @@ class User < ApplicationRecord
     global_moderator? || global_admin? || false
   end
 
+  # Which communities is this user a moderator (local or global) on?
+  # @return [Community[]] list of communities
+  def moderator_communities
+    if global_moderator?
+      Community.all
+    else
+      Community.joins(:community_users).where(community_users: { user_id: id, is_moderator: true })
+    end
+  end
+
+  # Which communities is this user an admin (local or global) of?
+  # @return [Community[]] list of communities
+  def admin_communities
+    if global_admin?
+      Community.all
+    else
+      Community.joins(:community_users).where(community_users: { user_id: id, is_admin: true })
+    end
+  end
+
   # Is the user a moderator on a given community?
   # @param community_id [Integer] community id to check for
   # @return [Boolean] check result
--- a/app/views/comments/_threads_list.html.erb
+++ b/app/views/comments/_threads_list.html.erb
@@ -23,7 +23,7 @@
     ].compact.join(' ')
   %>
 
-  <div class="post--comments-thread-wrapper" role="listitem">
+  <div class="js-comment-thread-wrapper post--comments-thread-wrapper" role="listitem">
     <% if thread_id.to_i == thread.id.to_i %>
       <%= render 'comment_threads/expanded', comment_id: comment_id,
                                              inline: true,
--- a/app/views/comments/_threads_list.html.erb
+++ b/app/views/comments/_threads_list.html.erb
@@ -3,12 +3,14 @@
 
     Variables:
       comment_threads : an Array of CommentThread instances to list
-    ? expand          : CommentThread ID to render expanded view for
+    ? comment_id      : Comment to show even if it would be hidden otherwise
+    ? thread_id       : CommentThread ID to render expanded view for
     ? show_deleted    : whether to display deleted comments in the expanded view
 %>
 
 <%
-  expand ||= defined?(expand) ? expand : nil
+  comment_id ||= defined?(comment_id) ? comment_id : nil
+  thread_id ||= defined?(thread_id) ? thread_id : nil
   show_deleted ||= defined?(show_deleted) ? show_deleted : false
 %>
 
@@ -22,8 +24,11 @@
   %>
 
   <div class="post--comments-thread-wrapper" role="listitem">
-    <% if expand.to_i == thread.id.to_i %>
-      <%= render 'comment_threads/expanded', inline: true, show_deleted: show_deleted, thread: thread %>
+    <% if thread_id.to_i == thread.id.to_i %>
+      <%= render 'comment_threads/expanded', comment_id: comment_id,
+                                             inline: true,
+                                             show_deleted: show_deleted,
+                                             thread: thread %>
     <% else %>
       <%= render 'comment_threads/collapsed', thread: thread %>
     <% end %>
--- a/app/views/comments/_threads_list.html.erb
+++ b/app/views/comments/_threads_list.html.erb
@@ -15,16 +15,6 @@
   %>
 
   <div class="post--comments-thread-wrapper" role="listitem">
-    <div class="post--comments-thread is-inline <%= state_class %>">
-      <% if thread.deleted %>
-        <i class="fas fa-trash h-c-red-600 fa-fw" title="Deleted thread" aria-label="Deleted thread"></i>
-      <% elsif thread.archived %>
-        <i class="fas fa-archive fa-fw" title="Archived thread" aria-label="Archived thread"></i>
-      <% elsif thread.locked? %>
-        <i class="fas fa-lock fa-fw" title="Locked thread" aria-label="Locked thread"></i>
-      <% end %>
-      <%= link_to thread.title, comment_thread_path(thread), class: 'js--comment-link', data: { thread: thread.id }, role: 'button' %>
-      (<%= pluralize(thread.reply_count, 'comment') %>)
-    </div>
+    <%= render 'comment_threads/collapsed', thread: thread %>
   </div>
 <% end %>
--- a/app/views/comments/_threads_list.html.erb
+++ b/app/views/comments/_threads_list.html.erb
@@ -3,6 +3,13 @@
 
     Variables:
       comment_threads : an Array of CommentThread instances to list
+    ? expand          : CommentThread ID to render expanded view for
+    ? show_deleted    : whether to display deleted comments in the expanded view
+%>
+
+<%
+  expand ||= defined?(expand) ? expand : nil
+  show_deleted ||= defined?(show_deleted) ? show_deleted : false
 %>
 
 <% comment_threads.each do |thread| %>
@@ -15,6 +22,10 @@
   %>
 
   <div class="post--comments-thread-wrapper" role="listitem">
-    <%= render 'comment_threads/collapsed', thread: thread %>
+    <% if expand.to_i == thread.id.to_i %>
+      <%= render 'comment_threads/expanded', inline: true, show_deleted: show_deleted, thread: thread %>
+    <% else %>
+      <%= render 'comment_threads/collapsed', thread: thread %>
+    <% end %>
   </div>
 <% end %>
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -124,7 +124,8 @@ class ActiveSupport::TestCase
 
   def assert_valid_json_response
     assert_nothing_raised do
-      JSON.parse(response.body)
+      parsed = JSON.parse(response.body)
+      assert_not_nil(parsed)
     end
   end
 
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -316,6 +316,7 @@ $(() => {
 
     if ($reply.is(':hidden')) {
       $reply.show();
+      $reply.find('.js-comment-field').trigger('focus');
     }
     else {
       $reply.hide();
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -123,6 +123,7 @@ $(() => {
     </form>`;
 
     $commentBody.html(formTemplate);
+    $commentBody.find('textarea#comment-content').trigger('focus');
 
     $commentBody.find(`#comment-content`).on('keyup', pingable_popup);
 
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -8,7 +8,9 @@ $(() => {
     const resp = await fetch(`/comments/post/${postId}`, {
       headers: { 'Accept': 'text/html' }
     });
+
     const data = await resp.text();
+
     $tgt.parents('.post--comments').find('.post--comments-container').html(data).trigger('ajax:success');
     $tgt.parents('.post--comments').find('.js-more-comments').remove();
   });
@@ -31,6 +33,7 @@ $(() => {
     const resp = await fetch(`${targetUrl}?inline=true&show_deleted_comments=${showDeleted ? 1 : 0}`, {
       headers: { 'Accept': 'text/html' }
     });
+
     let data = await resp.text();
 
     data = data.split("<!-- THREAD STARTS BELOW -->")[1];
@@ -107,7 +110,9 @@ $(() => {
       credentials: 'include',
       headers: { 'Accept': 'application/json' }
     });
+
     const data = await resp.json();
+
     const content = data.content;
 
     const formTemplate = `<form action="/comments/${commentId}/edit" method="POST" class="comment-edit-form" data-remote="true">
@@ -182,7 +187,9 @@ $(() => {
       credentials: 'include',
       headers: { 'Accept': 'text/html' }
     });
+
     const data = await resp.text();
+
     $modal.find('.js-follower-display').html(data);
   });
 
@@ -230,6 +237,9 @@ $(() => {
   const pingable = {};
   $(document).on('keyup', '.js-comment-field', pingable_popup);
 
+  /**
+   * @type {QPixelPingablePopupCallback}
+   */
   async function pingable_popup(ev) {
     if (QPixel.Popup.isSpecialKey(ev.keyCode)) {
       return;
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -153,9 +153,7 @@ $(() => {
     const commentId = $comment.attr('data-id');
     const isDelete = !$comment.hasClass('deleted-content');
 
-    const resp = await QPixel.fetchJSON(`/comments/${commentId}/delete`, {}, { method: isDelete ? 'DELETE' : 'PATCH' });
-
-    const data = await resp.json();
+    const data = await (isDelete ? QPixel.deleteComment(commentId) : QPixel.undeleteComment(commentId));
 
     if (data.status === 'success') {
       if (isDelete) {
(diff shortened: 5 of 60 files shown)


Oaphi:
Leaving the coverage miss be as the logic there hasn't changed at all _and_ both cases are covered already except for the audit log creation part, which definitely hasn't changed.

ArtOfCode- (approved):

app/models/user.rb
```diff
@@ -70,6 +70,13 @@ def same_as?(user)
     id == user.id
   end
 
+  # Does the user own a given post?
+  # @param post [Post] post to check
+  # @return [Boolean] check result
+  def owns?(post)
```
trichoplax, line 76: Not blocking approval, just commenting in case anyone can think of an improvement for this method name.

The meaning of the method is immediately apparent when reading this definition, but if only seeing the method name in use, it could be misinterpreted as meaning author (so a question author might not be expected to be the owner of all of the answers). I don't like any of the names I can think of, but I wonder if anyone can think of something better:
- recursively_owns?(post)
- can_comment_under?(post)
- is_related_to?(post)

Oaphi: Yeah, deciding how to name this method was a pain - I opted for `owns?` as we refer to such checks as "post owner", but agreed that the "or the parent, if any" is a good idea to reflect.

Couldn't think of a better name and still can't, though. `recursively_owns?` looks good, but it isn't clear what "recursively" means at a glance. `owner_or_op?` looks good to me too (`is_` is not idiomatic, as I've learned from one of the discussions with @ArtOfCode-) as long as readers can aprse that "op" stands for "original poster" (I can essily, not aure about the others)

trichoplax: Thanks for the idiom guidance - I'll bear it in mind.

If the comment rate limit is going to be applied even to a user's own posts and answers to them, does this method become redundant? That would avoid having to name it...

ArtOfCode-: What about `owns_post_or_parent?` Bit longer but makes it clear what it does.

Oaphi: Yup, looks *perfect* to me - I was looking for something exactly like that, much obliged!

app/helpers/comments_helper.rb
```diff
@@ -105,22 +105,21 @@ def get_pingable(thread)
...
         message = 'As a new user, you can only comment on your own posts and on answers to them.'
         if create_audit_log
           AuditLog.rate_limit_log(event_type: 'comment', related: post, user: user,
                                   comment: "limit: #{max_comments_per_day}")
         end
         [true, message]
-      elsif recent_comments >= max_comments_per_day
-        message = "You have used your daily comment limit of #{recent_comments} comments. Come back tomorrow to " \
+      elsif comments_count >= max_comments_per_day
+        message = "You have used your daily comment limit of #{comments_count} comments. Come back tomorrow to " \
```
cellio, line 122: Don't we need to return `true` and `message` in here?

cellio (commented):
If a user can't comment (for the reasons in this PR), how is the message delivered to the user?  I see the helper changes and the refactoring, but I don't see a view change involving the message.  What am I missing?

app/helpers/comments_helper.rb
Oaphi: Not sure what you mean by that? Both braches that are supposed to return a (true, string) tuple do retun it? Nothing about the original logic has changed there

cellio: Oh, it was collapsed away and I missed it, oops!

Oaphi:
> If a user can't comment (for the reasons in this PR), how is the message delivered to the user?

Because it isn't since the button will no longer be present if the user can't comment due to either lacking the privilege or being rate-limited in addition to the current reasons. Showing a notice as with the locked posts / posts with comments disabled will likely be a bit too noisy. If my memory serves me right, we decided to switch to always show the button but disable it instead with a tooltip explaining the reason - if that's the case, we can address the issue when we get to it.

Open to ideas for an interim solution until the PR is merged

Oaphi:
> . Is that a big change?

Not a big change, @cellio, just depends on my availability as usual 🙂. Will throw it in if I can spare the time

app/controllers/application_controller.rb
```diff
@@ -299,7 +272,7 @@ def pull_categories
   def check_if_warning_or_suspension_pending
     return if current_user.nil?
 
-    warning = ModWarning.where(community_user: current_user.community_user, active: true).any?
+    warning = ModWarning.to(current_user).where(active: true).any?
```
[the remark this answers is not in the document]
Oaphi: Agreed, it all started with me just being annoyed at having to visually parse the comments query over and over when making changes, but quickly spiraled out of control 😅. Added

app/models/site_setting.rb
```diff
@@ -25,6 +25,18 @@ def self.[](name)
...
   end
 
+  def self.[]=(name, value)
+    key = "SiteSettings/#{RequestContext.community_id}/#{name}"
+
+    setting = applied_setting(name)
+
+    typed_value = SettingConverter.new(value).send("as_#{setting.value_type.downcase}")
+
+    applied_setting(name).update(value: typed_value)
```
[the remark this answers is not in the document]
Oaphi: Whoops, missed that I ended up with a duplicate call here, thanks!

app/models/concerns/identity.rb
```diff
@@ -0,0 +1,12 @@
+module Identity
+  extend ActiveSupport::Concern
+
+  included do
+    # Is this record the same as a given other record?
+    # @param other [ApplicationRecord] record to compare with
+    # @return [Boolean] check result
+    def same_as?(other)
+      id == other.id
```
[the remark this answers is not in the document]
Oaphi: A bit overkill for what it was supposed to do, but sure, it'll also make it more useful (and I also learned a couple of things while trying to add proper tests for the behavior)

app/controllers/flags_controller.rb
```diff
@@ -10,7 +10,7 @@ def new
              PostFlagType.find params[:flag_type]
            end
 
-    recent_flags = Flag.where(created_at: 24.hours.ago..DateTime.now, user: current_user).count
+    recent_flags = Flag.by(current_user).recent.count
```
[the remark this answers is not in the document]
Oaphi: I initially wanted to do exactly that (something along the lines of `:recent, ->(time_ago) { ... }`), but decided to add a simpler scope first and see how it goes (we consistently treat 24 hours ago as "recent" across the codebase). Since you had the same thought, seems like I am not the only odd one out - let's parametrize indeed

app/models/post.rb
```diff
@@ -40,8 +40,12 @@ class Post < ApplicationRecord
...
   # Other validations (shared with suggested edits) are in concerns/PostValidations
 
-  scope :undeleted, -> { where(deleted: false) }
-  scope :deleted, -> { where(deleted: true) }
+  scope :bad, -> { where('score < 0.5') }
+  scope :by, ->(user) { where(user: user) }
+  scope :good, -> { where('score > 0.5') }
+  scope :problematic, -> { where('score < 0.25 OR deleted=1') }
+  scope :parent_by, ->(user) { includes(:parent).where(parents_posts: { user_id: user.id }) }
+  scope :recent, -> { where(created_at: 24.hours.ago..DateTime.now) }
```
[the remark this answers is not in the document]
Oaphi: Same response - agreed (resolving for housekeeping purposes)

app/models/user.rb
```diff
@@ -434,6 +407,38 @@ def do_soft_delete(attribute_to)
...
+  def max_comments_per_day_on_own_posts
+    SiteSetting[new? ? 'RL_NewUserCommentsOwnPosts' : 'RL_CommentsOwnPosts'] || 0
+  end
+
+  # Gets the max number of votes the user can make per day
+  # @return [Integer]
+  def max_votes_per_day
+    SiteSetting[new? ? 'RL_NewUserVotes' : 'RL_Votes'] || 0
+  end
+
```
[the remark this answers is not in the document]
Oaphi: Yup, it sure is - I was worried about scope creep at the time, but we are well past that point now,  so why not 😅

app/models/user.rb
```diff
@@ -55,44 +51,47 @@ def self.search(term)
...
+    edit.post.present? && can_update?(edit.post, edit.post.post_type)
   end
 
   # Can the user reject a given suggested edit?
   # @param edit [SuggestedEdit] edit to check
   # @return [Boolean] check result
   def can_reject?(edit)
-    edit.post.present? && can_update(edit.post, edit.post.post_type)
+    edit.post.present? && can_update?(edit.post, edit.post.post_type)
+  end
```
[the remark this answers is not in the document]
Oaphi: They don't - I just wasn't sure whether I'd end up with 2 different implementations or not after cleaning up the client-side checks. Done - opted to call the shared method `can_decide?(edit)` with `can_approve?` and `can_reject?` as aliases for consumer convenience. We might reach a point where `can_approve?` and `can_reject?` will require action-specific logic, but should that happen, we'll just unalias them back into real methods.

Oaphi:
> also address (or enable addressing) https://github.com/codidact/qpixel/issues/544 ?

I believe so, @cellio - doesn't seem like there's currently any issue with the "flag" button working - I've just successfully tested comment flagging. It's likely that there are more issues that this PR ends up addressing, but I haven't even started looking through issues yet

cellio:
The latest commit might address https://meta.codidact.com/posts/282342/282348#answer-282348 and https://meta.codidact.com/posts/282342/282382#answer-282382 (check in testing).

Oaphi:
For reference as the description is way too long as it is - the "showing X of Y comments" widget will also take less space & come before the "reply to this thread button" section (as it doesn't make sense to put the widget after it). Not married to the centering but, in my view, it's at least better than the left-aligned version (a proper redesign is out of scope of this PR):

Screenshot from 2025-06-27 18-15-02

app/assets/stylesheets/fonts.scss
```diff
@@ -0,0 +1,7 @@
+@import 'variables';
+
+@each $key, $value in $font-sizes {
+  .font-#{$key} {
+    font-size: $value;
+  }
```
ArtOfCode-, line 6: We have font size utilities available from Co-Design - are they not working for something here? If that's the case we should log an issue for it there and re-consider how they work.

app/assets/stylesheets/utilities.scss
```diff
@@ -362,3 +362,15 @@ span.spoiler {
...
+  .border-#{$side}-none {
+    border-#{$side}: none;
+  }
+}
+
+@each $align in $text-aligns {
+  .text-#{$align} {
+    text-align: $align;
+  }
+}
```
ArtOfCode-, line 376: Likewise, these feel like changes that should be part of Co-Design. Perhaps not work for this PR, but we should log the missing utilities so that they can be added later.

app/views/comments/_threads_list.html.erb
```diff
@@ -1,17 +1,30 @@
...
+      comment_threads : an Array of CommentThread instances to list
 %>
 
 <% comment_threads.each do |thread| %>
+  <%
+    state_class = [
+      thread.deleted ? 'is-deleted' : '',
+      thread.archived ? 'is-archived' : '',
+      thread.locked? ? 'is-locked' : ''
+    ].select(&:present?).join(' ')
```
ArtOfCode-, line 14:
```suggestion
    state_class = [
      thread.deleted ? 'is-deleted' : nil,
      thread.archived ? 'is-archived' : nil,
      thread.locked? ? 'is-locked' : nil
    ].compact.join(' ')
```

test/test_helper.rb
```diff
@@ -161,3 +171,106 @@ def load_host
...
+    get :show, params: { id: comment.id, format: format }
+  end
+
+  # Attempts to update a given comment
+  # @param comment [Comment] comment to update
+  # @param content [String] new content of the comment, if any
+  def try_update_comment(comment, content: 'Edited comment content')
+    post :update, params: { id: comment.id, comment: { content: content } }
+  end
+end
```
ArtOfCode-, line 276: Should this be its own file?

app/assets/stylesheets/fonts.scss
Oaphi: Yes, unfortunately, they are not - as you can see in the [typography section](https://design.codidact.org/utilities/typography/), these utilities are in absolute units _and_ do not match our usage in QPixel (the latter is the primary reason for the additions, though), so I ended up gathering `font-size` values scattered across QPixel's rulesets in one map.

These utilities should be moved to Co-Design, however, I feel like doing so now will end up stalling the PR for quite a while, so I'd keep them here until we can spare some bandwidth to move them over

app/assets/stylesheets/utilities.scss
Oaphi: `text-(left|right|center)` can probably go - I missed that the text alignment utils are described in the [typography section](https://design.codidact.org/utilities/typography/) of Co-Design (that said, I heavily prefer shorter WindiCSS/UnoCSS style `text-center` and the like to `has-text-align-center`, which is extremely verbose for what it does [but that's an overall issue with our utilities that needs to be discussed separately])

test/test_helper.rb
Oaphi: Agreed, just haven't got around to doing so - I've built the concern up incrementally while splitting the large comments controller test into manageable chunks

Oaphi:
Sigh, we merged in something with coverage misses, haven't we? Will fix a bit later.

P.S. Note to folks stumbling on the PR: it's 99% done, feel free to review even though it's not marked as ready yet to avoid wasting time, the PR is massive. I only plan on adding a couple of tests and translation strings

app/assets/stylesheets/fonts.scss
ArtOfCode-: Absolutely, not suggesting we add more work here. Let's file it for later.

Oaphi:
Aaaand we are done, _finally_ - have at it, folks!

cellio:
Testing notes:

- Tested the own-comments limit for someone without Participate Everywhere and worked as advertised: upon reaching the limit buttons are grayed out with suitable message, and buttons on others' posts are grayed out at all times with suitable message.

- Tested editing, flagging, deleting, undeleting a comment from the in-page view -- beautiful!

- Confirmed that curators, who can lock posts, cannot comment on them (as expected), and that the button is grayed out and has a suitable tooltip.  

- Also confirmed that moderators can comment on locked posts (as expected).  Future (do not block on this!): I wonder if we should change the presentation for mods at all for the "most people can't comment but you can" cases; they get a tooltip, but people might not notice that especially on mobile.

- When clicking on the "reply" button, either in-page or in a thread, can the textbox get focus automatically instead of the user having to click into it?

- I love that the tools menu is available in-page.  When I click on it from a comment thread way down on the page, though, it takes me to the top of the page (not sure if it's refreshing and defaulting to the top or what).  As a user, this means I click on "tools" and then don't see it until I page down again.

- As mentioned in chat, when you add a comment from the in-page view, we take you to the thread.  If the thread is already long this is the only way to see your new comment, but we probably don't want to change the redirect behavior based on whether your new comment would be visible.  I think leaving you on the in-page view where you started is ok *if* we can somehow indicate that your comment was successfully posted, since you might not see it.  I'm not sure of the best way to do that, both UX and code, so I think it's ok to redirect to the thread page until we figure that out.  But we should remember to alert people to this when we announce the new comment features.

- As a mod, I locked a thread from the in-page view and it took me to the thread.  There are no issues with showing new comments here, so let's leave you on the original page, adding the lock icon.

- Minor layout thing: when there are deleted comments among the first 5, the indication of such (for those who can see deleted comments) has lost the padding it used to have:

Screenshot

cellio (commented):
Things from my previous comment that I'd like us to fix on this PR (others can be deferred or might require more discussion):

- give focus to textbox after clicking "reply" (in-page and thread views)
- opening the tools menu should not take you back to the top of the page
- after locking a thread, stay on the page (don't go to the thread view)  (presumably also for unlock and archive, which I didn't test)

ArtOfCode- (approved):

cellio:
> Approved from a code POV - haven't run locally to look at the UI/UX but looks like @cellio has covered that.

Yup.  I haven't looked at the code, just the user-facing behavior.

Oaphi:
Finish line TODO list:

- [ ] consider displaying moderator exemption message inline with the new thread / reply to thread buttons;
- [ ] autofocus the reply textarea when the reply to thread button is clicked;
- [ ] preserve scroll state after opening tools (might be better suited for Co-Design);
- [ ] show the newly created comment inline if it's been created via the inline thread view;
- [ ] do not redirect to the full view after locking / unlocking a thread from the inline view;
- [ ] optional: improve styling of the "skipping deleted comment" thread section;

cellio:
> * preserve scroll state after opening tools (might be better suited for Co-Design);

Or if that's too hard (or sends us into Co-Design), if the refresh went to the post link instead of the page link (e.g. the specific answer), you'd at least land nearby.

Oaphi:
After successfully replying to a thread, we now redirect back to the post (unless the reply is made from the full view), but also ensure that the thread is expanded & that the newly created comment is shown even if it would be hidden due to the thread having 5 or more replies already (the count is adjusted accordingly):



For now, this is done via query parameters (which means that we can now do direct thread/comment links to inline views - that said, I haven't added any special styling to indicate that the comment is directly linked to - not yet at least). Ideally, this should be done as a new route `posts/<post_id>/threads/<thread_id>/comments/<comment_id>` (or something similar), but that's way too much work:


https://github.com/user-attachments/assets/a214555c-70b0-4b92-84bd-fbf694592ef8

Oaphi:
And this is what I opted for to make moderator exemption message easier to find - a muted icon will be shown to the right of the button with the exemption text as its title displayed on hover:

https://github.com/user-attachments/assets/3f4ec9e1-8b0c-4559-a9e0-2bfa42ca5dbb

cellio:
> After successfully replying to a thread, we now redirect back to the post (unless the reply is made from the full view), but also ensure that the thread is expanded & that the newly created comment is shown even if it would be hidden due to the thread having 5 or more replies already (the count is adjusted accordingly):

Looks great!  I was going to ask if, in that inline view after posting the comment, we could have some sort of "15 more comments" or "..." or something as a separator before the new comment, but the only person who will ever see that view is the author of that new comment, and if that person didn't look at the intervening comments before posting, I don't think a reminder now will change anything.  (I'm leaving this comment to raise and then explain away that issue. :-) )

> For now, this is done via query parameters (which means that we can now do direct thread/comment links to inline views - that said, I haven't added any special styling to indicate that the comment is directly linked to - not yet at least). Ideally, this should be done as a new route `posts/<post_id>/threads/<thread_id>/comments/<comment_id>` (or something similar), but that's way too much work:

Completely agreed -- if we want that (and I don't know if we do), that is a "future us" problem, not here!

Oaphi:
> if, in that inline view after posting the comment, we could have some sort of "15 more comments" or "..." or something as a separator before the new comment

Was already working on it 😇. This is what I intend to add (note that I also reduced size of all the "skipping M deleted", "skipping M more", and "showing M of N" widgets to save up on vertical space for long threads):



Later on, we can add a link that will load the skipped comments inline and subsequently get rid of the "showing M of N" widget.

Oaphi:
Probably won't move it from the draft status for a couple more days (want to improve coverage after the latest changes first), but I believe it's ready otherwise. Made re-review requests as there's been quite a few changes since the initial approvals (mostly what's listed in the [finish line comment](https://github.com/codidact/qpixel/pull/1647#issuecomment-3025651703))

cellio:
> Probably won't move it from the draft status for a couple more days (want to improve coverage after the latest changes first), but I believe it's ready otherwise. Made re-review requests as there's been quite a few changes since the initial approvals (mostly what's listed in the [finish line comment](https://github.com/codidact/qpixel/pull/1647#issuecomment-3025651703))

I'll test the things in the finish-line comment.  I assume someone else will review the code changes.

cellio:
Testing notes based on checklist:

- autofocus the reply textarea when the reply to thread button is clicked - I'm not getting auto-focus using Chrome; does it work in Firefox? (haven't tried yet) - I don't think we need to block on this; we can do as a followup

Confirmed working:
- display moderator exemption message inline with the new thread / reply to thread buttons
- preserve scroll state after opening tools  (took me nearby, not exactly on that answer, but ok; I tested on a busy page)
- show the newly created comment inline if it's been created via the inline thread view ❤️ 
- do not redirect to the full view after locking / unlocking a thread from the inline view
- optional: improve styling of the "skipping deleted comment" thread section
- clicking on the "edit" link after entering comment edit mode shouldn't break the discard edit button

cellio (approved):

cellio:
That fixed it, thanks!  (Tested Chrome and Firefox, though it doesn't sound like this was browser-dependent anyway.)

Issue #1638, opened 2025-05-25
Audit tag renames


cellio:
Tag merges are recorded in the audit log (available in admin tools), but tag renames are not.  The audit log is the only tag history we have right now and post history can be confusing for tag-related events.  Could we log renames the same way we audit merges?

PR #1648, opened 2025-06-10
audit tag renames

cellio:
Adds tag renames to the audit log -- merges are already there, and renames, like merges, don't leave visible history other than the log.  Fixes https://github.com/codidact/qpixel/issues/1638.

screenshot of log entry

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -121,13 +121,15 @@ class TagsController < ApplicationController
     status = false
 
     @tag.transaction do
+      old_tag_name = @tag.name
+
       status = @tag.update(name: params[:name])
 
       if status
         AuditLog.moderator_audit(event_type: 'tag_rename',
                                  related: @tag,
                                  user: current_user,
-                                 comment: "#{@tag.name} renamed to #{params[:name]}")
+                                 comment: "#{old_tag_name} renamed to #{params[:name]}")
       else
         raise ActiveRecord::Rollback
       end
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -128,6 +128,8 @@ class TagsController < ApplicationController
                                  related: @tag,
                                  user: current_user,
                                  comment: "#{@tag.name} renamed to #{params[:name]}")
+      else
+        raise ActiveRecord::Rollback
       end
     end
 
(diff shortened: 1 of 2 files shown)


cellio:
I should probably figure out how to add a test.  I do not understand the system-test failure (is it spurious?).

Oaphi:
> I do not understand the system-test failure (is it spurious?).

Yeah, system tests are just flaky, it's annoying, but for now until I or someone else sits down and fixes them, the only remedy is to rerun the workflow whenever it fails on CircleCI (https://app.circleci.com/pipelines/github/codidact - you should have access to it via the org - just "rerun from failed" if no one's around to do so - might take a couple of tries, though):

Oaphi:
> I should probably figure out how to add a test, but there's no merge test for me to piggyback on so I'm a bit lost.

Adding tests for audit logs should be relatively straightforward: just test the controller method in question and then ask the model to get you the most recent audit log of the relevant type - even checking that it's there at all should be enough

Oaphi:
I've added the test (not for the audit log, but for the controller method - it's more important to have the base functionality covered - if you want to get some practice, feel free to add one specifically for the log [best to just expand the test I've made] - do ping me if you need help with that) and several fixes (will comment on them separately).

app/controllers/tags_controller.rb
```diff
@@ -120,7 +120,7 @@ def children
   def rename
     status = false
 
-    Post.transaction do
+    @tag.transaction do
```
cellio, line 123: I got `Post.transaction` from `merge` a few lines down.  Do we need to fix it there too?

test/controllers/tags_controller_test.rb
```diff
@@ -170,8 +170,61 @@ class TagsControllerTest < ActionController::TestCase
...
     patch :update, params: { id: categories(:main).id, tag_id: tags(:topic).id,
                              tag: { parent_id: tags(:child).id, excerpt: 'things' } }
-    assert_response 400
+    assert_response :bad_request
     assert_not_nil assigns(:tag)
     assert_equal ["The #{tags(:child).name} tag is already a child of this tag."], assigns(:tag).errors.full_messages
   end
+
+  test 'should correctly rename a tag' do
+    sign_in users(:admin)
```
cellio, line 179: Should we test these as `:moderator` rather than `:admin`?  Moderators *should* be able to do this; if we ever had a bug in that, admin super-privileges might mask it.

ArtOfCode- (approved):

PR #1652, opened 2025-06-13
Code style improvements (mostly for tests)

Oaphi:
This PR is a bit of [yak shaving](<https://en.wiktionary.org/wiki/yak_shaving>) done during my morning routine. The changes mostly improve enforcing of our code style for tests & make a couple of related changes, specifically:

- Rubocop now targets Ruby >= 3.1 as we no longer support 2.7 (I disabled some of the 3.0+ rules for backwards compatibility);
- `Style/MethodCallWithArgsParentheses` is now enforced for tests with some exceptions to avoid making the diff even bigger than it is;
- `assert_response` now consistently uses [HTTP status code constants](https://gist.github.com/mlanett/a31c340b132ddefa9cca) instead of hardcoded values (constants are easier to understand unless one knows the codes by heart);
- It also adds a test `assert_valid_json_response` helper encapsulating the following (we use it a lot):
    ```ruby
    assert_nothing_raised do
      JSON.parse(response.body)
    end
    ```

### DIFF
--- a/test/integration/tasks_integration_test.rb
+++ b/test/integration/tasks_integration_test.rb
@@ -5,18 +5,18 @@ class TasksIntegrationTest < ActionDispatch::IntegrationTest
 
   test 'should deny access to anonymous users' do
     get '/maintenance'
-    assert_response(403)
+    assert_response(:forbidden)
   end
 
   test 'should deny access to non-developers' do
     sign_in users(:admin)
     get '/maintenance'
-    assert_response(403)
+    assert_response(:forbidden)
   end
 
   test 'should grant access to developers' do
     sign_in users(:developer)
     get '/maintenance'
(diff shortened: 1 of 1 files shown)


test/integration/tasks_integration_test.rb
```diff
@@ -5,18 +5,18 @@ class TasksIntegrationTest < ActionDispatch::IntegrationTest
 
   test 'should deny access to anonymous users' do
     get '/maintenance'
-    assert_response 403
+    assert_response(403)
```
ArtOfCode-, line 8: Are the symbols not available in integration tests?

ArtOfCode- (approved):

ArtOfCode-:
LGTM

PR #1659, opened 2025-06-24
Add registration rate limit

ArtOfCode-:
Adds a configurable rate limit for how quickly accounts can be created from the same IP. Related to the ongoing email issues.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -201,4 +201,15 @@ class AdminController < ApplicationController
       render :change_back
     end
   end
+
+  def do_email_query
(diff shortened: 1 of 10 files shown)


Oaphi (changes requested):
> You cannot call create unless the parent is saved

Unfortunately, `send_welcome_tour_message` causes this error as it was supposed to be an `after_create` callback. The same problem is likely to apply to the `ensure_websites` callback

Oaphi:
> Where are you seeing that? I haven't hit that error anywhere

Any attempt to manually sign up a user with the PR checked out - you don't get this error, @ArtOfCode-?

PR #1661, opened 2025-06-27
Add an admin email lookup tool

ArtOfCode-:
Allows admins to look up user accounts by email address (for i.e. deletion requests).

Only shows profiles in communities on which the querying user is an admin.


Oaphi:
I've also left some suggestions as a standalone PR targeted to this one (I don't know of a better way of making multi-file suggestions via GitHub)

Oaphi (approved):
LGTM, everything works as expected!

PR #1662, opened 2025-06-27
Add summary emails

ArtOfCode-:
Adds content summary emails sent to staff. Helps in raising our email volume to avoid AWS issues...

### DIFF
--- a/test/jobs/send_summary_emails_job_test.rb
+++ b/test/jobs/send_summary_emails_job_test.rb
@@ -1,6 +1,8 @@
 require 'test_helper'
 
 class SendSummaryEmailsJobTest < ActiveJob::TestCase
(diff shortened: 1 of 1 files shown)


Oaphi (changes requested):
Suggestions incoming shortly - it also seems it's the first time we've ever tried to pass `ActiveRecord::Relation` as a parameter to `.with(` - which I believe is not a supported type? Seems to be corroborated by the "Unsupported argument type: ActiveRecord::Relation" error when trying to implement tests.

#1667

Oaphi:
tsck, seems like mail delivery test might need better isolation, CircleCI didn't like the merged version at all

Oaphi (approved):
LGTM

PR #1663, opened 2025-06-27
make confirmation email clearer to (try to) reduce false spam flags

cellio:
Added "Codidact" to email subject lines (because a lot of spam is generic like "confirmation instructions"), and expanded the confirmation message to include "ignore if this wasn't you" guidance in hopes of reducing incorrect spam reports.

I didn't review other message bodies (only subject lines) because new accounts are by far the most likely source of unexpected email to innocent bystanders.

### DIFF
--- a/app/views/devise/mailer/confirmation_instructions.html.erb
+++ b/app/views/devise/mailer/confirmation_instructions.html.erb
@@ -1,6 +1,6 @@
-<p>Welcome to Codidact, <%= @resource.username %>! We're glad you're here.</p>
+<p>Welcome to Codidact, <%= @resource.username %>! We're glad you're here.  (If this wasn't you, please ignore this email.)</p>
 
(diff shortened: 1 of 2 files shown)


Oaphi:
Don't we want to reduce instances of hardcoded "Codidact" strings, though? Suggestion to make use of the `NETWORK_NAME` env var ( usage: `ENV.fetch('NETWORK_NAME', 'Codidact')` ) and locale string parameters

cellio:
> Don't we want to avoid hardcoding in "Codidact" strings, though? Suggestion to make use of the `NETWORK_NAME` env var ( usage: `ENV.fetch('NETWORK_NAME', 'Codidact')` ) and [locale variables](https://guides.rubyonrails.org/i18n.html#passing-variables-to-translations)

That's a good point.   I'll try to address in a followup.

PR #1668, opened 2025-07-01
Self-serve account deletion

### LINKED ISSUE #925: Self-serve profile deletion
When https://meta.codidact.com/posts/286332 was asked, profile deletions were destructive hard deletes and we rightly wanted to build protections.  Now, we have soft-deletes for profiles, and we get requests from people to delete their profiles that require manual work including confirming identity, and we should make this easier.

Proposal: 

Add a "delete this profile" option somewhere in account settings.

Require the user to re-enter the password to proceed.

Show explanation about the effects of deletion.  Ask user to choose: this community only, all communities.

Confirmation step: "your profile will be deleted on (list of communities); your posts will remain and no longer be attributed to you", require approval.  

Soft-delete and log the event as an annotation.

Send email confirming the deletion, with instructions on how to undelete and a deadline (?).  To protect people from having other people undelete their accounts, I think undeletion should require both a key that we include in that email and the password.  If the first version of this feature doesn't implement undelete and that has to be manual, that's ok -- it's a rare case, but one we should handle at some point.

I don't know if we actually need a deadline.  If somebody who was in good standing comes back six months later and says "I'd like to return", why not?  We're never going to hard-delete the account, so it can stay soft-deleted forever, right?

ArtOfCode-:
Adds self-serve account deletion.

Closes #925.


ArtOfCode-:
Ready for review, but let's hold off this one until #1647 is merged.

cellio:
We don't allow admins or mods to self-delete (good), but the check is at the end.  Instead of sending the user through the sudo path (password, then type user name, then get the error), can we check when the person clicks the button?

Other than that, looks great!  I tested all the special statuses (local/global mod/admin), and confirmed that staff alone doesn't block deletion.  The deletions show up in the audit log as expected, and that email address/password no longer works to sign in.  A self-deleted user *can* sign up again with the same email address and gets a new account.

ArtOfCode-:
Good catch - I meant to just not show the button for special users, perhaps with a caption instead, but forgot to add that.

cellio:
Grayed out with an explanation would be consistent with other buttons that are selectively not available.  But any way you want to handle it (at the beginning) is fine.

cellio (approved):
LGTM.  Defer merge until after 1647 merges.

Oaphi (approved):
LGTM code-wise (have done some manual testing too, but not beyond getting to the sudo screen). 
Left a bunch of minor suggestions (mostly upkeep-related) as a separate PR with this one as a target

Issue #1184, opened 2023-08-18
Links in flag-escalation email go to meta not the correct site


cellio:
I received an email notice of a flag escalation on one of our communities and used the link in the email to (try to) review.  This failed, however, because the link went to meta.codidact.com while using the post number from the correct community.  The workaround is to edit the URL to replace "meta", but we should find and fix whatever's baking in the Meta assumption in case it affects other emailed links too.

If it's _just_ a problem with escalation email we can lower the priority (affects few people, has workaround).  I set it to medium pending that investigation.

PR #1681, opened 2025-07-14
Update & test flag mailer

### LINKED ISSUE #1680: 500 error when escalating flag
Image

ArtOfCode-:
Closes #1680.

Still not 100% sure why this error is happening - the method is valid and works elsewhere and in dev & test. For now I've updated a couple of little bits I found and added a test to hopefully catch this if it does happen.


Oaphi (changes requested):
Confirmed reason for the error is the presence of `post_history_path` route helper in the mailer view (those are not available in the context, `*_url` helpers should be used). Suggestions incoming, as usual (on how to test for it & what to change). 

RC'ing just so as we don't close the issue without addressing it first, LGTM otherwise

Suggestions PR: #1686

cellio (commented):
Tested and now working in my dev environment.  I also noticed that the previous one, that caused the error that led to the bug report, is now marked in the flag view as escalated, so that part apparently worked even though there had been an error.  Just mentioning in case that affects making a reproducer for tests.

Oaphi:
And one last thing to actually close #1184: #1687

Oaphi (approved):
LGTM (although another pair of eyes on this won't hurt, cc @cellio)

cellio (approved):

PR #1688, opened 2025-07-16
Show pings in thread titles & strip markdown

### LINKED ISSUE #1262: Thread titles with user pings show @numbers
meta:289223

If a thread title -- the first hundred or so characters of the comment, if there's no explicit title -- contains a user ping, that ping is properly translated into the user name in the _comment_, but not in the _thread title_.  If it's easy to also insert the name into the title, let's do that.  If not, let's do something else to make thread titles not ugly.  I'm not sure what "something else" should be; at the beginning of the line we could just strip it, but that would make a mess if a comment begins with something like "The answer by @1234 points out that...".   Before we worry about that: is it easy to show the name in the title and avoid the problem?

Oaphi:
closes #1262

---

This is how thread titles with pings look like now:

Screenshot from 2025-07-16 03-33-37

And this is how they will look like after the change (not married to the exact formatting):

Screenshot from 2025-07-16 03-33-17

The render reuses the `rtl_safe_username` user helper to show the actual name, so it's safe in terms of deleted usernames.

---

In addition to making comment thread titles render pings, this PR also makes sure markdown is stripped out from auto-generated titles (when the user doesn't provide one of their own):

The `generate_thread_title` helper reuses our `strip_markdown` helper but additionally strips out leading quotes (open to moving the change into the `strip_markdown` helper, but it'll have to be carefully reviewed if so)

### DIFF
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -41,15 +41,31 @@ module CommentsHelper
     ping[2..-1].to_i
   end
 
+  # Gets a list of pinged user IDs from a given content
+  # @param content [String] content to get pinged user IDs from
+  # @return [Array<String>] list of pinged user IDs
+  def pinged_user_ids(content)
+    content.scan(/@#\d+/).map { |ping| uid_from_ping(ping) }
+  end
+
+  # Gets a list of pinged users for a given content
+  # @param content [String] content to get pinged users from
+  # @return [Hash{String => User}] list of pinged users
+  def pinged_users(content)
+    user_ids = pinged_user_ids(content)
+    User.where(id: user_ids).to_a.to_h { |u| [u.id, u] }
+  end
+
   ##
-  # Process a comment and convert ping-strings (i.e. @#1234) into links.
-  # @param comment [String] The text of the comment to process.
-  # @param pingable [Array<Integer>, nil] A list of user IDs that should be pingable in this comment. Any user IDs not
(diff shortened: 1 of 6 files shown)


app/helpers/comments_helper.rb
```diff
@@ -40,6 +54,16 @@ def render_pings(comment, pingable: nil)
     end.html_safe
   end
 
+  # Converts all ping strings (i.e. @#1234) in content into usernames for use in text-only contexts
+  # @param content [String] content to convert ping strings for
+  # @return [String] processed content
+  def render_pings_text(content)
+    content.gsub(/@#\d+/) do |id|
+      user = User.where(id: id[2..-1].to_i).first
```
cellio, line 62: What does this do if a comment pings more than one user?  (I'm wondering about the `first` call here.)

Oaphi: `id[2..-1]` takes a slice from the ping string, which is `@#<numeric id>`, resulting in just a numeric id. Then, `User.where(id: <resulting numeric id>).first` ensures only one user is returned. This is then repeated for each ping string due to how `gsub` works.

Tbh, this is not very efficient, but improving upon the logic is out of scope for this PR - this is basically an exact copy of what's in `render_pings` above

cellio: Thanks for explaining, and I agree it's fine to do it this way here.

cellio (approved):

PR #1693, opened 2025-07-21
Audit logs: add filter/search parameters

ArtOfCode-:
Adds the ability to filter audit logs based on log type, event type, and date.


Oaphi (approved):
LGTM, that was sorely needed, actually!

PR #1695, opened 2025-07-22
don't allow self-deletion for an account that (still) uses 2fa

cellio:
Small followup to https://github.com/codidact/qpixel/pull/1668.  While testing something else I noticed that we should probably make it harder for a 2FA user to self-delete.  Rather than trying to check the second factor (complicated), I propose that we not allow self-deletion in this case.  If you use 2FA and want to delete your account, first disable 2FA (which will require a code) and then proceed.

We ought to do something similar for the SSO case, but I don't know how to set up an SSO environment to develop/test in.  I propose spinning that off as a separate issue for the future and making sure we document this gap in the release notes.

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -37,6 +37,9 @@ class Users::RegistrationsController < Devise::RegistrationsController
     elsif @user.moderator?
       @user.errors.add(:base, I18n.t('users.errors.no_mod_self_delete'))
       render :delete
+    elsif @user.enabled_2fa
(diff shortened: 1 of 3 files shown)


Oaphi:
Maube we should require passing MFA for users with it enabled instead? That's more complex to implement, but if a user is already at the point of self-deletion, we might as well not cause further frustration.

cellio:
> Maube we should require passing MFA for users with it enabled instead? That's more complex to implement, but if a user is already at the point of self-deletion, we might as well not cause further frustration.

I'm proposing the easy fix now because self-delete is already merged so 2fa self-delete is possible without the second factor.  If we decide to instead do the second-factor check as part of the deletion, we can improve that later.

Issue #1655, opened 2025-06-17
Code standards: line length for Markdown


trichoplax:
The [Code Standards document](https://github.com/codidact/qpixel/blob/develop/CODE-STANDARDS.md) does not specify a maximum line length for Markdown files such as readmes. It specifies different line lengths for different languages:
- CSS: 120
- HTML: 140
- JavaScript: 120

Should we specify a line length for Markdown too?

Is there a reason to have a longer line length for HTML than for CSS and JavaScript? If the same people are reading these languages on the same screen with the same font size, would it make sense to have the same limit for all of them, and also use that for Markdown?

trichoplax:
If there's a reason to prefer 140, I'd be happy with that too. My personal preference is for consistency - I don't mind whether it's 120 for everything, or 140 for everything.

trichoplax:
If it's decided to keep different limits for different languages, having a specified limit for Markdown will still be useful, regardless of what that limit it.

trichoplax:
With URLs, even 140 might sometimes be a difficult limit to keep to. The CSS section of the current Coding Standards says:

> Please limit line length to 120 characters or less.  
Note: this rule is not enforced for arguments to `url()` and other possible corner cases where developers have no control over the length of arguments.

Perhaps we could have a similar exception for URLs for HTML?

Whether we settle on 120 or 140 for HTML, there are already many lines longer than either, for both `.html` files (in the database seeds) and `.html.erb` files (in the views), so there is work to be done at some point whether we change the limit or not.

How does the following sound?
- Make an explicit exception for long URLs in any language (rather than just for CSS).
- Explicitly state that the HTML line length limit applies also to `.html.erb` files.
- Settle on either 120 or 140 as a consistent line length limit for all languages.
- State the line length limit and the languages it applies to (HTML, CSS, JS, Ruby, Markdown) in one place near the top of the Code Standards, rather than repeat it for each language.

My personal preference would be for the consistent limit to be 120 if we have an exception for long URLs. As an example, `app/views/layouts/_head.html.erb` has our CDN URLS, and all of them give lines shorter than 120 with the exception of the polyfill link, which is well over 140 so would need an exception to the rule either way.

Oaphi:
>  there any objection to changing the rule to 2 spaces for all languages? 

No objection, that's the de-facto standard

>  Ruby isn't mentioned

[...]

PR #1709, opened 2025-07-24
Add ERB linting to the project

### LINKED ISSUE #1699: Add HTML/ERB linting
We don't currently lint our HTML/ERB code. We should add a way of doing this to bring it in line with other languages.

Oaphi:
closes #1699

related: #1655


Oaphi:
About 200 out, 412 more to go

Oaphi:
Had to basically restart the PR in case anyone's wondering what are all those force-pushes about.
We have around ~570 true positives reported. This PR tends to fall behind quickly, so I am changing my approach here - it's going to just sit there and have updates from `develop` merged in. Any changes actually fixing the reported issues should be made as standalone PRs or as part of working on other issues.

Oaphi:
Collaboration welcome - if anyone is up to taking on a chunk or two, it'll be appreciated (just avoid simply running autocorrect - I can do that myself just fine, prefer harder problems reported). Please take note of PRs linked here so as not to duplicate work

PR #1711, opened 2025-07-24
Fix for in-page comment reply never enabling the submit button

### LINKED ISSUE #1710: In-page comment reply never enables the submit button
This has previously worked, but in testing today, I found that I could not submit a reply in a comment thread from the in-page view.  The counter changes from red when I reach the minimum length, but the "add reply" button never lights up.  I *can* create a new comment thread and reply on the thread page; it's only reply + in-page that is broken.

I tested at https://github.com/codidact/qpixel/commit/29b0054795bbd74d5afba905476d93d44b03c101 and it worked there.  This is the commit before the merge of https://github.com/codidact/qpixel/pull/1702.  I then tested the commit for that merge and it did not work.  There is form-validation javascript in that PR that I don't understand well enough to dig further.

After we either fix or revert, we should probably publish a minor version (or amend 0.12 if that's possible).

Oaphi:
Closes #1710

### DIFF
--- a/app/views/comments/_new_thread_modal.html.erb
+++ b/app/views/comments/_new_thread_modal.html.erb
@@ -31,8 +31,8 @@
                         required: true,
                         data: { post: post.id,
                                 thread: '-1',
-                                character_count: ".js-character-count-#{post.id}" } %>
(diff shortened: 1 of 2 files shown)


cellio (approved):
LGTM.  I tested and confirmed that it fixes both issues.

PR #1713, opened 2025-07-28
Use correct host when rendering comment pings

### LINKED ISSUE #1708: Embed host in rendered ping tags
As a follow up to #1184 , we should also embed `host` in the helper that renders pings - it's not directly related to the issue at hand (although can surface in flag escalation emails if comments contain rendered pings), but has the same root cause (`deliver_later` schedules a job in a new thread, which resets `RequestContext`).

Oaphi:
closes #1708

Where previously for any scheduled email (or anywhere where `RequestContext` is not populated or is reset for that matter) we had pings use the default host (usually `meta.codidact.com`), which is incorrect, now the correct host is used:

Also (unrelated, unfortunately, but setting up a new PR _just_ for that is too much hassle) stops loading our production status page in dev & test environments (see lines 47-50, the rest has been properly autoformatted [apart from line length])


cellio (approved):
LGTM.  Tested locally and works as advertised.

PR #1715, opened 2025-07-28
NamespacedEnvCache fix + pinned links caching fix

Oaphi:
As per investigation [in chat](https://discord.com/channels/634104110131445811/668890340199235613/1399396382908878929)

This PR fixes a long-standing bug with NamespacedEnvCache's `construct_ns_key` not accounting for the fact that keys can also be instances of `Array` or `Hash` (`ActiveSupport::Cache::Store` normalizes such keys under the hood in `expanded_key`).

Fixing the abovementioned issue exposed the bug in our widget for pinned (featured) links: since it's now `cache`d, `time_ago_in_words` is cached alongside the content, resulting in incorrect relative times for all timed pinned links (that is those that have `shown_before` or `shown_after` set), which this PR also fixes.

Finally, it makes `config.action_controller.perform_caching` and `config.log_level` for the development mode overridable via environment variables (`PERFORM_CACHING` and `LOG_LEVEL` respectively) - caching is sometimes needed (and due to the `perform_caching` being explicitly set to `false`, launching the server with the `-C` flag has no effect), and even the `:info` level can be too much. No changes were made to the defaults

### DIFF
--- a/app/models/pinned_link.rb
+++ b/app/models/pinned_link.rb
@@ -1,4 +1,10 @@
 class PinnedLink < ApplicationRecord
   include MaybeCommunityRelated
   belongs_to :post
+
+  # Is the link time-constrained?
+  # @return [Boolean] check result
+  def timed?
+    shown_before.present? || shown_after.present?
+  end
 end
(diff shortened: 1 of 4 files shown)


cellio (commented):
Looks good based on inspection; not sure how to test.

ArtOfCode- (approved):

PR #1719, opened 2025-07-31
Minor fix for the users page search button alignment

Oaphi:
Fixes the following misalignment accidentally caused by an override specific to audit log filtering:

With the fix applied (exactly how it currently looks in prod):

The audit log filter button override is still preserved:


cellio (approved):
LGTM based on testing.  Did not review CSS 'cause that stuff is still kind of magic to me. :-)

ArtOfCode- (approved):

PR #1720, opened 2025-08-01
Fix lack of title search for posts

### LINKED ISSUE #387: Exact search on a post title doesn't find the post

Oaphi:
closes #387

And no, we can't just add a separate index for the `title` column, the way we use search we need the composite index

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -2,12 +2,12 @@ class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
   def self.fuzzy_search(term, **cols)
-    sanitized = sanitize_for_search term, **cols
+    sanitized = sanitize_for_search(term, **cols)
     select(Arel.sql("`#{table_name}`.*, #{sanitized} AS search_score"))
   end
 
   def self.match_search(term, **cols)
-    sanitized = sanitize_for_search term, **cols
+    sanitized = sanitize_for_search(term, **cols)
     select(Arel.sql("`#{table_name}`.*, #{sanitized} AS search_score")).where(sanitized)
   end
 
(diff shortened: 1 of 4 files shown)


trichoplax:
Not a suggestion, just a question to gain understanding:

> And no, we can't just add a separate index for the `title` column, the way we use search we need the composite index

By "the way we use search" do you mean the way we implement it behind the scenes, or what we allow the user to do?

If it's just a matter of implementation, is there a benefit to having separate indexes for title and body, and returning results that match in either (the union of 2 searches)? My guess is that would mean less resources spent on storing and updating indexes, but a slightly slower search query? Is this a trade off between speed of search and speed of everything else? I don't know if either matters at present.

cellio:
Let's do this after the deploy that we're readying now.  (Reduce risk now, then do this in the next deploy.)

Oaphi:
Note: there's going to be a couple more changes (should've probably set it in the draft state for now)

Oaphi:
> By "the way we use search" do you mean the way we implement it behind the scenes, or what we allow the user to do?

@trichoplax mostly just noting that it was suggested to use a separate index in the previous discussion, but attempting to do so without changing how `sanitize_for_search` works completely will cause the "Can't find FULLTEXT index matching the column list" DB error. 

We can separate the index out later when (if?) we allow to explicitly search only in title / body (like how SE does with the `title:` operator)

PR #1728, opened 2025-08-02
Reduce email chunks

### LINKED ISSUE #1724: Emails to all users don't send because they have too many recipients
After our recent [changes to admin-all emails](https://github.com/codidact/qpixel/pull/1691), these emails should go out chunked. However, on trying again there's a new error in the logs: `recipient count is greater than 50`.

The chunks need adjusting to <50 recipients.

ArtOfCode-:
The AWS limit of 50 recipients was triggering because we were sending TO one address, BCC 50 others - 51 recipients.

Reduced chunk size to 49, which should solve the problem.

Closes #1724.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -65,7 +65,7 @@ class AdminController < ApplicationController
 
     Thread.new do
       emails = User.where.not(confirmed_at: nil).where('email NOT LIKE ?', '%localhost').select(:email).map(&:email)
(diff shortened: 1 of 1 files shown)


Oaphi (approved):
For reference, this is the SES quota in question (we use 1 + N emails in BCC, hence the adjusted limit):

> 50 recipients per message.
> Note A recipient is any "To", "CC", or "BCC" address.

PR #1736, opened 2025-08-03
Fix for the "all events" filter tab on the user activity page

### LINKED ISSUE #1735: "Show all events" filter tab on the user activity page is not visually selectable
Due to an issue with handling the default "no filter" option, the "show all events" filter tab on the user activity page is never highlighted. The filtering is applied properly, so this is just a styling issue.

Oaphi:
closes #1735


cellio (approved):
LGTM.  Tested locally.

PR #1737, opened 2025-08-04
0valt/1732/close reasons

### LINKED ISSUE #1732: None of the custom close reason fields are validated
We do not have any validation for custom close reasons, which makes it possible to create empty reasons:

Image

The resulting reason looks quite bizarre in the list:

Image

More importantly, since we have no validations, another bug has gone unnoticed - that we silently swallow close reason creation errors

Oaphi:
closes #1732 

This PR ensures close reason names are validated for both minimum & maximum length (the latter is just so as crafty users aren't "rewarded" with a server error).

It also fixes lack of visible validation errors when trying to _update_ (we've fixed the create action earlier) an invalid close reason:

Screenshot from 2025-08-04 12-11-55

### DIFF
--- a/app/controllers/close_reasons_controller.rb
+++ b/app/controllers/close_reasons_controller.rb
@@ -30,7 +30,7 @@ class CloseReasonsController < ApplicationController
         redirect_to close_reasons_path
       end
     else
-      render :edit
+      render :edit, status: :bad_request
     end
   end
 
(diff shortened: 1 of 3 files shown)


ArtOfCode- (approved):

Oaphi:
> Approved but has conflicts.

Probably due to #1733 - will solve shortly

PR #1743, opened 2025-08-04
0valt/1112/search title pos

### LINKED ISSUE #1112: On search page, filter is above post-list header instead of below it
On category and profile post lists, the order is: count + buttons, then filter, then list, like this:

screenshot from test server

However, on the search page, these elements are reversed:
screenshot

This isn't the simple matter of moving a `div` on `app/views/search/search.html.erb` like I thought it would be, because the filter is baked into the "widget" that includes the search textbox and the advanced help.  So we'll need to do a little refactoring in order to be able to put the filter below the count/buttons.  I'm not sure why it was done this way and thus what the consequences would be of moving that code, so instead of fiddling with it, I'm writing this issue to look into later.

Oaphi:
closes #1112 

Filters are now properly aligned (note to readers: please do not put forms in reusable partials unless they are supposed to be fully self-contained):

Screenshot from 2025-08-04 23-18-26

### DIFF
--- a/app/assets/stylesheets/search.scss
+++ b/app/assets/stylesheets/search.scss
@@ -1,0 +1,16 @@
+@import 'variables';
+
+.search-filters:not([open]) {
+  border-top: 1px solid $muted-graphic;
+}
+
+.search-sorting {
+  align-items: center;
+  display: flex;
+  justify-content: space-between;
+
+  .search-sorting-types {
+    margin: 0.5em;
+    margin-right: 0;
+  }
+}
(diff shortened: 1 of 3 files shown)


ArtOfCode-:
> note to readers: please do not put forms in reusable partials unless they are supposed to be fully self-contained

Okay, now I'm curious.

ArtOfCode- (approved):

cellio (approved):
I didn't review the code (I see someone else did).  The behavior looks right (thanks!).

Oaphi:
> Okay, now I'm curious.

To elaborate a little, @ArtOfCode-, it's what made a trivial change a multi-file diff - because the `_widget` partial contained the whole form, it wasn't exactly easy to just change the position of the filters. As an example of the contrary, we do this correctly in the `categories/show` view by wrapping the filters partial in a form as needed:

```erb
  <%= form_tag request.original_url, method: :get do %>
    <%= render 'search/filters' %>
  <% end %>
```

PR #1746, opened 2025-08-04
Allow dev mode to impersonate users directly

Oaphi:
In the development environment, requiring the user to always submit a non-empty reason serves no purpose but is enough of an annoyance that it adds up over time, especially when manually testing access control-related changes - this PR allows dev mode to just go straight to impersonating a desired user:

https://github.com/user-attachments/assets/f23f1460-2c11-4c28-ab76-4014a60761c5

Plus it also cleans up the controller's code a little bit.


cellio:
While we're at it, can we remove the password requirement when *ending* impersonation?  (I would say always though I don't know why it's there, but at *least* in dev mode would sure be handy.)

Oaphi:
Probably best to do in a standalone PR, @cellio, we wanted to connect "sudo mode" to it

cellio:
> Probably best to do in a standalone PR, @cellio, we wanted to connect "sudo mode" to it

Whoops, sorry, edited my comment after our discussion and hadn't noticed your new one. :-)

PR #1751, opened 2025-08-05
fixed lack of visible error when trying to rename a tag to an invalid…

### LINKED ISSUE #545: Renaming a tag to an existing one fails silently
https://software.codidact.com/posts/282536

We shouldn't allow renaming a tag to an existing one, of course -- that's usually an accident.  Let's catch and report the error instead of leaving users who don't know to look for console errors in the dark.

Oaphi:
closes #545

Failing to rename a tag for some reason will now show a proper error notification:

Screenshot from 2025-08-05 05-15-42

We should improve the error message itself, but it's how we do error handling with many other AJAX-based actions (see, for example, nominating a post for promotion), so I'd rather address that in a centralized manner.

### DIFF
--- a/app/assets/javascripts/tags.js
+++ b/app/assets/javascripts/tags.js
@@ -202,7 +202,7 @@ $(() => {
         location.reload();
       }
       else {
-        console.error('Failed to rename tag, somehow');
+        QPixel.createNotification('danger', `Failed to rename the tag. (${resp.status})`);
       }
     }
   });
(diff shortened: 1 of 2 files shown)


cellio (commented):
Tested and works for me (and renaming to something new still works too :-) ).  I take it we can't, at the point where we're generating the message, tell why it failed?  This is fine if so; it's sure an improvement over a console message... Just wondering what it would take to say "can't rename to an existing tag" or something.

Oaphi:
>  I take it we can't, at the point where we're generating the message, tell why it failed?

We can, but I'd rather work on it in a separate PR that also addresses other issues (I'd like us to standardize on the JSON format for such failures - presently, we are quite inconsistent)

Issue #1760, opened 2025-08-07
Tell users if parts of their search query are incorrect


Oaphi:
Split off from #1128

PR #1761, opened 2025-08-07
Search fixes & improvements

### LINKED ISSUE #1128: Can't search for net negative votes, and trying it returns wrong results
https://meta.codidact.com/posts/288180

It should be possible to search for net negative votes.  For example, searching for `votes:-2` should match posts with +2/-4, +10/-12, +0/-2, etc.  But this doesn't actually work, and it instead returns all posts that otherwise match the search, i.e. this clause is ignored.

Two related issues, then:

1. Can we allow searching for net negative votes?
2. When part of a search is rejected for some reason, can we find a way to tell the user?  (This also came up when someone tried to search for tags using SE's notation instead of ours -- it produced matches based on the text not the tag.)

If 2 is hard then let's split it off.  Can we at least do 1?

Oaphi:
closes #1128

This PR ensures `votes:` operator can use negative values:

Screenshot from 2025-08-07 03-30-04

Does a general cleanup of qualifier parsing as well in preparation for #1760

### DIFF
--- a/test/helpers/search_qualifier_helper_test.rb
+++ b/test/helpers/search_qualifier_helper_test.rb
@@ -91,30 +91,26 @@ class SearchQualifierHelperTest < ActionView::TestCase
 
   test 'parse_category_qualifier should correctly parse category:' do
     ['=', '>', '<', '<='].each do |operator|
-      categories.map do |cat|
-        [operator == '=' ? cat.id.to_s : "#{operator}#{cat.id}", operator, cat.id]
-      end.each do |entry|
-        value, expected_op, expected_val = entry
+      categories.each do |cat|
+        value = operator == '=' ? cat.id.to_s : "#{operator}#{cat.id}"
 
         parsed = parse_category_qualifier(value)
         assert_equal :category, parsed[:param]
-        assert_equal expected_op, parsed[:operator]
-        assert_equal expected_val, parsed[:category_id]
+        assert_equal operator, parsed[:operator]
+        assert_equal cat.id, parsed[:category_id]
       end
     end
   end
 
   test 'parse_post_type_qualifier should correctly parse post_type:' do
     ['=', '>', '<', '<='].each do |operator|
-      post_types.map do |type|
-        [operator == '=' ? type.id.to_s : "#{operator}#{type.id}", operator, type.id]
-      end.each do |entry|
-        value, expected_op, expected_val = entry
+      post_types.each do |type|
+        value = operator == '=' ? type.id.to_s : "#{operator}#{type.id}"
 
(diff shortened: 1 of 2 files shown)


test/helpers/search_qualifier_helper_test.rb
```diff
@@ -0,0 +1,124 @@
...
+      assert_not matches_numeric?(value)
+    end
+  end
+
+  test 'matches_non_negative_int? should correctly match positive ints' do
+    ['123', '>=42', '9999', '0'].each do |value|
+      assert matches_non_negative_int?(value)
+    end
+
+    ['name', '-tag', '3.14'].each do |value|
```
cellio, line 39: add a negative number here (to confirm that it doesn't pick it up)?

test/helpers/search_qualifier_helper_test.rb
```diff
@@ -0,0 +1,124 @@
...
+    expected.each do |input, expect|
+      assert_equal expect, numeric_value_sql(input)
+    end
+  end
+
+  test 'date_value_sql should return correct operator, value, and timeframe' do
+    expected = {
+      '1' => ['', '1', 'MONTH'],
+      '1y' => ['', '1', 'YEAR'],
+      '<1y' => ['>', '1', 'YEAR'],
```
cellio, line 69: `<1y` should produce `>`?  I hope this test failed?  (And similar for the next line.)

test/helpers/search_qualifier_helper_test.rb
```diff
@@ -0,0 +1,124 @@
...
+      '>=2w' => ['<=', '2', 'WEEK']
+    }
+    expected.each do |input, expect|
+      assert_equal expect, date_value_sql(input)
+    end
+  end
+
+  test 'parse_answers_qualifier should correclty parse answers:' do
```
cellio, line 77: nit: "correclty"

cellio (commented):
I tested the behavior and reviewed the test cases.  I didn't review the implementation code.

test/helpers/search_qualifier_helper_test.rb
ArtOfCode-: This is correct, IIRC - a user-facing value of "<1y" ("less than one year ago") translates to an SQL query of "created_at > (one year ago)" - the operators switch.

Oaphi: After taking a quick look, yeah, what @ArtOfCode- said - the test correctly checks that the operator is inverted

cellio: Oh, I didn't realize it was just a relocation.  Thanks both of you for explaining.

PR #1764, opened 2025-08-07
Revision reference fix for rejected suggested edits

### LINKED ISSUE #1371: Suggested edit compares to the latest version, not the version that was edited
meta:291061

A suggested edit shows a diff.  This diff compares the edit to the latest version of the post, which is always correct for an _active_ suggested edit, but is not correct for one that has been resolved.

Example: [this edit](https://meta.codidact.com/posts/suggested-edit/1387) was proposed to revision 1 in [this history](https://meta.codidact.com/posts/291059/history), at 2024-03-13T00:28:17Z.  That suggested edit was rejected, and then the post was edited half an hour later (rev 2) and twice more after that.  The diff in the suggested edit compares the edit to the latest version.  It should compare it to rev 1, where the edit was proposed.

An unclear diff like this can appear to put changes in a user's mouth, so to speak, and unclear history makes review and moderation more challenging.

Oaphi:
closes #1371

I also added a maintenance task that should fix up the existing rejected suggested edits - we can run at any time we want after deployment (I've tested on the cases that I have available, but it's not enough to safely run in prod - let's run it on dev servers first when the time comes).

### DIFF
--- a/app/controllers/suggested_edit_controller.rb
+++ b/app/controllers/suggested_edit_controller.rb
@@ -42,11 +42,7 @@ class SuggestedEditController < ApplicationController
              before_tags: @post.tags.to_a,
              after_tags: @edit.tags }
 
-    before = { before_body: @post.body,
-               before_body_markdown: @post.body_markdown,
-               before_tags_cache: @post.tags_cache.dup,
-               before_tags: @post.tags.to_a,
-               before_title: @post.title }
+    before = before_attributes(@post)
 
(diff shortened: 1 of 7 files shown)


cellio (approved):
LGTM.  I tested the behavior of suggested edits but did not run the task locally.  I looked at the test code and it looks sound, except I don't understand concerns so no comment there.  Someone else should review the implementation code (looks fine to me but we want a stronger review).

PR #1768, opened 2025-08-08
Fix for modals not reacting to the Esc key & for file input field not gaining focus

### LINKED ISSUE #1366: User page tabs with no sidebar have a JS error in the developer console
**Describe the bug**
Those tabs on the user profile page that do not include the right hand side bar show the following JavaScript error message in the browser developer console:

```text
Uncaught (in promise) TypeError: prefs.global is undefined
```

**To Reproduce**
1. Go to your own user profile page (click on your avatar at the top right of any page).
2. Click on any tab that does not have a sidebar on the right hand side.
3. See error message in the developer console.

**Expected behavior**
The behaviour to a user appears to be working, but presumably is not applying the user's preferences as they cannot be accessed. It is expected that the user's preferences would be applied to all tabs.

Oaphi:
closes #1366

2 fixes:
- modals now react to the <kbd>Esc</kbd> key being presses as expected (they should close);
- file input field is now focused when the image upload modal is opened;

https://github.com/user-attachments/assets/bd9d61eb-52e6-4a78-8868-9ce485ff0a9b

### DIFF
--- a/app/assets/javascripts/modals.js
+++ b/app/assets/javascripts/modals.js
@@ -1,7 +1,9 @@
 document.addEventListener('DOMContentLoaded', () => {
-  document.addEventListener('keypress', (ev) => {
-    if (ev.code === 'Escape') {
-      document.querySelectorAll('.modal').forEach((el) => el.classList.remove('is-active'));
+  document.addEventListener('keyup', (ev) => {
+    if (ev.code === 'Escape' && !ev.metaKey && !ev.ctrlKey) {
+      document.querySelectorAll('.modal').forEach((el) => {
+        el.classList.remove('is-active');
+      });
     }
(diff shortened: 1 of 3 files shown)


cellio (commented):
On a post draft, I tested the image and link buttons.  The image file has focus (so you don't need the mouse), and escape closes both modals.  If you didn't mean to close a link modal, I see that your previous typing is still there if you choose it again, and only on that post (other posts start fresh) -- all expected behavior.

I'm not sure what other modals are affected.  I didn't observe any differences with close or flag operations.  I suppose "modal" means "thing that seems to pop over your screen and take control"?  I can't think of others of those off hand; happy to test if someone reminds me where.

Issue #714, opened 2021-10-26
Allow thread titles to be edited (with public record)


cellio:
https://meta.codidact.com/posts/282459
Also https://meta.codidact.com/posts/290349

When this came up again in a duplicate (https://meta.codidact.com/posts/284483), I suggested:

> I think it's reasonable for the person who created the thread to be able to edit the title so long as the only comments are from that user. As soon as there's a comment from someone else, though, there's the chance that changing the thread title can affect context for those comments (or even can be changed destructively), and -- unlike with posts -- readers can't audit these edits.

At this writing that suggestion is +6/-0.  People make typos in thread titles, or don't realize that the default title is an initial substring.  We should let them edit so long as they're only affecting their *own* comments.

We should log the edit, just in case thread titles are used for abusive purposes and we need to be able to review.

**Update:** allow anybody involved in the thread to rename it, and add an auto-comment to the thread recording the change: something like "title changed from (old) to (new) by (user)". This comment is owned by System, not the person making the change, so only mods can delete it.

Curators have the "tools" menu, which other users don't have.  This means either giving them the "tools" menu (with just the one option) or adding an 'edit title" link somewhere.

cellio:
Public logging: add a comment to the thread of the form "title changed from (old) to (new) by (user)".  This comment is owned by System, not the person making the change, so only mods can delete it.

PR #1773, opened 2025-08-12
Thread title validation

### LINKED ISSUE #1378: prevent validation errors when creating too-short new comment threads
meta:291094

Comments have a minimum length before the "post" button is enabled.  This check includes leading/trailing whitespace, but such whitespace is trimmed on submission and that can make the comment now too short.  In that case, you get either a red-box error or a 422 (with your comment gone), which is not good.

From the user perspective the best thing to do would be to apply that "trim" logic when checking to see whether to allow posting.  I don't know if that's expensive (since it has to be checked frequently as you type).

If that's not practical, then can we catch the problem in-page, so you get a message saying "too short after all, sorry" and your draft is still there so you can keep editing?

### LINKED ISSUE #1731: Thread title has no minimum or maximum length (including allowing zero length)
**Describe the bug**
When creating a comment thread, the length of the title is not constrained to either the minimum or maximum length imposed on comments. It has always been possible to create a single character title explicitly. It is also possible to create a zero character title implicitly now that quote blocks are stripped from automatically generated titles.

**To Reproduce**
1. Start to create a new comment thread under a post
2. Leave the title field empty
3. Make the body of the comment a quote block only
4. Post the comment
5. See no title for the newly created thread, because the title is zero length

**Expected behavior**
A minimum and maximum length should be imposed on an explicitly created thread title. A decision needs to be made on whether this should be the comment length limits, or separate limits for titles.

An implicitly created thread title (automatically generated when a user leaves the title field blank) should be similarly constrained. A decision needs to be made on what to do when the comment meets the length requirements, but omitting aspects such as a quote block results in an automatically generated title that is below the minimum (possibly zero characters).

If a distinct set of limits is imposed for titles (rather than using the comment limits), there also needs to be a decision on what to do if the title minimum length is set to be more than the comment minimum length, or this needs to be prevented when setting the limits.

Oaphi:
closes #1731
closes #1378

This PR:

- introduces validation for comment thread titles (minimum 1 char for now, maximum is configurable via a new `MaxThreadTitleLength` site setting, defaults to 255. Should not be set higher than that - we don't support it in the DB [on that note, we should introduce validation for such settings - it's technically possible to allow higher values]);
- adds missing character counter to the rename thread modal:
    
- makes character counters aware of markdown stripping if enabled (the field should set `markdown` data-attribute to `strip`) - to prevent confusing users, when that happens, the counter displays an info icon with a title explaining the situation:
    
- also switched character counters to ignore leading & trailing white spaces as we don't count those towards the limit in the first place (note that we also had this problem with posts and pretty much any other text fields);
- also ensures that at least a generic error message is shown when failing to rename threads (providing a proper list of errors is outside of the scope of this PR and should be done when addressing #714 );

### DIFF
--- a/app/assets/javascripts/character_count.js
+++ b/app/assets/javascripts/character_count.js
@@ -67,7 +67,7 @@ $(() => {
 
     const fullCount = $tgt.val().trim().length;
     const count = omitMarkdown ?
-                    QPixel.MD.stripMarkdown($tgt.val().trim(), { removeQuotes: true }).length :
+                    QPixel.MD.stripMarkdown($tgt.val().trim(), { removeLeadingQuote: true }).length :
                     fullCount;
 
     $info.toggleClass('hide', fullCount === count)
--- a/app/assets/javascripts/character_count.js
+++ b/app/assets/javascripts/character_count.js
@@ -65,9 +65,9 @@ $(() => {
     const $info = $counter.find('.js-character-count__info');
     const omitMarkdown = $tgt.data('markdown') === 'strip';
 
-    const fullCount = $tgt.val().length;
+    const fullCount = $tgt.val().trim().length;
     const count = omitMarkdown ?
-                    QPixel.MD.stripMarkdown($tgt.val(), { removeQuotes: true }).length :
+                    QPixel.MD.stripMarkdown($tgt.val().trim(), { removeQuotes: true }).length :
                     fullCount;
 
     $info.toggleClass('hide', fullCount === count)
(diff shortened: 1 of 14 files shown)


ArtOfCode- (approved):

cellio (commented):
LGTM with one small doc change.  I tested the assertions in the PR description and confirmed that everything works as described.  Nice!

cellio:
I don't see that "title can't be blank" error message in the diff here, so I guess that's pre-existing, but now that we strip markdown, people could hit that when they didn't even enter a title.  Could we amend that message to "title can't be blank after stripping markdown"?  That might not always be the reason you got the error, but it's at least a hint for what might be the most common case.

ArtOfCode- (approved):

PR #1779, opened 2025-08-13
Fix for the homepage runtime error when no categories are configured as such

### LINKED ISSUE #794: Puma caught this error: undefined method `id' for nil:NilClass (NoMethodError) - Home Page
Puma caught this error: undefined method `id' for nil:NilClass (NoMethodError)
/code/app/controllers/categories_controller.rb:171:in `update_last_visit'
/code/app/controllers/categories_controller.rb:24:in `homepage'

Oaphi:
closes #794

Technically not possible unless the user setting up QPixel never runs the initial seeds as we set the `Q&A` category as homepage _or_ unless an admin unchecks the "Set as homepage?" checkbox later. The former is unlikely, but the latter is very much plausible - we now redirect to the list of categories in such a case.

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -21,6 +21,12 @@ class CategoriesController < ApplicationController
 
   def homepage
     @category = Category.where(is_homepage: true).first
+
+    unless @category.present?
+      redirect_to categories_path
+      return
+    end
+
     update_last_visit(@category)
     set_list_posts
     render :show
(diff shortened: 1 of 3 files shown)


cellio (approved):
TIL that we allow that box to be unchecked on all categories. :-)  So naturally I wondered: if there can be zero, can there be two?  Tested, and no -- checking the box revokes it from anywhere else it's set, so there can be only one.  Good.

Redirecting to the category list is a nice solution.  LGTM.

PR #1780, opened 2025-08-13
Fix user deletions

### LINKED ISSUE #1772: Deleting users breaks when profile is too long
Examples: https://software.codidact.com/users/110610/mod, https://scientific-speculation.codidact.com/users/107458/mod

These profiles can't be deleted, returning a 500. The error logs show that the 500 is because of a MySQL error: `ActiveRecord::ValueTooLong: Mysql2::Error: Data too long for column 'comment' at row 1`

I'm assuming that this is because these users have long profiles, which cause the audit log comment to be too long when we try to stuff the user attributes into it.

ArtOfCode-:
Limits attribute lengths to 100 when creating audit logs for deletions.

Also removes the ability to destroy users to prevent confusion over what option to use.

Closes #1772.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,4 @@ node_modules
 
 # YARD data
 .yardoc
+doc/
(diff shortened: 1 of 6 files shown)


Oaphi (approved):
LGTM

Issue #1754, opened 2025-08-05
Improve handling of failed AJAX request responses


Oaphi:
Currently, our handling of failed AJAX request responses is all over the place: sometimes we fail silently and log to console; sometimes we create a notification with the HTTP status code; sometimes we rely on the `message` the server sends in the response payload.

Even when we do rely on the `message` field, oftentimes the server does not actually send one. Oh, and also sometimes we rely on the `error` field instead.

This issue is to track our efforts for streamlining our handling of failed AJAX requests:

- [ ] flag escalation (hardcoded);
- [ ] user soft deletion (jQuery artifact);
- [ ] category change (hardcoded + bare HTTP status code);
- [ ] category post type update (hardcoded + bare HTTP status code);
- [ ] category post type deletion (hardcoded + bare HTTP status code);
- [ ] question closure (hardcoded + bare HTTP status code OR hardcoded + server-sent `message`);
- [ ] flagging (hardcoded + server-sent `message`);
- [ ] flag escalation (hardcoded);
- [ ] flag handling (hardcoded + bare HTTP status code OR hardcoded + server-sent `message`);
- [ ] post image upload (hardcoded OR server-sent `error` OR hardcoded + bare HTTP status code);
- [ ] suggested edit handling (hardcoded + server-sent `message`);
- [x] tag renaming (hardcoded + bare HTTP status code);
- [ ] user ability / role granting / revoking (hardcoded + server-sent `message`);
- [x] voting (hardcoded + server-sent `message` + bare HTTP status code);
- [ ] Judaism-specific: suggesting reference (hardcoded);
- [x] saving user preferences (console-only);

I've already added a centralized `handleJSONResponse` API for error handling, we just need to connect it & align server responses with it.

PR #1782, opened 2025-08-13
Fix for required tags duplication & improvements to tag renaming

### LINKED ISSUE #939: When editing tags, clicking on a required tag adds it again
https://meta.codidact.com/posts/287037

On metas, a post must have one of four tags, and they are shown above the tags box.  You can click on one to add it.  You can do that even if the post already has that tag.  I started to create a new post and was able to add `bug` twice.

This is an issue specifically with clicking one of the tags.  If I type the tag names, I can only add a tag once.  Therefore the editor has logic to check; we need to have the "add by clicking" path do the same check that the "add by typing" path does.

The post seems to *record* the tag only once; the linked post has one `bug` tag and the edit history doesn't show a second one being removed.  (I thought it did duplicate; maybe it did at one time.)  So this seems to be an editor-only bug.

Oaphi:
closes #939 

Given that the fix is extremely minor, I've thrown in a couple of improvements to tag renaming, specifically, aligned the JSON response structure with the other methods (part of #1754) + the failure message is now server-side. 

Note to reviewers: new locale string files are only picked up by Rails servers only after reload

### DIFF
--- a/app/assets/javascripts/qpixel_api.js
+++ b/app/assets/javascripts/qpixel_api.js
@@ -494,5 +494,15 @@ window.QPixel = {
     const data = await resp.json();
 
     return data;
+  },
+
+  renameTag: async (categoryId, tagId, name) => {
+    const resp = await QPixel.fetchJSON(`/categories/${categoryId}/tags/${tagId}/rename`, { name }, {
+      headers: { 'Accept': 'application/json' }
+    });
+
+    const data = await resp.json();
+
+    return data;
   }
 };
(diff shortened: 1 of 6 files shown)


ArtOfCode- (approved):

cellio (approved):
LGTM.  Tags no longer duplicate (for either edits or suggested edits).  It looks like the other changes are internal, or at least I couldn't figure out a user-visible change that I could test manually.

PR #1789, opened 2025-08-15
Fix for new comment thread title character count behavior

### LINKED ISSUE #1787: Incorrect disabling of Create thread button
meta:294485

> When starting a new comment thread, the "Create thread" button is enabled whenever I type or remove a character in the "Your comment" field, but disabled whenever I type or remove a character in the "Comment thread title" field.

Start new comment thread form, with characters being alternately added and removed in the comment and title fields

See [discussion on Discord](https://discord.com/channels/634104110131445811/668890340199235613/1405696955560103946) suggesting this may be due to having multiple identical forms on the same page. I have since tested this on questions with no answers, and the problem does not occur there, which supports the suggested cause.

I don't know why this only sometimes causes the problem though. The Meta question gives an example where one of the answers does not have the problem, and something about opening the form under that answer or interacting with it causes the problem to stop happening for other forms on the same page.

> Specifically, the problem does not affect the new comment thread form below [this answer by another user](https://meta.codidact.com/posts/289322/293600#answer-293600), and interacting with that new comment thread form prevents the problem from affecting other new comment thread forms on the same page that are open at the same time. Opening multiple forms without opening one on that answer does not prevent the problem. For example, if I open the form below my question and one of my answers, they both still have the problem. Only the linked answer prevents the problem, and it prevents it on all other new comment thread forms on the same page, whether they are already open, or opened after it.

Oaphi:
closes #1787

### DIFF
--- a/app/views/comments/_new_thread_modal.html.erb
+++ b/app/views/comments/_new_thread_modal.html.erb
@@ -43,10 +43,10 @@
                          class: 'form-element js-thread-title-field',
                          data: { post: post.id,
                                  thread: '-1',
-                                 character_count: '.js-character-count-thread-title',
+                                 character_count: ".js-character-count-thread-title-#{post.id}",
                                  markdown: 'strip' } %>
       <%= render 'shared/char_count',
-                 type: 'thread-title',
+                 type: "thread-title-#{post.id}",
                  min: 1,
                  max: maximum_thread_title_length %>
 


cellio (approved):

trichoplax (changes requested):
The problem with a different form being checked is fixed, but another problem has been introduced.

To reproduce:
1. Click "Start new comment thread".
2. Enter sufficient characters in the "Your comment" field.
3. Notice that the "Create thread" button is enabled (as expected).
4. Enter a single character in the "Comment thread title" field.
5. Notice that the "Create thread" button remains enabled (as expected).
6. Remove the single character from the "Comment thread title" field.
7. Notice that the "Create thread" button is disabled.

I would expect that the title would only disable the create button when it has a non-zero number of characters below the minimum. Zero characters should not disable the create button because the title is optional. Since the minimum is hardcoded to 1 this just means I would never expect the create button to be affected by the title minimum. This would be different if in future we made the minimum a setting.

I can't see why this problem isn't already present in production. The minimum length is already set to 1. However, in production the empty title field shows "0 / 255" whereas in this branch the empty title field shows "0 / 1". This happens whether the new comment thread is under the first, last, or only post on the page.

I'd usually raise this as a separate bug, but in this case the pull request will introduce the problem to `develop`, so it seems best to fix the problem either in this pull request or in a pull request that is merged before this one.

trichoplax:
For reference, the new bug has been raised as #1799.

PR #1793, opened 2025-08-15
Rename system test to follow convention of being plural

trichoplax:
I'd previously renamed the PostTest system test to PostSystemTest to avoid an error when running tests and systems tests together (since the non-system tests also have a PostTest). However, I've since discovered that the convention for avoiding this clash is to have non-system tests named with singular nouns, and system tests named with plural nouns.

Updating to reflect this allows removing my previous explanatory comment, which is no longer required now that convention is being followed.

### DIFF
--- a/test/system/posts_test.rb
+++ b/test/system/posts_test.rb
@@ -1,8 +1,6 @@
 require 'application_system_test_case'
 
-# Renamed from PostTest to PostSystemTest to avoid clash with test/models/post_test.rb
-# when running rails test:all
-class PostSystemTest < ApplicationSystemTestCase
+class PostsTest < ApplicationSystemTestCase
   # -------------------------------------------------------
   # Create
   # -------------------------------------------------------


trichoplax:
> Non-blocking: can you point to where the naming convention is described, I don't recall there being one, so would like to read up on it?

I meant convention in the informal sense of what is common rather than a written rule.

I couldn't find an explicit statement that this is convention, but it was hinted at in a question about what to do when the singular and plural are the same: [How to name tests for uncountable models](https://discuss.rubyonrails.org/t/how-to-name-tests-for-uncountable-models-controllers/89570) (which is what led me to make the change).

Before making the change I checked the test guide, and it did seem that [non-system tests use singular](https://guides.rubyonrails.org/testing.html#writing-your-first-test) and [system tests use plural](https://guides.rubyonrails.org/testing.html#system-testing), but that's not a strong convention, just what the examples suggest. The examples only show the user choosing singular and plural, not Rails imposing it or even defaulting to it for system tests.

Oaphi:
> Before making the change I checked the test guide, and it did seem that [non-system tests use singular](https://guides.rubyonrails.org/testing.html#writing-your-first-test) and [system tests use plural](https://guides.rubyonrails.org/testing.html#system-testing), but that's not a strong convention, just what the examples suggest

Good enough for me, thanks

Oaphi (approved):
LGTM

PR #1796, opened 2025-08-16
Separate post unfollow to avoid reversal on accidental second call

trichoplax:
Unlike following and unfollowing comment threads, which are already separate routes, following new threads on a post was using a single route that toggled the state, which could cause unintended results if accidentally called multiple times. This pull request splits it into a follow route and an unfollow route, so when the user clicks they can only get the advertised result.

AKA [idempotence](https://en.wikipedia.org/wiki/Idempotence).

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -276,11 +276,13 @@ class CommentsController < ApplicationController
 
   def post_follow
     @post = Post.find(params[:post_id])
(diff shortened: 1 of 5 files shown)


Oaphi (commented):
Question: with this changes, what will happen if a user tries to follow a post they already follow?

Oaphi (approved):
LGTM

PR #1800, opened 2025-08-17
Remove min length attr from the new thread title field

### LINKED ISSUE #1799: New thread minimum title length prevents zero length
**Describe the bug**
Entering a zero length title disables the "Create thread" button, even though the title is marked as optional.

**To Reproduce**
1. Click "Start new comment thread".
2. Enter sufficient characters in the "Your comment" field.
3. Notice that the "Create thread" button is enabled (as expected).
4. Enter a single character in the "Comment thread title" field.
5. Notice that the "Create thread" button remains enabled (as expected).
6. Remove the single character from the "Comment thread title" field.
7. Notice that the "Create thread" button is disabled.

**Expected behavior**
I would expect that the title would only disable the create button when it has a non-zero number of characters below the minimum. Zero characters should not disable the create button because the title is optional. Since the minimum is hardcoded to 1 this just means I would never expect the create button to be affected by the title minimum. This would be different if in future we made the minimum a setting.

**Additional context**
This bug has not yet reached production, so I recommend delaying deploying `develop` until this bug has been fixed.

Oaphi:
closes #1799 

Overfixed #1731 a bit - `min` is not needed on the create new thread field

As it's an extremely minor fix, the rest of the PR is just a bunch of fixes for the comment docs for our views.
In case someone wonders what's that all about, this is how our comments look otherwise (and the rest of the views' highlighting breaks as well):

image

As opposed to how it's expected to look like:

image

### DIFF
--- a/app/views/comments/_new_thread_modal.html.erb
+++ b/app/views/comments/_new_thread_modal.html.erb
@@ -47,7 +47,6 @@
                                  markdown: 'strip' } %>
       <%= render 'shared/char_count',
                  type: "thread-title-#{post.id}",
-                 min: 1,
                  max: maximum_thread_title_length %>
 
       <%= submit_tag 'Create thread',
--- a/app/views/comments/_new_thread.html.erb
+++ b/app/views/comments/_new_thread.html.erb
@@ -1,11 +1,11 @@
 <%#
-  Renders new comment thread button & modal
+   "Renders new comment thread button & modal
 
-  Variables:
-    post : post to create a new thread for
-    text : text to display on the button
-    user : user that will create a new thread
-%>
+    Variables:
+      post : post to create a new thread for
+      text : text to display on the button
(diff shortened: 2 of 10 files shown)


app/views/comments/_new_thread_modal.html.erb
```diff
@@ -47,7 +47,6 @@
                                  markdown: 'strip' } %>
       <%= render 'shared/char_count',
                  type: "thread-title-#{post.id}",
-                 min: 1,
```
Oaphi, line 50: To make reviewing easier, this is the only change

trichoplax (approved):
All looks good. Bug no longer reproducible.

As for adding double quotes around Doc comments, this is unnecessary for me (my syntax highlighter happens to work fine without them), but since adding them causes no problem and fixes the syntax highlighting for some people, I see no reason not to.

PR #1807, opened 2025-08-24
Fix redirects to paths only used in AJAX requests after signing in

### LINKED ISSUE #1805: Logging in after opening a thread inline redirects to the thread API endpoint
**Steps to reproduce**
 
1. Go to any post, for instance https://a.dev.codidact.org/posts/721
2. Click on the title of a comment thread, which opens it on the page
3. Click the sign in button on the top right
4. Sign in
5. You are redirected to a buggy page, for instance https://a.dev.codidact.org/comments/thread/177/content?inline=true&show_deleted_comments=0

Image

I believe this is the HTML content of the thread which is meant to be dynamically added to the page, but for some reason is being used as the redirect.

Oaphi:
closes #1805

Also fixes every other instance of where it's probably possible to end up in the same situation as in the related issue that I managed to find, do let me know if there are any missing.

### DIFF
--- a/app/assets/javascripts/qpixel_api.js
+++ b/app/assets/javascripts/qpixel_api.js
@@ -372,6 +372,8 @@ window.QPixel = {
   fetchJSON: async (uri, data, options) => {
     const defaultHeaders = {
       'X-CSRF-Token': QPixel.csrfToken(),
+      // this is necessary for request.xhr?
+      'X-Requested-With': 'XMLHttpRequest',
       'Content-Type': 'application/json',
     };
 
--- a/app/assets/javascripts/qpixel_api.js
+++ b/app/assets/javascripts/qpixel_api.js
@@ -158,7 +158,8 @@ window.QPixel = {
     const myselfPromise = fetch('/users/me', {
       credentials: 'include',
       headers: {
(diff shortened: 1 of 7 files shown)


cellio (commented):
Behavior LGTM; definitely needs someone more fluent to review the code.

app/assets/javascripts/qpixel_api.js
```diff
@@ -393,17 +381,20 @@ window.QPixel = {
...
 
   getJSON: async (uri, options = {}) => {
+    const { headers = {} } = options ?? {};
+
     return QPixel.fetchJSON(uri, {}, {
       ...options,
+      headers: {
+        'Accept': 'application/json',
+        ...headers,
+      },
```
ArtOfCode-, line 391: Should we include the `X-Requested-With` header in here too?

ArtOfCode- (approved):

app/assets/javascripts/qpixel_api.js
Oaphi: Yeah, we should, missed it actually

ArtOfCode- (approved):

PR #1811, opened 2025-08-28
Eager load improvements for post lists

Oaphi:
This PR ensures we eager load associations used in post lists to avoid N+1 queries when loading category post lists. In particular, it significantly reduces the amount of queries made by `generic_share_link` (that needs `community`) and category tag ids getters (that need respective tags).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -261,7 +261,7 @@ class ApplicationController < ActionController::Base
     end
 
     # eager loading revived collections' used relation to prevent N+1 queries
-    @pinned_links = @pinned_links.includes(post: [:community])
+    @pinned_links = @pinned_links.list_includes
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -259,6 +259,10 @@ def pull_pinned_links_and_hot_questions
             .order('score DESC').limit(SiteSetting['HotQuestionsCount']).all
       end
     end
+
+    # eager loading revived collections' used relation to prevent N+1 queries
+    @pinned_links = @pinned_links.list_includes
+    @hot_questions = @hot_questions.list_includes
```
Oaphi, line 265: Eager loading is done in this manner here to faciliatate eager loading of collections reviced from cache.

ArtOfCode- (approved):

PR #1812, opened 2025-09-01
Add mod spammer review tool

### LINKED ISSUE #1729: Add an automated query/cleanup for potential spam accounts
Current query for these:

```sql
select
    u.id,
    u.email, 
    u.username,
    concat('https://', c.host, '/users/', u.id) as profile_link,
    u.profile_markdown
from users u
inner join community_users cu on u.id = cu.user_id and cu.id = (select id from community_users where user_id = u.id order by reputation desc limit 1)
inner join communities c on cu.community_id = c.id
left join posts p on p.user_id = u.id
left join comments cm on cm.user_id = u.id
left join votes v on v.user_id = u.id
where u.profile_markdown is not null
  and u.profile like '%href="%'
  and u.created_at >= date_sub(current_timestamp, interval 12 month)
  and u.deleted = false
  and u.email not like '%localhost'
  and cu.reputation = 1
group by u.id
having count(distinct p.id) = 0
   and count(distinct cm.id) = 0
   and count(distinct v.id) = 0
order by u.id
```

We should add this as an automated query/cleanup job in some form.

Discussion:

* How can we avoid this catching legitimate users? Do we time-limit to only the last 24 hours if this is running daily? Hopefully any established user would have been around longer than that, although we still run the risk of catching the odd user who's customised their profile but not yet _done_ anything.
* There is sufficient volume of these that manually reviewing each one is a heavy job. How can we lighten that? Can we automate any part of this?

ArtOfCode-:
Closes #1729

Adds a tool for moderators to review accounts that are potential spam profiles.

### DIFF
--- a/app/assets/javascripts/moderator.js
+++ b/app/assets/javascripts/moderator.js
@@ -38,4 +38,13 @@ $(() => {
       QPixel.createNotification('danger', `<strong>Failed:</strong> Unexpected status (${req.status})`);
     }
   });
+
+  QPixel.DOM.addSelectorListener('click', '.js-bulk-check', async (ev) => {
+    const tgt = ev.target;
+    const action = tgt.dataset.check;
+    const checkboxes = document.querySelectorAll('.js-spammer-form input[type="checkbox"]');
+    checkboxes.forEach((checkbox) => {
+      checkbox.checked = action === 'all';
+    });
+  });
 });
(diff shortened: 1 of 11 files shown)


cellio:
I'm not sure how to safely test this, so this comment is based only on the diff.

Could we have a "handled" tab/view, like for flags, with links to spam-deleted profiles?  (These should use the original names, not "deleted user".)  In addition to confirming the output, this gives them (and us) a way to see how big the problem is on a community over time, and if needed, a way to look into a specific case where the user number is unknown.  We're soft-deleting, so the profiles are still there for mods.

An alternative (that could be done in a separate change later) would be to give mods a list of recently-deleted *users*, filterable by all/self/mod/spam.  Maybe that'd be better?

Oaphi (approved):
LGTM

PR #1815, opened 2025-09-02
Fix ordering of cached collections upon revive

ArtOfCode-:
Moves order clauses out of `fetch_collection` blocks.


Oaphi (approved):
not sure if it's ready for review, but in any case - LGTM, tested, works as expected, and we now have a regression test on top of it, nice!

PR #1817, opened 2025-09-02
Fix tag search ordering

### LINKED ISSUE #1816: Tag search is no longer sorted by name
Fixing tag search server error had an unintended side-effect of losing `order` as it's on the second unioned query.

Oaphi:
closes #1816

Tags are supposed to be order by in ascending order, name matches first, then excerpt matches


cellio (commented):
Tested and LGTM.  Code changes look reasonable to me but I'd like another pair of eyes on them.

ArtOfCode- (approved):

PR #1818, opened 2025-09-03
Fix hints width on the last step of the tour

### LINKED ISSUE #247: Tour is too squished to be readable on phones
https://meta.codidact.com/questions/278901

The tour uses a side-by-side layout, with the question on the left and the tutorial text on the right.  This works badly on phones (see linked post for screenshots).  On a small viewport, can we use a vertical alignment instead, question above and tutorial text below?

Oaphi:
closes #247 with a hammer this time

The PR also significantly reworks question 3 of the tour to make it work smoothly on narrow screens:

https://github.com/user-attachments/assets/751e9471-4f05-4e35-8a27-7752f603ab84


ArtOfCode- (approved):
Approved for code changes

PR #1820, opened 2025-09-05
Assorted fixes for errors

Oaphi:
Fixes 2 server errors (likely no longer possible due to us switching to soft deletions of users, but we still have to account for full deletion requests and issues caused by legacy behavior of hard-deleting users) and a couple of other, non-crashing errors:

1. 500 when trying to render a post deleted by a user that no longer exists;
2. 500 when trying to render a post with reactions by a user that no longer exists;
3. Unexpected request (always fails) for comment thread locking when opening the modal due to selector conflicts;
4. Comment threads lacking the `locked_at` column despite being lockable (`UnknownAttributeError` error);
5. Accessing thresholds for abilities with one or more thresholds set to `0` (`FloatDomainError` error);
6. Lack of feedback for validation failures when trying to restrict comment threads (as a bonus, trying to **archive**, **delete**, **follow**, or **lock** a thread no longer reloads the full page, just updates the the thread in place);

I haven't checked if there are open issues because this is a followup for a chat discussion, but feel free to link them if you know of any.

Note for reviewers: this PR requires migrations to be run when checked out and rolled back when left to preserve your sanity. Don't forget to `rails db:migrate` -> `rails db:rollback` as needed.

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -19,17 +19,13 @@ $(() => {
     return $tgt.closest('.js-comment-thread-wrapper')[0] ?? null;
   };
 
-  $(document).on('click', '.post--comments-thread.is-inline a', async (evt) => {
-    if (evt.ctrlKey) { return; }
-
-    evt.preventDefault();
-
-    const $tgt = $(evt.target);
-    const $threadId = $tgt.data('thread');
-    const wrapper = getCommentThreadWrapper($tgt);
-
-    openThread(wrapper, $threadId);
-  });
+  /**
+   * @param {HTMLElement} wrapper
+   * @returns {boolean}
+   */
+  const isInlineCommentThread = (wrapper) => {
+    return !!wrapper.querySelector('[data-inline=true]');
+  };
 
   /**
    * @param {HTMLElement} wrapper
@@ -45,8 +41,24 @@ $(() => {
     window.hljs && hljs.highlightAll();
   }
 
+  $(document).on('click', '.post--comments-thread.is-inline a', async (evt) => {
+    if (evt.ctrlKey) {
+      return; // TODO: do we need this early exit?
+    }
+
+    evt.preventDefault();
+
+    const $tgt = $(evt.target);
+    const $threadId = $tgt.data('thread');
+    const wrapper = getCommentThreadWrapper($tgt);
+
+    openThread(wrapper, $threadId);
+  });
+
   $(document).on('click', '.js-show-deleted-comments', (ev) => {
-    if (ev.ctrlKey) { return; } // do we really need it?
+    if (ev.ctrlKey) {
+      return;
+    } // do we really need it?
 
     ev.preventDefault();
 
@@ -80,7 +92,9 @@ $(() => {
     }
 
     if (isDeleted) {
-      $container.append(`<i class="fas fa-trash h-c-red-600 fa-fw" title="Deleted thread" aria-label="Deleted thread"></i>`);
+      $container.append(
+        `<i class="fas fa-trash h-c-red-600 fa-fw" title="Deleted thread" aria-label="Deleted thread"></i>`
+      );
       $container.addClass('is-deleted');
     }
 
@@ -171,8 +185,7 @@ $(() => {
       if (isDelete) {
         $comment.addClass('deleted-content');
         $tgt.removeClass('js-comment-delete').addClass('js-comment-undelete').val('undelete');
-      }
-      else {
+      } else {
         $comment.removeClass('deleted-content');
         $tgt.removeClass('js-comment-undelete').addClass('js-comment-delete').val('delete');
       }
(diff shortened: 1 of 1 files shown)


cellio (approved):
Tested posts and reactions from deleted users.  (I don't have any destroyed ones, so we should double-check this on the dev server.)  LGTM.

app/views/abilities/show.html.erb
```diff
@@ -124,21 +124,21 @@
...
-    <% end %>
-  </div>
-<% end %>
+  <% end %>
 
-<div class="h-p-2">
-  <%= raw @ability.description %>
-</div>
+  <div class="h-p-2">
+    <%= raw @ability.description %>
```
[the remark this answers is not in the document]
Oaphi: ugh, these checks just waste time. This is fine, it's a seeded value

test/fixtures/posts.yml
```diff
@@ -480,3 +480,18 @@ question_without_votes:
   community: sample
   category: main
   license: cc_by_sa
+
+post_scorer_question:
+  post_type: question
+  title: This is a question for testing post scores
+  body: This post MUST be positively scored to count towards the score
```
cellio, line 487: World's tiniest nit: for the benefit of future readers, maybe don't say "positively scored" when we mean > 0.5?  0.3 is positive mathematically but not what you meant.

app/controllers/comments_controller.rb
```diff
@@ -219,47 +219,85 @@ def thread_rename
     redirect_to comment_thread_path(@comment_thread.id)
   end
 
+  def archive_thread
```
cellio, line 222: For other reviewers: the code in this block is copied from elsewhere, no changes to logic.

app/views/reactions/_list.html.erb
```diff
@@ -22,6 +22,9 @@
     <% end %>
 </div>
 
+<%# TODO: eager load instead %>
```
cellio, line 25: Is this something we need to address in this PR, or a note for later?  (Fine with me either way; I just didn't know whether to block on it, so I'm just commenting.)

cellio (commented):
LGTM.  I recommend a second code-reviewer as a sanity check, particularly if you're going to handle those TODOs about new routes here rather than later.

app/views/reactions/_list.html.erb
Oaphi: Just a note for later, I tend not to add TODOs unless it's critical to fix before updating. This one in particular is just an observation that we can remove `host` hoisted out of the loop if we preload `community` (otherwise, it'll result in N+1 queries)

app/controllers/comments_controller.rb
Oaphi: As a note for other reviewers - I went ahead and split the `thread_restrict` action, so archiving, deleting, locking, and following threads need to be tested.

cellio: I tested archive/restore, delete/undelete, lock/unlock, and follow/unfollow, all from the in-page tools menu.  All of that worked fine.

I then repeated the tests on the thread page.  For lock, archive, delete, and follow, the action worked but when the page was refreshed, it showed the "collapse" and "show thread" controls (from the in-page view), which don't make sense here:

screenshot with extra controls highlighted

Reloading the page removes them, so this is only an issue on the immediate response to the action, which I guess is assuming context from the in-page view?  (I.e. the Tools menu can't tell where it is, maybe?)  Not a big deal but a bit weird -- can be deferred if this isn't a trivial fix.

Interestingly, this does *not* happen for unlock, restore, undelete, and unfollow.  I looked for a difference in the controller (since these used to be rolled up into two methods, "restrict" and "unrestrict"), but I'm not spotting it if that's where it's coming from.  The `comment_thread` view isn't changed on this PR, but this behavior doesn't happen on the dev server as of [5c83ab90](https://github.com/codidact/qpixel/commit/5c83ab9075f385f7e076e9310962c18e34ac3f5f) (2025-09-05 19:17:57Z).

Oaphi: Eh, that's likely just a side-effect of that we now replace the thread's content instead of reloading the whole page. It works as expected, I suspect we also have the same issue when expanding threads from inline view. Good catch, let's fix it while we are at it

cellio (approved):
Ruby code LGTM.  JS code doesn't raise any alarms but we should have someone who understands JS better take a look at that part.

I tested the items listed in the PR description and everything that I'm able to test works fine, ~~except for one minor weird rendering thing described in an earlier comment.  It's ok with me if we file an issue to fix that later.~~ (fixed)  Also, as mentioned earlier, I don't have any destroyed users to experiment on; I don't see anything here that would make that *worse*, and it very likely fixes problems from that, so if we can't test it in dev, that shouldn't block us.  I think we have some destroyed users on the dev server but am not certain.

PR #1826, opened 2025-09-12
Keyboard event-related fixes + notifications improvements

### LINKED ISSUE #1824: Opening inbox item in a new tab with ctrl-click closes notifications in active tab
meta:293782

If you open your notifications and right-click to open in a new tab, the page you're on doesn't change.  In particular, notifications remain open (and you have a new tab).

If, instead, you use ctrl-click (cmd-click on Mac) to open an item in a new tab, we close the notifications.  We should leave the active tab unchanged, as in the first case.

(I don't know if touch interfaces have two different ways to launch links in new tabs, and if so, how they behave.)

Oaphi:
closes #1824
completes meta:293809 (hasn't been able to find an issue that corresponds to it)

Also finally moves our keyboard shortcuts manager from a global `_codidactKeyboard` variable to the `QPixel` namespace (`Keyboard`), as well as simplifies its initialization handling.

\+ fixes the inbox icon doing nothing for users with JavaScript disabled (it's now linked to the full inbox page);
\+ allows inbox notifications (on the full page) to be filtered by status (all / read / unread), as well as ordered by "oldest first";
\+ adds an icon indicating whether the notification is read or not (we don't have _any_ currently):

Screenshot from 2025-10-07 17-56-50

### DIFF
--- a/app/assets/stylesheets/notifications.scss
+++ b/app/assets/stylesheets/notifications.scss
@@ -9,7 +9,7 @@
   border-top: 1px solid $muted-graphic;
   display: grid;
   gap: 0.75em;
-  grid-template-columns: max-content auto;
+  grid-template-columns: auto 1fr;
   align-items: center;
   padding: 0.5em;
 
--- a/app/views/notifications/_notification.html.erb
+++ b/app/views/notifications/_notification.html.erb
@@ -1,0 +1,15 @@
+<%#
+   "Renders an inbox notification
+
+    Variables:
+      notification : Notification to render
+"%>
+
+<div class="notification">
+  <div class="notification-content">
+    <div class="form-caption">
+      <%= notification.rendered_timestamp %>
+    </div>
+    <a href="<%= notification.link %>"><%= render_pings_text(notification.content) %></a>
+  </div>
+</div>
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -29,14 +29,7 @@
     <h2>Unread</h2>
   <% end %>
   <% unread_notifications.each do |notif| %>
-    <div class="notification">
-      <div class="notification-content">
-        <div class="form-caption">
-          <%= CGI.unescape_html(notif.created_at.strftime('%b&nbsp;%-d,&nbsp;%Y&nbsp;at&nbsp;%H:%M')).html_safe %>
-        </div>
-        <a href="<%= notif.link %>"><%= render_pings_text(notif.content) %></a>
-      </div>
-    </div>
+    <%= render 'notifications/notification', notification: notif %>
   <% end %>
 <% end %>
 
@@ -45,14 +38,7 @@
     <h2>Read</h2>
   <% end %>
   <% read_notifications.each do |notif| %>
-    <div class="notification">
-      <div class="notification-content">
-        <div class="form-caption">
-          <%= CGI.unescape_html(notif.created_at.strftime('%b&nbsp;%-d,&nbsp;%Y&nbsp;at&nbsp;%H:%M')).html_safe %>
-        </div>
-        <a href="<%= notif.link %>"><%= render_pings_text(notif.content) %></a>
-      </div>
-    </div>
+    <%= render 'notifications/notification', notification: notif %>
   <% end %>
 <% end %>
 


cellio:
> Does this also fix [Stuck red border after reducing below maximum characters](https://meta.codidact.com/posts/293635) (asking because it is so closely related to the linked meta post, so might be worth testing)?

Yes.  I tested both linked meta cases on this branch in Firefox and saw red borders where they're supposed to be and not otherwise.

cellio:
I pulled this branch again (to test the red-outline thing for comments), and *now* I'm seeing cmd-click open a notification in a separate tab (Firefox and Chrome).  It also marks the notification as read, but right-clicking to open in a new tab does not (never did).  Which behavior do users of cmd-click/ctrl-click expect?

I don't have a Windows machine to test ctrl-click on.

trichoplax:
> Which behavior do users of cmd-click/ctrl-click expect?

I don't know which behaviour is expected, but the inconsistencies are probably best dealt with in a separate issue since they go beyond what is being changed here.

> It also marks the notification as read, but right-clicking to open in a new tab does not (never did).

In addition to things like this not being picked up by the JavaScript code for users who have JavaScript, users with JavaScript disabled get no response when clicking on the new notification circle or the envelope icon, so they have no way to view the notifications unless they know the URL for the stand alone notifications page, and even once there they cannot mark notifications as read.

Maybe things that don't yet result in a notification being marked as read could be added to a separate issue to make notifications available to users with JavaScript disabled, along with a way to mark notifications as read without JavaScript?

trichoplax:
I've added #1833 for the larger task. Feel free to add subtasks to that one that don't need to be addressed here.

ArtOfCode- (approved):

cellio:
Nice improvements!  `cmd-click` now opens an inbox item in a new tab and marks it as read (tested in Chrome), and with JS disabled the notification icon takes me to the inbox page (tested in Firefox).  The new filter buttons are great, too!

I am confused by the icons on the inbox page.  In the notifications drop-down, I always assumed that the icon corresponded to the action in the link -- "mark read" = open the envelope, "mark unread" = close that thing back up in its envelope for later processing.  But now we're using an icon tied to an action to report a status, and it feels like maybe they're backwards?  An item with the opened envelope is *unread* on the inbox page:

screenshot of inbox page with notifications pane opened; the first few items are unread

It becomes clear enough what the icons mean if you use the buttons, but I wonder if we should reverse them, and/or add the icons into the buttons too, so you'd have the "key" right there when viewing the page, like this:

mockup of 3 buttons, with icons by Read and Unread

Note: if these icons on the full inbox page were *controls*, i.e. clicking them changed the status, then I'd agree that we'd want the icons as they are.  It would be good to add controls, which we do not need to do *here*.  So maybe, having written all this, I've talked myself into having *no* icons now, neither the lines nor the buttons, and adding icons when they are controls like in the notifications pane.  I'd definitely like to hear other thoughts here.

Oaphi:
> I am confused by the icons on the inbox page. In the notifications drop-down, I always assumed that the icon corresponded to the action in the link -- "mark read" = open the envelope, "mark unread" = close that thing back up in its envelope for later processing.

To be honest, those icons are there only because I failed to find a better way of separating read notifications from unread - suggestions welcome. Adding icons to the filter buttons sounds good to me, but let's see if someone can come up with a better approach

cellio:
> To be honest, those icons are there only because I failed to find a better way of separating read notifications from unread - suggestions welcome. Adding icons to the filter buttons sounds good to me, but let's see if someone can come up with a better approach

Is the light blue background, like in the notifications pane, possible?   And then that saves the icon "slot" for a control later.

cellio:
I noticed that this PR is waiting for a review from me, but I'd like to do something about the icons per my last comment first.

Oaphi:
> Is the light blue background, like in the notifications pane, possible? And then that saves the icon "slot" for a control later.

That's probably going to be a bit too noisy on a full-blown page, methinks. How about we actually split the lists (they are already split and just *rendered* as a single list) and make it clear which notifications are which? I am thinking of something along the lines of:

Screenshot from 2025-11-29 16-34-42

With no unnecessary headers when filtering by "read" or "unread":

Screenshot from 2025-11-29 16-34-45
Screenshot from 2025-11-29 16-34-50

cellio:
> That's probably going to be a bit too noisy on a full-blown page, methinks. How about we actually split the lists (they are already split and just _rendered_ as a single list) and make it clear which notifications are which? I am thinking of something along the lines of:

Yes, that's much better!  As you said, they're split already; in the notifications modal we bubble up the unread ones even if newer ones are read.  So this is consistent.  I like your design.

Oaphi:
> Yes, that's much better! As you said, they're split already; in the notifications modal we bubble up the unread ones even if newer ones are read. So this is consistent. I like your design.

Agreed then - everything should be on the branch, will merge if your review doesn't come up with any issues

Issue #1808, opened 2025-08-25
Tracking issue for unnecessary requests


Oaphi:
This issue keeps track of unnecessary requests causing server load for no reason:

- [ ] `/users/me/filters/default?category=undefined` on the "Filters" tab of the user page;

PR #1827, opened 2025-09-12
Don't request category default filters on pages that don't have categories

Oaphi:
Partially addresses #1808 (unnecessary requests for category default filters on pages that don't have a category such as user tabs). As a side-effect, finally adds proper Prettier config to format our JS assets (note that [prettier-plugin-brace-style](https://github.com/ony3000/prettier-plugin-brace-style) is needed to support our choice of brace style - Prettier's team is [dead set on not supporting it](https://github.com/prettier/prettier/issues/840)).

### DIFF
--- a/app/assets/javascripts/filters.js
+++ b/app/assets/javascripts/filters.js
@@ -1,4 +1,4 @@
-$(() => {
+document.addEventListener('DOMContentLoaded', () => {
   $('.js-filter-select').toArray().forEach(async (el) => {
     const $select = $(el);
     const $form = $select.closest('form');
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/filters.js
```diff
@@ -1,158 +1,169 @@
...
+              computeEnables();
+
+              // Name is not one of the presets, i.e user is creating a new preset
+              if (!preset) {
+                return;
+              }
+
+              for (const [name, value] of Object.entries(preset)) {
+                const $el = $form.find(`.form--filter[data-name=${name}]`);
+                if (value?.constructor == Array) {
```
Oaphi, line 107: Note to self - we can safely switch to `Array.isArray(value)` here

trichoplax:
> I personally don't have an opinion on the matter (I rarely do when it comes to formatting - happy to work with any as long as a formatter can be configured to automatically bring code into required shape).

I feel the same way. Only asked in case it would save us a dependency (if everyone happens to not mind).

ArtOfCode- (approved):

PR #1834, opened 2025-09-16
Switch to cookie-based CSRF

Oaphi:
This PR is a **major** change in how we handle CSRF. Due to us having to heavily cache pages in prod, we ended up with a ton of issues where authenticity tokens get cached with the page, resulting in stale tokens breaking critical paths. This happens because the default strategy is to output the token to the page itself via `csrf_meta_tags` - of course, if it gets cached for a while (and we currently cache up to 24 hours), all bets are off.

With this PR, we'll start sending and checking an **HTTP-only** `csrf_token`  cookie alongside the session cookie. This effectively bypasses the need to send `X-CSRF-Token` (hence, no need to make the cookie JavaScript-readable) as long as all our POST-like requests include credentials (which send cookies over).

This changes critical functionality, so needs to be thoroughly tested before deployment

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -16,6 +16,7 @@ Rails.application.configure do
   # Full error reports are disabled and caching is turned on.
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
+  Rack::MiniProfiler.config.disable_caching = false
 
   # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
   # or in config/master.key. This key is used to decrypt credentials (and other encrypted files).


app/assets/javascripts/qpixel_api.js
```diff
@@ -7,12 +7,6 @@ const validators = [];
 let popped_modals_ct = 0;
 
 window.QPixel = {
-  csrfToken: () => {
```
Oaphi, line 10: Since CSRF token is now sent as a cookie, there's no longer a need to read the token from DOM or add it as a `X-CSRF-Token` header for that matter.

app/assets/javascripts/qpixel_api.js
```diff
@@ -320,10 +314,8 @@ window.QPixel = {
 
   fetch: async (uri, init) => {
     const defaultHeaders = {
-      'X-CSRF-Token': QPixel.csrfToken(),
       // X-Requested-With is necessary for request.xhr? to work
       'X-Requested-With': 'XMLHttpRequest',
-      'Content-Type': 'application/json',
```
Oaphi, line 326: This is a partially related fix that ensures `QPixel#fetch` can be used with non-JSON (`Content-Type` request header) requests to avoid issues with forgetting to set `X-Requested-With` or forwarding credentials. Should not break anything

ArtOfCode- (approved):

app/controllers/errors_controller.rb
```diff
@@ -1,7 +1,7 @@
 # Provides web actions that represent errors. Rails' standard error pages are static HTML with inline CSS; by using
 # a custom error controller we get all the layouts and CSS.
 class ErrorsController < ApplicationController
-  protect_from_forgery except: [:error]
+  protect_from_forgery except: [:error], store: :cookie
```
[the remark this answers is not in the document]
Oaphi: Eh?

Issue #1838, opened 2025-09-17
Sort decided edits/flags by decision time


cellio:
The flags page (moderators) and the edits page (editors) both have a tab for things that have already been decided.  For things *pending* decisions, we show the oldest first, which makes sense (prioritize the things that have been waiting longest).  For the list of things that have been handled, however, it would be better to sort by the handle time, so that other users can more easily see recent activity.

I assume this is a simple matter of an "order by" clause in a couple queries, but I'm in the middle of something else right now so I'm filing this instead of fixing it.

cellio:
I assume adding the buttons is pretty straightforward, so no need to separate that out.  I think we want to have the decided ones default to decision time, but it makes sense to give people the option to sort differently.

PR #1842, opened 2025-09-21
Sorting options for suggesged edits

Oaphi:
This PR adds a reusable configurable partial for sorting options and connects it to suggested edit lists as a proof of concept (as well as to partially address #1838). I wanted sorting order options to just be a togglable icon, but our icon font seems way out of date as it lacks any (if I am just blind, do let me know):

Screenshot from 2025-09-21 15-10-54

Note that it also moves the category list into `<details>` in preparation for adding proper filters - it wastes a lot of space with little benefit to the user (pending suggested edits are scoped per category by default, so category switching is always optional):

Screenshot from 2025-09-21 15-10-44

Screenshot from 2025-09-21 15-25-54

Finally, the PR also enables the categories table for decided edits too as being able to quickly switch categories is no less useful there.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,5 +1,6 @@
-# Provides web actions that represent errors. Rails' standard error pages are static HTML with inline CSS; by using
-# a custom error controller we get all the layouts and CSS.
+# Provides web actions that represent errors. Rails' standard error pages are static HTML with inline CSS;
+# by using a custom error controller we get all the layouts and CSS, as well as error reporting.
+# Note that it requires +consider_all_requests_local+ to be +false+ (see environment config files for details)
 class ErrorsController < ApplicationController
   protect_from_forgery with: :exception, except: [:error], store: :cookie
 


config/environments/development.rb
```diff
@@ -12,9 +12,11 @@
   # Do not eager load code on boot.
   config.eager_load = false
 
-  # Show full error reports and disable caching.
+  # Set this to +true+ unless you want to skip exceptions_app:
+  config.consider_all_requests_local = false
```
ArtOfCode-, line 16: Can we make this configurable, `true` by default? This is what enables Rails' default error pages in development, which provide more detail than our own error logging.

app/views/shared/_sorting.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  # Defaults
+  default_order = defined?(default_order) ? default_order : nil
+  default_type = defined?(default_type) ? default_type : nil
+%>
+
+<div class="flex gap-sm">
+  <div class="button-list is-gutterless">
+    <% [:asc, :desc].each do |order| %>
+      <% is_active = params[:order]&.to_sym == order || (params[:order].nil? && default_order == order) %>
+      <%= link_to order.to_s.humanize, request.params.merge(order: order),
```
ArtOfCode-, line 20: For sorting icons, you should have `fa-sort`, `fa-sort-up`, and `fa-sort-down` available.

config/environments/development.rb
Oaphi: @ArtOfCode- tbf, our error reports are quite good - I've never had to actually wade into the extended stack to resolve one. That said, I have no issue with making this configurable - I am not sure that the default should be `true` as it basically disables `ErrorsController` without the user realizing it, but have no explicit preference.

cellio (approved):
Looks great!  I didn't generate enough suggested edits to trip pagination in my testing, but since the sorting is done in the controller I'm assuming that's not an issue (but want to mention it in case it could be).

INFO: Would we be able to use this same shared sorting widget for existing sort buttons, like the users page and the user posts page?  This is not a change request; just asking if, as it appears, this is sufficiently general that we could.

Oaphi:
> Would we be able to use this same shared sorting widget for existing sort buttons, like the users page and the user posts page?

Yup, @cellio, that's the idea - not having sorting _order_ options is a pet peeve of mine (ahem, ahem, SE...)

config/environments/development.rb
Oaphi: Done, @ArtOfCode- - I alresdy decided not to change the default when making `consider_all_requests_local` configurable, so it was only a matter of adding a note to the controller

ArtOfCode- (approved):

PR #1843, opened 2025-09-21
Modify tag page query so it doesn't take 94 years

ArtOfCode-:
Updates the query used in the tag page posts list so that it isn't 3 miles long and takes 94 years to run. See Discord.


Oaphi (approved):
LGTM

PR #1848, opened 2025-09-22
Assorted fixes

Oaphi:
This PR contains several assorted fixes:

- we should not require users to be signed in to get `filters` - it breaks filter select for signed out users (the action had all the necessary checks inside);
- server error (500) when `@active_filter` is missing a name (to reproduce, append a `predefined_filter` search query param **without a value**);
- server error (500) when trying to directly access a post without a category (any post with a post type that doesn't belong to one);
- server error (500) when trying to render user cards for users without a community user;
- not a fix, but related: the "Version" line of error reports is now a link for searching by that option now that we can do that;

### DIFF
--- a/app/views/admin/_error_report.html.erb
+++ b/app/views/admin/_error_report.html.erb
@@ -28,7 +28,7 @@
     <% if report.version? %>
       <br/>
       <strong>Version:</strong>
-      <%= report.version %>
+      <%= link_to report.version, request.params.merge(version: report.version) %>
     <% end %>
   </p>
 
--- a/app/views/admin/_error_report.html.erb
+++ b/app/views/admin/_error_report.html.erb
@@ -28,7 +28,7 @@
     <% if report.version? %>
       <br/>
       <strong>Version:</strong>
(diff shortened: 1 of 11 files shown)


app/views/admin/_error_report.html.erb
```diff
@@ -28,7 +28,7 @@
     <% if report.version? %>
       <br/>
       <strong>Version:</strong>
-      <%= report.version %>
+      <%= link_to report.version, request.params.merge(version: report.version) %>
```
ArtOfCode-:
```suggestion
      <%= link_to report.version, query_url(version: report.version) %>
```

ArtOfCode- (changes requested):

app/views/admin/_error_report.html.erb
Oaphi: @ArtOfCode- maybe we should update other instances of this while we are at it? I only used this for consistency in the first place.

ArtOfCode- (approved):

PR #1852, opened 2025-09-28
Improve tag rename error handling & messages

### LINKED ISSUE #1851: Improve error message for failed tag renames
If renaming a tag fails, we give a fairly generic error message ("failed to rename the tag").  This message is raised in the tags controller in `rename`, where we only know that it failed (but not why).  

A common reason for a rename to fail is that the target tag exists.  Can we check for that case and have  a specific error message like "a tag with that name already exists"?  This would mean that errors for *other* reasons (like malformed names) would still have a different (generic) message than the merge case.

The relevant code is in `TagsController.rename`.

Oaphi:
closes #1851

Also fixes an old bug preventing validation for tags with spaces from taking effect.

Error notification when the new tag name contains spaces:

Screenshot from 2025-09-29 14-21-04

Error notification when the new tag name matches an already existing tag:

Screenshot from 2025-09-29 14-21-14

As a treat, it also ensures old tag names are quoted in the rename prompt (on a tangent, we should move from a prompt to a proper modal dialog, but this is outside of the scope of this PR):

image

Compared to how it is now:

Screenshot from 2025-09-28 12-52-15

Finally, it also changes error handling for tag creation (see https://github.com/codidact/qpixel/pull/1852#issuecomment-3343951054) a bit:

- removes duplicate error flashes after failing to create tags;
- improves the combined error message shown upon failure to create a tag (no actual changes, just a nice side-effect of other changes):

    Screenshot from 2025-09-29 14-52-59

### DIFF
--- a/app/assets/javascripts/qpixel_api.js
+++ b/app/assets/javascripts/qpixel_api.js
@@ -430,9 +430,9 @@ window.QPixel = {
   },
 
   handleJSONResponse: (data, onSuccess, onFinally) => {
-    const is_failed = data.status === 'failed';
+    const isFailed = data.status === 'failed';
 
-    if(is_failed) {
+    if(isFailed) {
       if (data.message) {
         QPixel.createNotification('danger', data.message);
       }
@@ -447,7 +447,7 @@ window.QPixel = {
 
     onFinally?.(data);
 
-    return !is_failed;
+    return !isFailed;
   },
 
   flag: async (flag) => {
--- a/app/assets/javascripts/qpixel_api.js
+++ b/app/assets/javascripts/qpixel_api.js
@@ -430,19 +430,24 @@ window.QPixel = {
   },
 
   handleJSONResponse: (data, onSuccess, onFinally) => {
-    const is_modified = data.status === 'modified';
-    const is_success = data.status === 'success';
+    const is_failed = data.status === 'failed';
 
-    if (is_modified || is_success) {
-      onSuccess(/** @type {Parameters<typeof onSuccess>[0]} */(data));
+    if(is_failed) {
+      if (data.message) {
+        QPixel.createNotification('danger', data.message);
+      }
+
+      for (const error of data.errors ?? []) {
+        QPixel.createNotification('danger', error);
+      }
     }
     else {
-      QPixel.createNotification('danger', data.message);
+      onSuccess(/** @type {Parameters<typeof onSuccess>[0]} */(data));
     }
 
     onFinally?.(data);
 
-    return is_success;
+    return !is_failed;
   },
 
   flag: async (flag) => {
--- a/app/assets/javascripts/qpixel_api.js
+++ b/app/assets/javascripts/qpixel_api.js
@@ -433,12 +433,22 @@ window.QPixel = {
     const isFailed = data.status === 'failed';
 
     if(isFailed) {
-      if (data.message) {
-        QPixel.createNotification('danger', data.message);
-      }
+      const { errors = [], message } = data;
 
-      for (const error of data.errors ?? []) {
-        QPixel.createNotification('danger', error);
+      if (message) {
+        const fullMessage =
+          errors.length > 1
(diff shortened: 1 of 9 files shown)


app/controllers/tags_controller.rb
```diff
@@ -155,7 +155,7 @@ def rename
       render json: { status: 'success', tag: @tag }
     else
       render json: { status: 'failed',
-                     message: I18n.t('tags.errors.rename_generic'),
+                     errors: @tag.errors.messages.values.flatten,
```
ArtOfCode-:
```suggestion
                     errors: @tag.errors.full_messages,
```

app/assets/javascripts/qpixel_api.js
```diff
@@ -430,19 +430,24 @@ window.QPixel = {
...
     }
     else {
-      QPixel.createNotification('danger', data.message);
+      onSuccess(/** @type {Parameters<typeof onSuccess>[0]} */(data));
     }
 
     onFinally?.(data);
 
-    return is_success;
+    return !is_failed;
```
ArtOfCode-:
```suggestion
    const isFailed = data.status === 'failed';

    if(isFailed) {
      if (data.message) {
        QPixel.createNotification('danger', data.message);
      }

      for (const error of data.errors ?? []) {
        QPixel.createNotification('danger', error);
      }
}
else {
      onSuccess(/** @type {Parameters<typeof onSuccess>[0]} */(data));
}

onFinally?.(data);

    return !isFailed;
```

ArtOfCode- (changes requested):

app/controllers/tags_controller.rb
Oaphi: The reason why I opted to not use `full_messages` is because it prepends the attribute's name, leading to our error messages looking like `Name Tag name should contain at least 1 character.` - that's not Rails's problem, though, it's a problem with our validation error messages. Will commit this shortly after updating locale strings to match

cellio:
Whatever we're doing to make nicer error messages, we might want to also apply to tag *creation* if that's easy.  This is from the "new tag" button on the tags page:

Name Tag names may not contain spaces

Also, I don't know if this is something about a dev environment or something else, but for *one* of the two errors from the rename path, I got an error message with a timestamp in it:

Screenshot

The other one was as shown in the description.

app/assets/javascripts/qpixel_api.js
```diff
@@ -430,19 +430,34 @@ window.QPixel = {
...
+              ? `${message} (${errors[0].toLowerCase().trim()})`
+              : message;
+
+          QPixel.createNotification('danger', fullMessage);
+      }
+      else {
+        for (const error of errors) {
+          QPixel.createNotification('danger', error);
+        }
+      }
```
Oaphi, line 452: The logic for handling failed responses is as follows:

- If only the `message` field is present, show it directly;
- If only the `errors` field is present, create a notification for each error listed;
- If both `message` and `errors` are present:
    - if only one error is present, append a lowercase version of it in parentheses (`<message> (<error>)`);
    - otherwise, use `message` as the notification's header and display `errors` as an unordered list (matches what our `flash` logic does);

Oaphi:
> Also, I don't know if this is something about a dev environment or something else, but for one of the two errors from the rename path, I got an error message with a timestamp in it:

@cellio IIRC, the timestamp is shown when there's already an error notification present on the page - see `createNotification` in the client-side API. We should make it prettier, but that's a bit out of scope for this PR.

cellio:
> @cellio IIRC, the timestamp is shown when there's already an error notification present on the page - see `createNotification` in the client-side API. We should make it prettier, but that's a bit out of scope for this PR.

Ah, ok!  If this PR didn't add it, then definitely don't do anything about it *here*.  And maybe not at all; I didn't realize the connection to the other error that was already on the page, and in that rare case it does seem useful to distinguish somehow.  Let's leave it alone.

cellio (approved):
Tested with latest changes.  LGTM.  Didn't code-review JS.

ArtOfCode- (approved):

PR #1854, opened 2025-09-28
Remove extremely high z-index for Select2 containers causing them to render over modal backgrounds

### LINKED ISSUE #1546: Misplaced highlighting during image upload
meta:293269

During image upload, most of the page is grayed out to focus on the modal -- except two input fields farther down the page, which didn't get the message.  Doesn't affect functionality but looks weird.

Oaphi:
closes #1546

This is necessary - `z-index` needs to be added only when there is _no other way,_ not the other way around, otherwise it always ends up in issues like this one (not to mention that Select2 already controls the property for its dropdowns, there's no need to interfere with it). I've checked what I could - do let me know if `z-index` now needs to be slightly increased in some places to avoid overlapping.

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -47,7 +47,6 @@ select.form-element {
 
 .select2-container {
   width: 100% !important;
-  z-index: 8999;
 }
 
 .select2-results__option {


ArtOfCode- (approved):

cellio (approved):
LGTM.  Tested image upload per the issue and checked a few other modals (flag, close, tools, notifications).  Everything looks fine.

Oaphi:
Note to self: check mobile header dropdowns before merge

PR #1860, opened 2025-10-01
Allow Markdown in post titles

### LINKED ISSUE #1857: Allow CommonMark in post titles
meta:294604

User request: it should be possible to use at least a subset of Markdown that we support in posts in their titles. There's a good case for allowing inline code, although I am not so sure about allowing emphasis.

cc @RokeJulianLockhart

Oaphi:
closes #1857

This PR enables Markdown rendering in post titles.
Allowed formatting types are configurable per community.

This is how posts will look like in lists with _all_ formatting types enabled:

and in the expanded view:

The new site setting is called `AllowedPostTitleFormattingTypes`:

Screenshot from 2025-10-01 13-04-36

As a side-effect, it also enables markdown rendering for site settings descriptions - this allows us to be more expressive in them (of course, descriptions are sanitized).

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -115,11 +115,6 @@ class Post < ApplicationRecord
     parent.nil? ? category.tag_set : parent.category.tag_set
   end
 
-  # @return [Boolean]
-  def meta?
-    false
-  end
-
   # Used in the transfer of content from SE to reassign the owner of a post to the given user.
   # @param new_user [User]
   def reassign_user(new_user)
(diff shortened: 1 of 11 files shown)


app/models/post.rb
```diff
@@ -257,19 +198,85 @@ def related_posts_for(user)
     end
   end
 
+  # predicates:
```
cellio, line 201: I'm assuming these are just relocations that confused the diff (and a couple renames for top level and spam flags).  I skimmed but didn't carefully compare these to the deleted blocks.

cellio (approved):
LGTM.  Tested various combinations, and also confirmed that I can't type gibberish into the new field's value.

app/models/post.rb
Oaphi: The diffs look fine to me - do you use the "unified" view, perchance? This one can get _really_ atrocious, I exclusively use the side-by-side one

cellio: Thanks for the tip.  The side-by-side view is too narrow for me (at the font sizes I need), but it's good to remember that it exists and I can use it to check specific things like this.

ArtOfCode- (approved):

PR #1862, opened 2025-10-02
Move 'native' sorting order option to filters

### LINKED ISSUE #1253: Support native/imported/all in filters, for sites with imported content
Prompted by, but not exactly, meta:289139 .

A few communities support imported content, and those communities have an extra sorting button on the category page for "native".  It's actually a sort, not a filter, but the user expects a filter.  Now that we have filters, can we fold it in there?

The presence of the button is controlled by this logic:
```
    <% if SiteSetting['AllowContentTransfer'] %>
      <%= link_to 'Native', request.params.merge(sort: 'native'),
                  class: "button is-muted is-outlined #{params[:sort] == 'native' ? 'is-active' : ''}",
                  title: 'exclude imported posts' %>
    <% end %>
```

On sites with imported content only, could the filters widget add a control with a drop-down menu for native, imported, any, like this?  This is similar to the Status field for open, closed, any.

native-mockup

Oaphi:
closes #1253

This PR implements a new filter governing whether imported content are included in post lists. It also removes the sorting option as per discussion in [Discord](https://discord.com/channels/634104110131445811/634104110131445815/1423338179506733187) and in [Native sorting button blocks other sorting orders](https://meta.codidact.com/posts/289141). It also starts work on making our filters widget responsive.

Available options are:
- "any" (returns both native and imported);
- "native" (excludes imported);
- "imported" (returns only imported);

Coincidentally, it also fixes the `InterestingSubscriptionScoreThreshold` site setting's value type - it does not work with our score system as it's of type `integer`, whereas it should be a `float`. I opted to fix it up with a data migration to be really sure everyone gets the fixed value - but open to moving it to a maintenance task if necessary.

Note for reviewers: there are 2 new migrations.

---

Very large width and wider - 4 column layout:

Screenshot from 2025-10-02 21-38-15

Large width - 3 column layout:

Screenshot from 2025-10-02 21-38-34

Medium width - 2 column layout:

Screenshot from 2025-10-02 21-38-45

Small width and narrower - 1 column layout (note that fixing the remaining parts of the filters widget is outside the scope of this PR):

Screenshot from 2025-10-02 21-38-55

### DIFF
--- a/app/assets/stylesheets/search.scss
+++ b/app/assets/stylesheets/search.scss
@@ -4,6 +4,24 @@
   border-top: 1px solid $muted-graphic;
 }
 
+.search-filters-constraints {
+  display: grid;
+  column-gap: 0.5em;
+  grid-template-columns: repeat(4, 1fr);
+
+  @media screen and (max-width: $screen-lg) {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @media screen and (max-width: $screen-md) {
+    grid-template-columns: repeat(2, 1fr);
+  }
+
+  @media screen and (max-width: $screen-sm) {
+    grid-template-columns: repeat(1, 1fr);
+  }
+}
+
 .search-sorting {
   align-items: center;
   display: flex;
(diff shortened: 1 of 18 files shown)


cellio:
Worth documenting here: The "Native" button is only shown on the category page if `AllowContentTransfer` is true.  This PR correctly (IMO) uses that same check for the filter, only showing the "source" option if that setting is true.  (*Most* of our communities, and probably *all* of everyone else's, won't have content imported from SE, so showing this option in those cases would be confusing.)

ArtOfCode- (approved):

PR #1866, opened 2025-10-05
Add audit logs for mod warning creations

Oaphi:
As per [discussion in Discord](https://discord.com/channels/634104110131445811/668890340199235613/1424214885104881714) - this minor PR adds an audit log for when someone creates a mod warning:

Screenshot from 2025-10-05 14-12-49

We already audit mod warning _lifts_ - warning creation is just as important for accountability.

Tangentially, this PR also adds a small fix for audit log category, type, and subject column wrapping (`overflow-wrap: anywhere` is way too liberal for those).

### DIFF
--- a/app/controllers/mod_warning_controller.rb
+++ b/app/controllers/mod_warning_controller.rb
@@ -70,7 +70,11 @@ class ModWarningController < ApplicationController
     @warning.update(active: false, read: false)
     @user.community_user.update is_suspended: false, suspension_public_comment: nil, suspension_end: nil
 
-    audit_warning('suspension_lift', @warning)
+    audit_warning('warning_lift', @warning)
+
+    if @warning.suspension?
+      audit_warning('suspension_lift', @warning)
+    end
 
     flash[:success] = 'The warning or suspension has been lifted. Please consider adding an annotation ' \
                       'explaining your reasons.'
--- a/app/controllers/mod_warning_controller.rb
+++ b/app/controllers/mod_warning_controller.rb
@@ -46,12 +46,15 @@ class ModWarningController < ApplicationController
     @warning = ModWarning.new(author: current_user, community_user: @user.community_user,
                               body: params[:mod_warning][:body], is_suspension: is_suspension,
                               suspension_end: suspension_end, active: true, read: false)
+
     if @warning.save
       audit_warning('warning_create', @warning)
 
       if is_suspension
         @user.community_user.update(is_suspended: is_suspension, suspension_end: suspension_end,
                                     suspension_public_comment: params[:mod_warning][:suspension_public_notice])
+
+        audit_warning('suspension_create', @warning)
       end
 
       redirect_to user_path(@user)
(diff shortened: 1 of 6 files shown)


cellio:
Are we able to distinguish between warnings and suspensions?  (`is_suspension` is there in the properties; I'm talking about the name of the event.)

app/controllers/mod_warning_controller.rb
```diff
@@ -65,8 +67,7 @@ def lift
     @warning.update(active: false, read: false)
     @user.community_user.update is_suspended: false, suspension_public_comment: nil, suspension_end: nil
 
-    AuditLog.moderator_audit(event_type: 'warning_lift', related: @warning, user: current_user,
-                             comment: "<<Warning #{@warning.attributes_print} >>")
+    audit_warning('warning_lift', @warning)
```
cellio, line 73: Let's call this `suspension_lift`; warnings are one-time messages and can't be lifted.  (I realize it was already like this when you got here.)

cellio (approved):
LGTM.  One small wording request, but the wording was already confused and you followed suit, so it's ok to do that separately instead of creeping scope here.

Oaphi:
> Are we able to distinguish between warnings and suspensions?

I suppose so, just needs some discussion as to whether it's worth it - @ArtOfCode-?

app/controllers/mod_warning_controller.rb
Oaphi: Yeah, suspensions probably need to be delineated more - currently, it's a bit hard to reason about them given that they are treated as special warnings

ArtOfCode-:
> > Are we able to distinguish between warnings and suspensions?
> 
> I suppose so, just needs some discussion as to whether it's worth it - @ArtOfCode-?

I'd agree - let's have separate `warning_create` and `suspension` events. I don't think it needs to be one or the other - i.e. if we suspend someone, that could create two audits, a `warning_create` and a `suspension`, which may make implementation easier.

cellio:
> I'd agree - let's have separate `warning_create` and `suspension` events. I don't think it needs to be one or the other - i.e. if we suspend someone, that could create two audits, a `warning_create` and a `suspension`, which may make implementation easier.

Sounds good to me.  Two entries for a suspension, one for the message and one for the actual suspension, are fine, and allows the suspension (only) to be paired with a cancelled suspension.  It also allows mods to easily see just the list of who's been suspended.

Oaphi:
Done - we now add the following audit logs:

- `warning_create` for _any_ new mod warning;
- `suspension_create` if the created mod warning is a suspension;
- `warning_lift` for _any_ warning being lifted (it's technically possible to lift non-suspension warnings);
- `suspension_lift` if the lifted mod warning is a suspension;

Audit logs after creating a warning that is also a suspension:

Screenshot from 2025-10-07 05-24-44

Audit logs after lifting a warning that is also a suspension:

Screenshot from 2025-10-07 05-27-04

ArtOfCode- (approved):

PR #1867, opened 2025-10-05
Fix tag drafts not being deleted upon creation / update

### LINKED ISSUE #1859: Drafts are not cleared after creating tags
Likely a side-effect of purging client-side removal of drafts upon post submission: we ensure draft is deleted server-side (which is the correct behavior), but we don't for tags.

Oaphi:
closes #1859

This PR adds a `DraftManagement` concern to both facilitate the fix and avoid duplication (all our controllers share the same draft saving & deletion logic). The concern is named this way to match Rails's convention of naming such controller concerns (see, for example, `RequestForgeryProtection`)

### DIFF
--- a/test/controllers/tags_controller_test.rb
+++ b/test/controllers/tags_controller_test.rb
@@ -111,12 +111,21 @@ class TagsControllerTest < ActionController::TestCase
 
     users.each do |user|
       sign_in(user)
-      try_create_tag(category, tag_set, name: "test-tag-#{user.id}")
+
+      draft_path = create_tag_path(id: category.id)
+      tag_name = "test-tag-#{user.id}"
+
+      @controller.stub('params', { tag_name: tag_name }) do
+        @controller.do_save_draft(user, draft_path)
+      end
+
+      try_create_tag(category, tag_set, name: tag_name)
 
       if !@controller.helpers.user_signed_in?
         assert_redirected_to_sign_in
       elsif user.can_edit_tags?
         assert_response(:found, "Expected user '#{user.name}' to be able to create tags")
+        assert_draft_deleted(user, draft_path, :tag_name)
       else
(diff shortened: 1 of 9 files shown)


test/controllers/tags_controller_test.rb
```diff
@@ -111,12 +111,21 @@ class TagsControllerTest < ActionController::TestCase
...
     users.each do |user|
       sign_in(user)
-      try_create_tag(category, tag_set, name: "test-tag-#{user.id}")
+
+      draft_path = create_tag_path(id: category.id)
+      tag_name = "test-tag-#{user.id}"
+
+      @controller.stub('params', { tag_name: tag_name }) do
+        @controller.do_save_draft(user, draft_path)
+      end
```
Oaphi, line 120: This is an unfortunate mess, but doing this "the proper way" would require quite a bit of work that's outside of the scope of the PR (although I am not against making these changes if necessary): move all this to integration tests, move `save_draft` and `delete_draft` to concern's actions, etc.

test/controllers/tags_controller_test.rb
```diff
@@ -276,6 +285,9 @@ class TagsControllerTest < ActionController::TestCase
   # @param category [Category] category to create the tag in
   # @param tag_set [TagSet] tag set the tag belongs to
   def try_create_tag(tag_set, category, **opts)
+    # this is needed for draft deletion until we switch from request.referer
+    @request.set_header('HTTP_REFERER', create_tag_path(id: category.id))
```
Oaphi, line 289: And this is only here because we rely on `request.referer` when creating and deleting drafts. I'd like to work on switching from that approach separately, but open to doing it "while we are at it" too.

ArtOfCode- (approved):

PR #1870, opened 2025-10-08
Add complaints/reporting procedure

### LINKED ISSUE #1868: Create a complaints/reporting process outside of the existing moderation toolset
Labelled as complexity: average because while this shouldn't be too difficult, there's quite a bit of work involved.

Requirements:
- [x] Does _not_ have to be accessible directly on every post, but must be linked in the footer
- [x] Allows users to make reports or complaints directly to staff users
- [x] Does not require users to be signed in to make a report
- [x] Allows several types of reports, including:
    - [x] Appeal against Foundation decision (consider how this can be broadened to not-us installations)
    - [x] Illegal content report
- [x] Allows staff and reporters to respond to reports, creating a conversation log
    - [x] Does _not_ need to handle user replies via email; emails may direct users to log in to respond.
- [x] Allows a member of staff to be assigned to a report to handle it
- [x] Allows members of staff to add non-public comments to a report, adding to the log but _not_ showing or emailing these to reporters
- [x] Allows staff to mark reports as closed once they are dealt with, preventing further responses
- [x] Sends emails:
    - [x] Allows users to opt out of receiving return emails about their report, both at time of initial reporting and subsequently
    - [x] Confirms receipt of report to reporter
    - [x] Confirms to reporter when report has been reviewed

ArtOfCode-:
Closes #1868.


ArtOfCode-:
Git screwup - I've managed to include some old release commits in here, but they're all already merged in so no harm.

app/views/complaints/show.html.erb
```diff
@@ -0,0 +1,65 @@
...
+<div class="grid">
+  <div class="grid--cell is-12 is-6-md is-6-lg">
+    <div class="widget">
+      <div class="widget--body">
+        <h2 class="has-font-size-base h-m-t-0">Report</h2>
+        <div class="grid has-font-size-caption">
+          <div class="grid--cell is-3">Reported URL</div>
+          <div class="grid--cell is-9 has-color-yellow-700">
+            <i class="fas fa-exclamation-triangle"></i>
+            <%= link_to @complaint.reported_url, @complaint.reported_url, class: 'is-yellow' %>
```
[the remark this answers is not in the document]
ArtOfCode-: Confirmed.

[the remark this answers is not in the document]
ArtOfCode-: Resolved by 622fdfe.

cellio:
After switching the branch I went to Reports (of course there are none yet) and got this:

```
ActionView::Template::Error (undefined method `[]' for nil:NilClass

    AppConfig.safety_center['statuses'].map { |k, v| [v['name'], k] }
                           ^^^^^^^^^^^^):

Causes:
NoMethodError (undefined method `[]' for nil:NilClass

    AppConfig.safety_center['statuses'].map { |k, v| [v['name'], k] }
                           ^^^^^^^^^^^^)
    10:       <div class="form-group">
    11:         <%= label_tag :status, 'Status' %>
    12:         <%= select_tag :status,
    13:                        options_for_select(statuses_select_options, selected: params[:status]),
    14:                        include_blank: true, class: 'form-element js-complaint-status-select' %>
    15:       </div>
    16: 
  
app/helpers/complaints_helper.rb:3:in `statuses_select_options'
app/views/complaints/reports.html.erb:13
app/views/complaints/reports.html.erb:8
app/controllers/complaints_controller.rb:98:in `reports'

```

Presumably need to make something nil-safe.

I then tried to move on, but clicking "report" from the footer hits another nil-unsafe path:

```
NoMethodError (undefined method `[]' for nil:NilClass

    @report_types = AppConfig.safety_center['report_types'].select { |_k, t| t['enabled'] }
                                           ^^^^^^^^^^^^^^^^):
  
app/controllers/complaints_controller.rb:12:in `report'
```

In addition to fixing these, we probably want to add test cases to confirm that a new deployment won't have these issues.

Oaphi:
> After switching the branch I went to Reports (of course there are none yet) and got this:

More like we need to prepopulate the `safety_center` config YAML file - these config files are supposed to always be present (probably won't hurt to make all the config fetches `nil`-safe, though).

Hmm, seems like the file is present?

cellio:
Answer: I needed to restart the server to pick up the changes, oops.

We probably still want to make these nil-safe, though.

cellio:
Love the new alerts for staff!  (Someday we might need to differentiate more finely; the marketing intern probably shouldn't see these alerts.  But that can be a later consideration.)

cellio:
I used the letter opener to check what mail we sent.  The "report received" message includes a link, but that link goes to `meta.codidact.com` and is thus 404 on my server.  Whatever we do to get the correct domain for new-account confirmation email, we need to do here too.

cellio:
On the Reports page, the default filter is "", which shows none of them.  I had to choose "Reviewed" to see the report I just handled.  Could we show them all by default and let people restrict, like we do elsewhere?

cellio:
INFO: one of the options we offer is to dispute an action taken against you.  How does a user know that an action of this type has been taken?  In particular, there's nothing visible to the user that indicates that content was deleted for this reason versus mod deletion.

cellio:
I submitted a report while logged in but opted out of email notifications.  We sent email anyway.

cellio:
Sorry about the noise.  I'll try to collect testing notes in one place with checkboxes so you can mark them off as they're addressed.

- [x] Make [these things](https://github.com/codidact/qpixel/pull/1870#issuecomment-3408690481) nil-safe.
- [x] Suppose I'm a staff member (no admin privileges, but support staff) and I receive a report to handle. After I investigate and determine that action needs to be taken above my pay grade, how do I route it to an admin? I looked for a "reassign" or "escalate" option but didn't find it.
- [x] I used the letter opener to check what mail we sent.  The "report received" message includes a link, but that link goes to `meta.codidact.com` and is thus 404 on my server.  Whatever we do to get the correct domain for new-account confirmation email, we need to do here too.
- [x] On the Reports page, the default filter is "", which shows none of them. I had to choose "Reviewed" to see the report I just handled. Could we show them all by default and let people restrict, like we do elsewhere?  (Update: *when* there are new reports it shows them, with the filter of "" not "New".  If we're going to focus on new ones, which is legit, let's show "new" in the dropdown and add an "all".)
- [x] One of the options we offer is to dispute an action taken against you. How does a user know that an action of this type has been taken? In particular, there's nothing visible to the user that indicates that content was deleted for this reason versus mod deletion.
- [x] I submitted a report while logged in but opted out of email notifications. We sent email anyway.
- [x] All the language is oriented around complaints, as expected.  But when reviewing an appeal, I didn't know which option to choose -- does "upheld" mean I agree with the appeal (overturning the original action), or does it mean I looked again and the original complaint was correct and your appeal is declined?  We might need to special-case this, also because...
- [x] The email we send for resolved appeals refers to the "content you reported".

Oaphi:
Made a couple of suggestions partially addressing issues mentioned in @cellio's [comment](https://github.com/codidact/qpixel/pull/1870#issuecomment-3408906706): https://github.com/codidact/qpixel/pull/1879

ArtOfCode-:
@cellio 

> On the Reports page, the default filter is "", which shows none of them.

The default filter there is by status for `['new', 'assigned', 'responded']`, so by default shows only _active_ reports (i.e. reports which have not been reviewed or closed). If you'd already reviewed the report it wouldn't show up there. I should be able to change the blank option to make that clearer.

> I submitted a report while logged in but opted out of email notifications. We sent email anyway.

That was a deliberate decision - we still send the report-received email as confirmation, but nothing further (the report-reviewed email should not have sent - did it?). I'll add something to make that clearer on the form.

cellio:
> The default filter there is by status for `['new', 'assigned', 'responded']`, so by default shows only _active_ reports (i.e. reports which have not been reviewed or closed). If you'd already reviewed the report it wouldn't show up there. I should be able to change the blank option to make that clearer.

Ah, I totally failed to figure that out, but it makes sense to default to that.  "All open reports" is great, thanks.  Still might later want an "All", but that's less important with the existence of a single "waiting for us" option.

> > I submitted a report while logged in but opted out of email notifications. We sent email anyway.
> 
> That was a deliberate decision - we still send the report-received email as confirmation, but nothing further (the report-reviewed email should not have sent - did it?). I'll add something to make that clearer on the form.

Ah, that makes sense.  And yes, we did not send the report-reviewed email, so all good here (with that language update you mentioned).

app/controllers/complaints_controller.rb
```diff
@@ -154,6 +154,35 @@ def update_status
...
+
+    new_content_type = helpers.content_type(params[:new_content_type])
+    if new_content_type.nil?
+      flash[:danger] = 'Invalid content type.'
+      redirect_back fallback_location: complaint_path(@complaint.access_token) and return
+    end
+
+    update_params = { content_type: params[:new_content_type] }
+    message = "Content type changed to #{new_content_type['name']} by #{current_user.username}."
+    @comment = @complaint.comments.new(user_id: -1, internal: true, content: message)
```
cellio: FYI, @Oaphi just added a helper to get the system user, to avoid hard-coding constants.

ArtOfCode-:
@cellio that should be all your notes completed except this one:

> I used the letter opener to check what mail we sent. The "report received" message includes a link, but that link goes to meta.codidact.com and is thus 404 on my server. Whatever we do to get the correct domain for new-account confirmation email, we need to do here too.

Reports aren't specifically attached to a community because they're not handled by the community. They're also not attached (in the database sense) to a post, they just have an associated URL. I've used the correct host for the new legal-deletion email because that has an associated post, but for generic report emails we have to fall back on the default host (which is specified in config, and should be changed by anyone deploying QPixel that isn't us).

cellio:
> Reports aren't specifically attached to a community because they're not handled by the community. They're also not attached (in the database sense) to a post, they just have an associated URL. I've used the correct host for the new legal-deletion email because that has an associated post, but for generic report emails we have to fall back on the default host (which is specified in config, and should be changed by anyone deploying QPixel that isn't us).

Oh, I see!  I found the config file and edited it, confirming that it works as you said.  (Note to self: we should add something to the installation guide pointing to that, so other instances don't accidentally redirect to us.)

Is "add reply" supposed to send email?  I don't see any.  We send email upon submission and when it's resolved, but if we have questions while investigating, I thought adding a reply would notify the reporter.

ArtOfCode-:
> Is "add reply" supposed to send email? I don't see any. We send email upon submission and when it's resolved, but if we have questions while investigating, I thought adding a reply would notify the reporter.

I went back and forth on this. That was the original plan, but I thought that could result in quite a few emails. I could be steered either way on this, I think.

cellio:
> > Is "add reply" supposed to send email? I don't see any. We send email upon submission and when it's resolved, but if we have questions while investigating, I thought adding a reply would notify the reporter.
> 
> I went back and forth on this. That was the original plan, but I thought that could result in quite a few emails. I could be steered either way on this, I think.

Putting myself in the position of someone reporting something I'm very concerned about, I think I would be unhappy if I sent a report into the void and opted in to email, time passed because I didn't know they needed more info, and the next thing I heard was that the issue has been resolved as disputed because it was unclear.  I would rather send email (if the reporter opted in) saying there's a response we need you to look at.   We don't expect most tickets to do this at all, and for the ones that do, the volume and cycle time are driven by humans, so there shouldn't be floods of email.

cellio:
Testing update:

- All the user-visible items from https://github.com/codidact/qpixel/pull/1870#issuecomment-3408906706 are addressed.  (That's all but the first; the first is a code-review point that looks to be addressed as well but I didn't test specifically for that.)
- I agree that https://github.com/codidact/qpixel/pull/1870#issuecomment-3412289623 was pilot error (no bug).
- Confirmed that "add reply" now sends email if the person opted in and warns about it if opted out (https://github.com/codidact/qpixel/pull/1870#issuecomment-3412555232).
- Tested legal deletions (that staff can, that mods can't, that history is redacted).

New issues (update: now fixed)
- Moderators can restore legal-deleted posts, though the original content is still redacted.  (Minor, but would be nice to prevent, analogous to how curators can't restore mod-deleted posts.)
- Some confusion about the "reporter responded" state that is being sorted out as I write this.

Everything else looks great!  (Have not done any code review, just testing.)

cellio:
One more small thing -- as staff, I was able to legal-delete a post that was already legal-deleted.  That would be a no-op except that (1) it updates the deletion time and attribution and (2) it generates a new (reviewed) report.

This might not be fixable without a bigger change, because we probably *do* need staff to be able to legal-delete regularly-deleted posts.  You'd know more about this than I do, but I'm guessing that "still visible to curators" won't be good enough for OSA or DMCA take-downs.  It's ok with me if we defer addressing this, though it's something to bear in mind for reporting if we need to report out.

app/jobs/auto_close_complaints_job.rb
```diff
@@ -0,0 +1,11 @@
+class AutoCloseComplaintsJob < ApplicationJob
+  queue_as :default
+
+  def perform(*_args)
+    complaints = Complaint.where(status: 'reviewed')
+                          .where(Arel.sql('status_updated_at <= ?', 14.days.ago))
```
cellio, line 6: I seem to recall a past review where I was confused by a `<=` that felt like it should be `>=` or vice versa, and that the answer was some sort of rails/sql magic.  Since I can't test the script (without hackery), I just want to double-check that this operator is correct despite how it looks to this less-experienced reader?

app/mailers/complaints_mailer.rb
```diff
@@ -0,0 +1,33 @@
+class ComplaintsMailer < ApplicationMailer
+  helper :application, :post_types, :complaints
+
```
cellio, line 3: Ok these obviously work, but I'm confused where these functions get/construct the message bodies?  Is it worth explaining in a comment for future readers?

app/models/complaint.rb
```diff
@@ -0,0 +1,65 @@
...
+  ##
+  # Update the complaint's status, create a comment to record the change, and send emails to the right people.
+  # @param new_status [String] The new status to set, from safety_center.yml.
+  # @param attribute_to [String] Who should the status change be attributed to? Username only.
+  def update_status(new_status, attribute_to = nil)
+    dt = DateTime.now
+    update(status: new_status, status_updated_at: dt)
+    attribution = attribute_to.nil? ? 'automatically' : "by #{attribute_to}"
+    comments.create(content: "Status updated to #{new_status} at #{dt.iso8601} #{attribution}.", internal: true,
+                    user_id: -1)
```
cellio: Here's another hard-coded `-1`, if you're making updates anyway.  (NOT a blocker.)

app/views/complaints/report.html.erb
```diff
@@ -0,0 +1,127 @@
...
+<p>
+  Thank you for taking the time to make a report. If you've seen harmful, abusive, or illegal content on our
+  communities, you can report this to us here. You can also use this page if you've received a message saying we've
+  classified your content as harmful, abusive, or illegal and you wish to contest it.
+</p>
+
+<div class="notice is-warning has-color-yellow-900 flex-row-always ai-c has-font-size-caption h-m-b-4">
+  <i class="fas fa-exclamation-triangle"></i>
+  <p>
+    Please don't use this tool to complain about moderator actions, including deletion of posts. Raise these in the
```
cellio: Nit: let's say "moderation actions", not "moderator actions", to include stuff that curators and other non-mods can do.

app/views/complaints/report.html.erb
```diff
@@ -0,0 +1,127 @@
...
+      </div>
+      <%= text_field_tag :email, user_signed_in? ? current_user.email : nil, class: 'form-element',
+                         disabled: user_signed_in?, required: true %>
+    </div>
+
+    <div class="form-group">
+      <%= label_tag :updates, 'Do you wish to receive updates about this report?' %>
+      <div class="form-caption">
+        We will let you know once we have investigated your report, and once we've taken any necessary action and closed
+        your report. If you do not wish to receive these updates, please untick the box. You will always receive an
```
cellio: Nit: technically, we will always send confirmation, but we can't promise that the reporter will receive it.  I'd change this to "We will always send email to confirm..." for clarity (nit) and to avoid making promises out of our control (I hope a nit but one never knows how things will be interpreted by legal authorities).

config/config/safety_center.yml
```diff
@@ -0,0 +1,150 @@
+statuses:
```
cellio, line 8: We declare these statuses in a config file, but we also have code that depends on these literal names.  Assuming that YAML files support comments, can we add one here to say that?  I assume that deployments could add to this list without harm, but if they remove something or change names (e.g. to non-English ones), things will break, right?

config/config/safety_center.yml
```diff
@@ -0,0 +1,150 @@
...
+        content: our community team have reviewed your report and have decided that the content you reported is not
+          actionable. This does not mean that the content cannot be removed by community moderators for breaking
+          community-specific rules, just that it does not breach our Content Policy.
+      abusive: *illegal_disputed
+      copyright: *illegal_disputed
+      appeal:
+        content: our community team have reviewed your appeal and have decided that the action taken in your case was
+          appropriate.
+
+report_types:
```
cellio, line 73: I think these are also hard-coded in places too, so a comment to that effect would be good.

config/config/safety_center.yml
```diff
@@ -0,0 +1,150 @@
...
+    description: abusive or otherwise harmful content
+  copyright:
+    enabled: true
+    name: Report of copyright infringement
+    description: copyright infringement
+  appeal:
+    enabled: true
+    name: Classification Appeal
+    description: an appeal regarding how we've handled your content
+
```
cellio, line 90: Suggestion: comment citing Ofcom for this list?  (Signpost for other deployments who need to comply with OSA, and indicator that we didn't just make this up.)

test/models/complaint_comment_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class ComplaintCommentTest < ActiveSupport::TestCase
```
cellio, line 3: Was there supposed to be a test here?  (Also applies to the model.)

cellio (commented):
Looks great!  We should discuss if https://github.com/codidact/qpixel/pull/1870#issuecomment-3413202794 is a concern now; other than that and a few small comments as part of this review, LGTM.

I reviewed ruby code but not JS.

app/jobs/auto_close_complaints_job.rb
ArtOfCode-: Yep. `14.days.ago` resolves to a specific date/time, as does `status_updated_at`. In other words, it's saying "(absolute date) is less than or equal to (absolute date)", and "less than" for dates means "before".

app/mailers/complaints_mailer.rb
ArtOfCode-: That's a Rails strcture thing - they're in `app/views/complaints_mailer`.

test/models/complaint_comment_test.rb
ArtOfCode-: I hadn't written any, no. This just gets auto-generated by Rails when you generate the model.

ArtOfCode-:
> One more small thing -- as staff, I was able to legal-delete a post that was already legal-deleted. That would be a no-op except that (1) it updates the deletion time and attribution and (2) it generates a new (reviewed) report.

I'm happy to defer this if you are. You're right that we need staff to be able to legal-delete already-deleted posts, so just disallowing this for deleted posts isn't the answer and - as discussed - there's no flag to differentiate the two.

cellio:
> > One more small thing -- as staff, I was able to legal-delete a post that was already legal-deleted. That would be a no-op except that (1) it updates the deletion time and attribution and (2) it generates a new (reviewed) report.
> 
> I'm happy to defer this if you are. You're right that we need staff to be able to legal-delete already-deleted posts, so just disallowing this for deleted posts isn't the answer and - as discussed - there's no flag to differentiate the two.

Filed https://github.com/codidact/qpixel/issues/1880 to track this.

cellio (approved):
LGTM!

PR #1873, opened 2025-10-10
expand who can rename threads, with public log

### LINKED ISSUE #714: Allow thread titles to be edited (with public record)

cellio:
Currently, only curators can rename comment threads.  We've had many requests over the years from authors who didn't like the auto-generated title or wanted to fix a typo, but renames were also silent, so we wanted to restrict this to mitigate against abuse.  (After all, a silently-renamed thread puts that name in the mouth of the person who created that thread.)

This change adds attribution for renamed threads, like this:

comments showing who renamed from what to what

The comment is owned by System, not the renamer, so that regular users can't delete the comments.  These comments are created directly and do not ping followers (just as we do not notify followers for other state changes like lock and archive).

Non-curator users can only rename threads they've participated in, like this:

two threads, one showing the rename control and one not

Fixes https://github.com/codidact/qpixel/issues/714.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -220,13 +220,26 @@ class CommentsController < ApplicationController
       return
     end
 
+    orig_title = @comment_thread.title
     title = helpers.strip_markdown(params[:title], strip_leading_quote: true)
     status = @comment_thread.update(title: title)
+
+    if status
+      # Comment is owned by System so regular users can't delete it. Without
+      # this record, the title would be attributed to the thread creator,
+      # which can be abused.
+      log_msg = Comment.new(post: @post, content: 'Thread renamed from \"' + orig_title + '\" to \"' + title + '\" by ' + current_user.username, user: User.find(-1), comment_thread: @comment_thread, has_reference: false)
+      comment_status = log_msg.save
+    end
 
     unless status
       flash[:danger] = I18n.t('comments.errors.rename_thread_generic')
     end
 
+    unless comment_status
+      flash[:danger] = I18n.t('comments.errors.comment_not_posted')
+    end
+      
     redirect_to comment_thread_path(@comment_thread.id)
   end
 
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -228,7 +228,13 @@ class CommentsController < ApplicationController
       # Comment is owned by System so regular users can't delete it. Without
       # this record, the title would be attributed to the thread creator,
       # which can be abused.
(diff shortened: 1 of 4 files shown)


app/controllers/comments_controller.rb
```diff
@@ -220,13 +220,32 @@ def thread_rename
...
     title = helpers.strip_markdown(params[:title], strip_leading_quote: true)
     status = @comment_thread.update(title: title)
 
+    if status
+      # Comment is owned by System so regular users can't delete it. Without
+      # this record, the title would be attributed to the thread creator,
+      # which can be abused.
+      log_msg =
+        Comment.new(post: @post,
+                    content:
```
cellio, line 233: Pretty sure the line wasn't too long before rubocop insisted on "fixing" it, hence this extra line break.

Oaphi (dismissed):
I believe the latest set of commits is currently missing (we've definitely discussed schema changes and not hardcoding the renamer user's username in the comment's message) - I don't recall if those commits were supposed to be added later, RC'ing just so as we don't merge without them by accident.

cellio:
> I believe the latest set of commits is currently missing (we've definitely discussed schema changes and not hardcoding the renamer user's username in the comment's message) - I don't recall if those commits were supposed to be added later, RC'ing just so as we don't merge without them by accident.

I've fixed the literal name in favor of a link as we mentioned.  If the renaming user is deleted, this shows "deleted user" as expected, without leaking the original name.

Arguably the schema changes could be done separately; we already had renaming and already failed to record the user and time.

Oaphi (approved):
LGTM

ArtOfCode- (approved):

PR #1878, opened 2025-10-14
0valt/1875/audit logs

### LINKED ISSUE #1875: Audit log table is a bit wider than filters / sorting
Backlogging so as I do not forget - likely related to recent fixes to wrapping (we no longer break "anywhere" in the type, event, and related columns):

Image

Oaphi:
closes #1875

This PR reworks the audit log table to align with how our error reports look like:

Screenshot from 2025-10-15 02-13-12

Screenshot from 2025-10-15 02-23-57

This makes the logs easier to parse at a glance, solves wrapping and overflow issues, and allows us to make the logs as verbose as we want.

It also makes any `related` object that is a `User` a proper link (we should do that for all other types too, but that's a bit out of scope for this PR).

### DIFF
--- a/app/assets/stylesheets/audit_logs.scss
+++ b/app/assets/stylesheets/audit_logs.scss
@@ -1,7 +1,9 @@
 @import 'variables';
 
 .audit-log-filters {
-  .form-group-horizontal .actions, .select2-container {
+
+  .form-group-horizontal .actions,
+  .select2-container {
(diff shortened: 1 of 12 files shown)


cellio (approved):
Very nice change!  I reviewed the ruby code but not the JS.

ArtOfCode- (approved):

PR #1886, opened 2025-10-22
flag history view should indicate that you are viewing another user's…

### LINKED ISSUE #1885: User flag history page should indicate that you are viewing another user's history
For users who can view flag history pages for other users, we should indicate that this is the _other_ user's page, not the current one's (took a solid minute to realize that we haven't broken this view):

Image

Oaphi:
closes #1885 

Here's how current user's page header will look like:

Screenshot from 2025-10-22 07-17-47

and here's how it will look like when viewing the same page for another user:

Screenshot from 2025-10-22 07-17-42

As a bonus, the PR makes the view's descriptions configurable via translation strings.

### DIFF
--- a/config/locales/strings/en.flags.yml
+++ b/config/locales/strings/en.flags.yml
@@ -2,8 +2,8 @@ en:
   flags:
     descriptions:
       history_other: >
-        Below is a list of all the flags that the user have cast,
-        the outcome of each them, and handling moderators' feedback, if any.
+        Below is a list of all the flags this user has cast,
+        the outcome of each of them, and handling moderators' feedback, if any.
       history_self: >
         Below is a list of all the flags you have cast and the outcome of each of them.
         Moderators handling your flags may also have left you feedback.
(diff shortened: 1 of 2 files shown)


config/locales/strings/en.flags.yml
```diff
@@ -1,5 +1,12 @@
 en:
   flags:
+    descriptions:
+      history_other: >
+        Below is a list of all the flags that the user have cast,
```
cellio: Language corrections: "has cast" (not "have"), "each of them" (missing "of").

Suggestion: "all the flags this user..." (as opposed to "that the user") -- just feels like it's a little more specific when we're on one user's page.

Suggestion: "...and any feedback from the moderator who handled the flag".  (Only one mod can handle and thus leave feedback, and "handling" as an adjective could be a small speed bump for people less fluent in English who might try to pass it as a verb form.)

cellio (commented):
Thanks for moving these strings from code to an i18n file.

cellio (approved):
LGTM.  If you want to make them parallel, you could change the "self" text to: "Below is a list of all the flags you have cast with their outcomes and any moderator feedback."

ArtOfCode- (approved):

Issue #1664, opened 2025-06-27
Use network name in email subjects without baking in "Codidact"


cellio:
https://github.com/codidact/qpixel/pull/1663 made subject lines on email we send less generic, because a lot of "Confirmation instructions" email was getting marked as spam and I thought making it more specific would help avert the automatic reaction.  However, that made the problem of hard-coding the string "Codidact" worse.

The subject lines are in a strings YAML file, and I don't know of a way to make those dynamic.  We need to have the code that sends the email construct a subject line from `ENV.fetch('NETWORK_NAME', 'Codidact')` and a generic string (and make the strings generic again).

Issue #1396, opened 2024-09-08
Reference to the Codidactyl in non-Codidact instances


trichoplax:
**Describe the bug**
When viewing the user page for a deleted user, the following message is displayed, which may be confusing for a user of an instance other than Codidact:

> User accounts make an excellent snack for the Codidactyl.

**To Reproduce**
1. Create a user account in a development environment.
2. Delete the user, making a note of their user id.
3. View your own user page.
4. Change the URL to use the deleted user's user id instead of your own.
5. See a deleted user message, including the message mentioning the Codidactyl.

**Expected behavior**
Hardcoded messages should be suitable for all instances, not just Codidact. Anything that only applies to Codidact should be conditional, so that it only appears when viewed from the Codidact instance.

Personally, I would also question whether we should have in-jokes at all in the user interface as the Codidact communities grow, as new users will have no context.

cellio:
Agreed this should be configuration or generic.  It's fine to have a "mascot" that shows up in passing, and maybe we should have a "what is the Codidactyl?" help page that we can link to from such mentions, but if it gets in the way of comprehension, we should adjust.  I love the Codidactyl, but we want it to be friendly, not befuddling.

PR #1888, opened 2025-10-26
Assorted improvements

### LINKED ISSUE #1900: Preserve sort order when applying a filter
meta:293964

Post lists (category page, search results, maybe others) have sort buttons -- activity, score, etc.  When these pages have filters, applying the filter resets the sort to the default.  From the user perspective, a filter is "refine what I'm seeing" more than "show me a new set of stuff", so it is surprising to have the sort change.

Filtering and sorting are orthogonal concepts; can we make them orthogonal in the code too?

Oaphi:
This PR includes assorted improvements in preparation for more significant changes:

- `do_soft_delete` method on the `User` model is renamed to `soft_delete` to match the naming convention (`do_*` methods must have a counterpart without `do_`);
- soft deletion tests are expanded to cover both `profile` and `user` deletions;
- `CommunityUser` now has a corresponding `soft_delete` method;
- HTML markup of the deletion confirmation email is fixed (we were missing a closing `</li>` tag);
- two new _global_ site settings are added: `NetworkURL` and `NetworkName` & as many views as feasible have been switched to them from the hardcoded `https://codidact.com/` and `Codidact` strings (related: #1396, #1664);
- a new `User.system` static method is added to simplify system user lookup;
- tests no longer attempt to make requests to external URIs  (system tests are still allowed to do so as they need the assets to work) thanks to the [WebMock](https://github.com/bblimke/webmock) gem (test env only);
- 2FA mailer actions are now fully covered with tests;
- a fix for the regular expression checking that the post contains image links with no alt text (see #1889 - not closing it as we can benefit from further improvements to validation logic there, but at least the main concern is addressed by the PR);
- current sorting option is no longer lost upon applying a filter (closes #1900);
- comment pings now properly register and render for system users (re: meta:294917):
    
- applied search filter select is now properly styled:
    image
    Compared to:
    image

- search filters select & actions now render normally on narrow screens:
    Screenshot from 2025-11-08 01-35-27
    Compared to:
    image

- comment threads now properly wrap actions on narrow screens (and are more responsive overall):
    Screenshot from 2025-11-11 00-07-17
    Compared to:
    Screenshot from 2025-11-11 00-07-31
    And to stuff like this in prod:
    image

To be able to test the changes, it is _necessary_ to:

- run `bundle install` to add the new gem (tests will break without it);
- run `rails db:seed` to populate the new site setting;

A request for reviewers: we need to _thoroughly_ check that the mailers actually pick up the setting, especially the donation ones. I checked some, but not all _yet_.


cellio:
New site setting shows up as "uncategorized" -- can we improve on that?  "Site details" seems plausible.

screenshot of new setting

cellio:
Testing notes (will edit this comment as I go):

- [ ] `NetworkName` in email?
    - [x] new-user signup: confirmed
    - [x] self-deleted account: confirmed
    - [x] password change: does not use a network name, but does have a hard-coded URL for resets that was already present (not a regression).  I initially thought we could use the new NetworkURL setting instead, but the URL in the email is to a particular community, so that won't work.
    - [x] email to all moderators: confirmed has network name in subject line 
    - [ ] email to all users: did not previously include network name and doesn't now either (someday we might make that consistent with the mod email, but can wait) -- see details below
    - [ ] subscription email: ? I haven't gotten my dev server to send subscription email at all, so I can't check the content, but this looks like all the other email changes so don't block on this.
    - [ ] not sure how to test donation email without setting up actual Stripe stuff in the dev env? (I don't have an API key)
    - [x] legal deletion: confirmed

- [x] Account deletion: 
   - [x] self-delete and mod delete work as expected
   - [x] ~~haven't checked yet if this is a regression, but if a mod deletes a profile on one community, the "all communities" list for the user still shows the link to everyone (mods see the deleted user, others see the skull & crossbones) -- I *think* this is what we did already but noting it here to confirm~~ - not a regression (this is what we do already)

- sanity check on system user change: renamed a comment thread and the comment was correctly attributed to System (this is an existing use of that code that I knew about)

--

Regarding the admin email to mods/all users, this is how the subject lines and senders look now.  Mod mail includes the network name; all-user mail doesn't.  We can update it here or spin it off -- don't care which, just noticed in passing.

screenshot

db/seeds/site_settings.yml
```diff
@@ -681,6 +681,13 @@
     The content of a post is shown in short in lists (e.g. category post overview or in search).
     This setting controls how many characters of a post are shown.
 
+- name: NetworkName
+  value: Codidact
+  value_type: string
+  community_id: ~
```
cellio, line 688: add:
`category: SiteDetails`

app/controllers/users_controller.rb
```diff
@@ -326,17 +326,15 @@ def soft_delete
 
     case params[:type]
     when 'profile'
-      AuditLog.moderator_audit(event_type: 'profile_delete', related: @user.community_user, user: current_user,
```
cellio, line 329: Breadcrumb: auditing was moved into `soft_delete`, not removed (confirmed that audit logs are still generated).

db/seeds/site_settings.yml
Oaphi: Done: 

Screenshot from 2025-10-28 05-53-15

Note that since our seeds are, unfortunately, append-only when it comes to entities, you'll have to manually drop the site setting from the database (i.e., `delete from site_settings where name = 'NetworkName';`) and then run seeds again for the change to take effect.

app/models/user.rb
```diff
@@ -476,12 +482,16 @@ def active_flags(post)
     post.flags.where(user: self, status: nil)
   end
 
-  def do_soft_delete(attribute_to)
+  # Soft-deletes the user (username, password, and email are irrevocably reset!)
```
cellio, line 498: maybe note in the comment that the reason these are irrevocably reset is data privacy laws (we don't want future editors to decide to "fix" this "bug")

app/views/application/keyboard_tools.html.erb
```diff
@@ -1,9 +1,13 @@
 <h1>Keyboard tools</h1>
```
cellio: Haven't confirmed visually because I don't yet know how to navigate to this page...

app/views/devise/registrations/new.html.erb
```diff
@@ -8,7 +8,8 @@
 <% end %>
 
 <div class="notice is-info">
-  <i class="fas fa-info-circle"></i> If you have an account on another <a href="https://codidact.com/">Codidact site</a>,
+  <i class="fas fa-info-circle"></i> If you have an account on another 
+  <a href="https://codidact.com/"><%= SiteSetting['NetworkName'] %> site</a>,
```
cellio: Is there a setting for that URL, or is that a problem for later?

app/views/donation_mailer/donation_successful.text.erb
```diff
@@ -11,8 +11,8 @@ exactly what we've received and how we've spent it, which you'll find on Meta (h
 
 If you're not already, get involved! Here are some more things you can do:
 
- * Join our mailing list (https://codidact.org/)
- * Browse and join our communities (https://codidact.com/)
+* Join our mailing list (https://codidact.org/)
+* Browse and join our communities (https://codidact.com/)
 
 Thanks and see you around,
-The Codidact Team
```
cellio, line 18: If some other network sets up donations, they'll go to them not us, right?  Meaning the baked-in text in the email (with our URLs) is wrong but we're not actually intercepting their donations?  In that case, I agree we should use the setting here (and the problem of the incongruous email can be a future problem since it already exists).

app/views/two_factor_mailer/backup_code.html.erb
```diff
@@ -3,19 +3,19 @@
...
-  recently added backup codes for every account using 2FA, so that you can still log in if you lose access to your
+  You're receiving this email because you have two-factor authentication enabled on your <%= SiteSetting['NetworkName'] %> network account.
+  We've added backup codes for every account using 2FA, so that you can still log in if you lose access to your
   authenticator app.
 </p>
 
 <p>
   Your backup code can be found in your user account settings, under
-  <a href="<%= two_factor_status_url %>">Two-Factor Authentication settings</a>. Please save it somewhere safe.
+  <a href="<%= two_factor_status_url(host: @host) %>">Two-Factor Authentication settings</a>. Please save it somewhere safe.
```
cellio, line 13: What's the effect of this change?  (Haven't tried to test 2fa stuff yet.)

config/locales/strings/en.g.yml
```diff
@@ -24,6 +24,3 @@ en:
     type: 'type'
     user: 'user'
     version: 'version'
-
-  platform:
```
cellio, line 28: Is there a way to confirm that we got all the call sites?  (A grep for "platform" would probably be noisy.)

db/seeds/abilities/mod.html
```diff
@@ -6,4 +6,4 @@ <h2>What does this ability allow me to do?</h2>
 
 <h2>How do I earn this ability?</h2>
 
-<p>Moderators are elected by the community, or appointed by the Codidact Team.</p>
\ No newline at end of file
+<p>Moderators are elected by the community, or appointed by the Codidact Team.</p>
```
cellio, line 9: If you want to remove the reference, you can change this to "appointed by administrators".

cellio (changes requested):
I got through the code and seeds but haven't looked at tests yet.  I'll come back for those, but meanwhile, here are some assorted small comments.

app/models/user.rb
Oaphi: What'd you say if we explicitly called the subaction `anonymize` and made it a proper method? This way we'd get a self-documenting and reusable helper

cellio: Good idea!

app/views/devise/registrations/new.html.erb
Oaphi: No, I don't think we have one yet - was too tired after the main changes to go further than just allowing the name to be configurable, but can while we are at it

app/views/two_factor_mailer/backup_code.html.erb
Oaphi: Should link to the correct host instead of the default one if the email is sent with `deliver_later!` instead of `deliver_now!` (remember the issues with the wrong host in mailed links from a while back?)

config/locales/strings/en.g.yml
Oaphi: You should search for `platform.network_name` - it's how locale strings are usually accessed. I think I updated every single instance, but won't hurt to double-check

app/models/user.rb
Oaphi: Done, `User` model now has a proper `anonymize` method that can act in two modes: save immediately or just update the attributes (needed for the `soft_delete` method)

db/seeds/abilities/mod.html
Oaphi: I'd like to, but this one is in our HTML seeds - I have plans to provide them with template variables that are replaced with configurable values when seeds are run, which is a bit out of scope of this PR

app/views/application/keyboard_tools.html.erb
Oaphi: Apparently, the route got explicitly removed (while the view and the controller action are still there) in https://github.com/codidact/qpixel/commit/9815057d0e089b03d942969034d8836d7f9756e2#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5L215. @ArtOfCode- (just making sure we are all up to speed regarding the removal)?

app/views/devise/registrations/new.html.erb
cellio: Or we can do that separately -- not trying to add scope here.  If we *had* a variable I'd want to use it.  When we add the variable we'll probably find other places that need to use it, so that could end up being as big as the name replacement.

config/locales/strings/en.g.yml
cellio: Thanks -- looks like you got them all.

db/seeds/abilities/mod.html
cellio: Right, seeds can't use variables -- I was saying that instead of "Codidact Team"  we could say "administrators" here, thus side-stepping the issue (no network reference at all).

app/views/devise/registrations/new.html.erb
Oaphi: Done - we now have a new global site setting called `NetworkURL` (not married to the name / description pair):

Screenshot from 2025-10-29 08-07-19

db/seeds/abilities/mod.html
Oaphi: Ah, I just meant that since we have no way of retroactively amending seeded entities (other than posts), we might as well postpone the change to when we can allow seeded entities to be updated.

app/views/donation_mailer/donation_successful.text.erb
Oaphi: > If some other network sets up donations, they'll go to them not us, right?

Correct (thankfully) - as long as they specify _their_ account credentials, donations should go to them.

> Meaning the baked-in text in the email (with our URLs) is wrong but we're not actually intercepting their donations?

Given the above, yup, the baked-in text is misleading for other deployments

app/views/devise/registrations/new.html.erb
cellio: Tested, LGTM.

test/controllers/users_controller_test.rb
```diff
@@ -69,26 +69,63 @@ class UsersControllerTest < ActionController::TestCase
...
+
+    try_soft_delete_user('user', users(:standard_user))
+
     assert_response(:success)
     assert_not_nil assigns(:user)
-    assert_equal true, assigns(:user).deleted
+    assert assigns(:user).deleted
+  end
+
+  test 'only global moderators or admins should be able to soft-delete users' do
```
cellio, line 97: Admins, global mods, and the user can soft-delete across the network (i.e. the whole account).  Local mods, using a different button in the UI, can soft-delete the community user only, not affecting the rest of the network.  Do/should we have tests that cover both of these variants?  I note that this is `users_controller`, so maybe we already have the corresponding tests for community users (a separate class)?

test/models/user_test.rb
```diff
@@ -315,4 +314,28 @@ class UserTest < ActiveSupport::TestCase
...
     assert_equal 1, local_result.size
   end
+
+  test 'anonymize should correctly clean up user data' do
+    std = users(:standard_user)
+
+    anonymized_name = "user#{std.id}"
+    anonymized_email = "#{std.id}@deleted.localhost"
+
+    [true, false].each do |dirty|
```
cellio: INFO: I don't think I've ever seen a loop on `[true, false]` before.  What/why?  (Feel free to explain in chat and resolve this; this just seemed like the easiest way to ask the question with context.)

cellio (approved):
LGTM (someone else should also review).  Testing notes/status are in my earlier comment.  I read through the changes in the test cases and they look like they're testing the right things to me (did not run).  I think my remaining earlier comments are either N/A or deferrable -- nothing blocking there.  (Do note that I can't currently test donations.)

app/views/donation_mailer/donation_successful.text.erb
trichoplax: Since it's out of scope for this pull request, I've raised #1892 to make sure it doesn't get forgotten.

test/controllers/users_controller_test.rb
Oaphi: Done, @cellio, expanded the soft deletion for the `profile` type a little - it now checks every user that's `at_least_moderator?`, so should cover both local and global mods and admins.

cellio (approved):
Confirmed this fixes https://meta.codidact.com/posts/294917.  (Someone other than me should still review the code, particularly JS.)

Oaphi:
Sorting preservation done, @cellio, included in this PR as it was a very minor change

cellio (approved):
Confirmed fixes https://github.com/codidact/qpixel/issues/1900.

cellio (approved):
Tested the responsive changes (filters, comments) -- LGTM.  ~~I'm not seeing the styling on the "apply" button on filters (in Chrome), but I also do not want to block on that since this is not a regression and this PR is already large and long-running.  Let's get this reviewed and merged, and we can look into that separately.~~ - we talked; I misunderstood the change; all is fine.

app/models/user.rb
```diff
@@ -476,12 +482,31 @@ def active_flags(post)
     post.flags.where(user: self, status: nil)
   end
 
-  def do_soft_delete(attribute_to)
+  # Anonymizes the user (f.e., for the purpose of soft deletion)
+  # @param dirty [Boolean] if set to +false+, will persist the changes
+  def anonymize(dirty: false)
```
ArtOfCode-: suggestion: rename this parameter - my instinct for a `dirty` param is that it indicates a "dirty" operation, i.e. one with side-effects - meaning `true` would persist changes. Sounds like there are multiple interpretations of this - let's rename it something more descriptive. `persist_changes`, perhaps?

app/views/application/keyboard_tools.html.erb
ArtOfCode-: Looking at it, I think this is a remnant of keyboard tools before we moved it to user preferences. I think this can go.

ArtOfCode- (approved):
Couple of suggestions but in general LGTM

app/models/user.rb
Oaphi: Yeah, I wasn't sure how to call it either - opted to use `dirty` to reflect that it's related to [ActiveModel::Dirty](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html), but not married to the choice - will change, `persist_changes` sounds descriptive and unambiguous to me, let's go with it (done).

ArtOfCode- (approved):

PR #1904, opened 2025-11-14
query_url helper should correctly rebuild merged query

### LINKED ISSUE #1902: NoMethodError related to tags from category page
I think this is related to filters, based on the URL.  From error logs in prod (one example of a pattern):

```
at https://meta.codidact.com/categories/3?predefined_filter=&min_score=&max_score=&min_answers=&max_answers=&status=any&include_tags=%5B%22394%22,%20%22377%22%5D&sort=score

    NoMethodError: undefined method `all?' for "[\"394\", \"377\"]":String

    if params[:include_tags]&.all? { |id| id.match? valid_value[:integer] }
                            ^^^^^^

/var/apps/qpixel/app/helpers/search_helper.rb:117:in `params_to_qualifiers'
/var/apps/qpixel/app/controllers/categories_controller.rb:173:in `set_list_posts'
/var/apps/qpixel/app/controllers/categories_controller.rb:19:in `show'
```

Oaphi:
closes #1902 

The issue in question can be reproduced by (and should not be reproducible with the PR applied):
1. Creating a saved filter with a couple of included tags (doesn't matter which, but it's best to add at least 2).
2. Applying the filter from step 1 in a category view (doesn't matter which one).
3. Attempting to switch to a different sorting type / order - at this point, there should be a server error.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -48,7 +48,7 @@ module ApplicationHelper
   end
 
   ##
-  # Utility to add additional query parameters to a URI.
+  # Adds additional query parameters to a URI.
   # @param base_url [String, nil] A base URI to which to add parameters. If none is specified then the request URI for
   #   the current page will be used.
   # @param params [Hash{#to_s => #to_s}] A hash of query parameters to add to the URI.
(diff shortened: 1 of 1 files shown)


cellio (approved):
Tested and confirmed the error (on `develop`) and the fix (on this branch).

ArtOfCode- (approved):

PR #1907, opened 2025-11-17
show pending edits on deleted posts (and the indicator) IFF current user can handle them

### LINKED ISSUE #1496: Suggested edits on deleted posts are not visible in the list of pending edits
meta:292505
https://software.codidact.com/posts/294790

If a post is deleted but has a pending suggested edit, which can happen rarely, that suggested edit doesn't show up in the list of pending edits, only on the deleted post itself.  For users who can see deleted posts, let's include such entries on the list of pending suggestions, with the pink backgrounds to convey that these posts are currently deleted.

Why wouldn't we just throw out pending edits if a post is deleted?  Because maybe user A is trying to help user B salvage a post and we don't want to lose that.

To reproduce:

- User 1 (no abilities beyond Participate Everywhere) creates a post.
- User 2 (no abilities ditto) suggests an edit.
- User 1 deletes the post with the edit still pending.
- User 3 (moderator) views the pending edits, and that one is not there.
- User 1 can't see the list of pending edits but can view the deleted post, which shows the pending edit.  (And -- just tested -- can approve the edit, leaving the post deleted as expected.)

cellio:
Fixes https://github.com/codidact/qpixel/issues/1496.

Test 1: two pending edits, one on a deleted post.

Editor (who can't see deleted posts):

one pending edit

Editor-curator (can see deleted posts):

two pending edits, one with deleted styling

----

Test 2: one pending edit, on a deleted post.

Editor sees nothing, and note that the indicator dot isn't there:

no pending edits

Editor-curator sees the edit and the indicator dot:

one pending edit

----

No changes have been made to the suggested-edit page itself.  Post authors could always see those, so they were already visible without any special styling, but maybe now that we're broadening who can see them we should also do something there to indicate that the *edit* is still there but the underlying *post* is currently deleted.

### DIFF
--- a/app/controllers/suggested_edit_controller.rb
+++ b/app/controllers/suggested_edit_controller.rb
@@ -14,13 +14,9 @@ class SuggestedEditController < ApplicationController
 
     @category = params[:category].present? ? Category.find(params[:category]) : nil
 
-    @edits = SuggestedEdit.where(post: Post.in(@category),
-                                 active: !@show_decided)
+    @edits = SuggestedEdit.accessible_to(current_user, @category)
+                          .where(active: !@show_decided)
                           .order(@sort_type => @sort_order)
-
-    @edits_undeleted = SuggestedEdit.where(post: Post.undeleted.in(@category),
-                                           active: !@show_decided)
-                                    .order(@sort_type => @sort_order)
   end
 
   def show
--- a/app/controllers/suggested_edit_controller.rb
+++ b/app/controllers/suggested_edit_controller.rb
@@ -19,9 +19,9 @@ class SuggestedEditController < ApplicationController
                           .order(@sort_type => @sort_order)
 
     @edits_undeleted = SuggestedEdit.where(post: Post.undeleted.in(@category),
-                                 active: !@show_decided)
-                          .order(@sort_type => @sort_order)
-end
+                                           active: !@show_decided)
+                                    .order(@sort_type => @sort_order)
+  end
 
   def show
     render layout: 'without_sidebar'
--- a/app/controllers/suggested_edit_controller.rb
+++ b/app/controllers/suggested_edit_controller.rb
@@ -14,10 +14,14 @@ class SuggestedEditController < ApplicationController
 
     @category = params[:category].present? ? Category.find(params[:category]) : nil
 
-    @edits = SuggestedEdit.where(post: Post.undeleted.in(@category),
+    @edits = SuggestedEdit.where(post: Post.in(@category),
                                  active: !@show_decided)
                           .order(@sort_type => @sort_order)
-  end
+
+    @edits_undeleted = SuggestedEdit.where(post: Post.undeleted.in(@category),
+                                 active: !@show_decided)
+                          .order(@sort_type => @sort_order)
+end
 
   def show
     render layout: 'without_sidebar'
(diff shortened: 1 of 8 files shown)


app/controllers/suggested_edit_controller.rb
```diff
@@ -14,9 +14,13 @@ def category_index
 
     @category = params[:category].present? ? Category.find(params[:category]) : nil
 
-    @edits = SuggestedEdit.where(post: Post.undeleted.in(@category),
+    @edits = SuggestedEdit.where(post: Post.in(@category),
                                  active: !@show_decided)
                           .order(@sort_type => @sort_order)
+
+    @edits_undeleted = SuggestedEdit.where(post: Post.undeleted.in(@category),
```
cellio: It would be nice to fix the near-duplication here but I don't know if it's practical.

app/views/suggested_edit/category_index.html.erb
```diff
@@ -35,7 +35,11 @@
     </tr>
     <% categories.each do |cat| %>
       <% next if (cat.min_view_trust_level || -1) > (current_user&.trust_level || 0) %>
-      <% sug_edits = SuggestedEdit.where(post: Post.undeleted.in(cat), active: !@show_decided).count %>
```
cellio, line 38: I don't know if this edit-counting code should really be in the view, but that's where I found it so I expanded it here at least for now.

app/controllers/suggested_edit_controller.rb
Oaphi: Fixed (apologies for a direct push, wanted to avoid the overhead of adding a new PR), this duplication is unnecessary - we now have `accessible_to` methods for narrowing down results based on access.

app/views/suggested_edit/category_index.html.erb
Oaphi: Yeah, it should not, but let's tackle it separately - for now, I only simplified the check a little

app/controllers/suggested_edit_controller.rb
cellio: I never object to direct collaboration on something I've marked as ready!  There was no risk of colliding with anything I had in progress; no need to do extra PRs.  Thanks for the help!  (I forgot we had those methods -- much better.)

app/models/suggested_edit.rb
```diff
@@ -22,6 +22,15 @@ class SuggestedEdit < ApplicationRecord
   scope :by, ->(user) { where(user: user) }
   scope :rejected, -> { where(active: false, accepted: false) }
 
+  # Gets suggested edits scoped for a given user and category
+  # @param user [User] user to check
+  # @param category [Category] category to check
+  # @return [ActiveRecord::Relation<SuggestedEdit>]
+  def self.accessible_to(user, category)
```
cellio, line 29: I forgot about this too, but categories can have restricted access -- is this covered already or do we need to add a check here?

cellio (commented):
Thanks for the simplifications!  Much better.

I don't know why this is happening now (seems unrelated to your changes), but after handling all pending edits, editors still see the silver dot.  I don't *think* it did that before but maybe it did; the focus here was on the false positives from deleted posts and that's fixed now, but there's still something weird going on.

cellio (commented):
LGTM now -- thanks for catching and fixing the cache issue.

I tested suggested edits in restricted categories to make sure that's still fine.  Editors can see the suggestions but can't approve them, which is a change we added a while back and I wanted to make sure was still there.  It is; we're good.

ArtOfCode- (approved):

PR #1909, opened 2025-11-25
Make ability score percent calculation resilient against negative values

Oaphi:
This PR makes our ability score percent calculation more resilient against incorrect values, specifically against negative thresholds. The new helper method is intended to be expanded in the future to cover other cases (such as setting score thresholds to a value higher than 1.0).

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -68,7 +68,7 @@ class Ability < ApplicationRecord
 
   private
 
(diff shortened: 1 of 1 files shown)


cellio (approved):
LGTM.  As noted, we're not fixing the frontend here, just the backend, so you can set bad values and we'll handle them without causing server errors.

PR #1917, opened 2025-11-30
Assorted fixes for the Code Golf's leaderboard

### LINKED ISSUE #1911: Code Golf leaderboard excludes answers from deleted users
https://codegolf.codidact.com/posts/294481

The leaderboard for Code Golf challenges shows/links answers along with byte count, language, and user.  If a visible answer is owned by a deleted user, that answer is not included in the leaderboard.  The board should show all non-deleted answers and either omit the user name or say "deleted user" for affected answers.

Oaphi:
This PR fixes several issues with the dashboard:

1. Answers from deleted users now show up as expected:
     image
2. The leaderboard no longer shows `undefined` instead of username / language if either can't be determined (I chose "N/A" but not married to it - deferring to @trichoplax's judgment call):
    image
3. The leaderboard now always renders in the correct place (before the number of answers header) - see the issue for details

closes #1911 

As a side-effect of these changes, we now have two more hooks that userscripts / community assets can rely on: `js-deleted-post` and `js-answers-header` (neither of these is likely to go away, so let's make sure layout changes don't screw up with user-built stuff).

### DIFF
--- a/app/views/posts/show.html.erb
+++ b/app/views/posts/show.html.erb
@@ -23,7 +23,7 @@
 
 <% if @post.post_type.has_answers %>
   <% if num_answers > 0 %>
-    <h2><%= pluralize(num_answers, 'answer') %></h2>
+    <h2 class="js-answers-header"><%= pluralize(num_answers, 'answer') %></h2>
   <% end %>
   <% if num_answers > 1 %>
     <div class="button-list is-gutterless has-float-right">
--- a/app/views/posts/show.html.erb
+++ b/app/views/posts/show.html.erb
@@ -23,7 +23,7 @@
 
 <% if @post.post_type.has_answers %>
   <% if num_answers > 0 %>
-    <h2 class="js-answers-header"><%= pluralize(num_answers, 'answer') %></h2>
+    <h2 id="answers" class="js-answers-header"><%= pluralize(num_answers, 'answer') %></h2>
   <% end %>
   <% if num_answers > 1 %>
     <div class="button-list is-gutterless has-float-right">
--- a/public/assets/community/codegolf.js
+++ b/public/assets/community/codegolf.js
@@ -330,9 +330,7 @@
       // If x were undefined, it would be automatically sorted to the end, but not so if x.score is undefined, so this needs to be stated explicitly.
       sort = (x, y) => typeof x.score === "undefined" ? 1 : x.score - y.score;
 
-      document
-        .querySelector(".post:first-child")
-        .nextElementSibling.insertAdjacentElement("afterend", embed);
+      document.querySelector(".js-answers-header")?.insertAdjacentElement('beforebegin', embed);
 
       refreshBoard(sort);
     } else if (
@@ -342,9 +340,7 @@
       // If x were undefined, it would be automatically sorted to the end, but not so if x.score is undefined, so this needs to be stated explicitly.
       sort = (x, y) => typeof x.score === "undefined" ? 1 : y.score - x.score;
 
-      document
-        .querySelector(".post:first-child")
-        .nextElementSibling.insertAdjacentElement("afterend", embed);
+      document.querySelector(".js-answers-header")?.insertAdjacentElement("beforebegin", embed);
 
       refreshBoard(sort);
     }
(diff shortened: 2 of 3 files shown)


public/assets/community/codegolf.js
```diff
@@ -331,9 +330,7 @@
       // If x were undefined, it would be automatically sorted to the end, but not so if x.score is undefined, so this needs to be stated explicitly.
       sort = (x, y) => typeof x.score === "undefined" ? 1 : x.score - y.score;
 
-      document
-        .querySelector(".post:first-child")
-        .nextElementSibling.insertAdjacentElement("afterend", embed);
+      document.querySelector(".js-answers-header")?.insertAdjacentElement('beforebegin', embed);
 
       refreshBoard(sort);
```
trichoplax, line 335: Optional change while we're here (feel free to skip):

These lines appear to be identical in both the `if` and `else if`. I wonder if it's worth changing the `else if` to an `else` so that all cases are covered (making the descending sort the default when there is no "code-golf" or "lowest-score" tag). Would this allow these lines to be moved after the `if`/`else` block to remove the duplication?

app/views/posts/show.html.erb
```diff
@@ -23,7 +23,7 @@
 
 <% if @post.post_type.has_answers %>
   <% if num_answers > 0 %>
-    <h2><%= pluralize(num_answers, 'answer') %></h2>
+    <h2 class="js-answers-header"><%= pluralize(num_answers, 'answer') %></h2>
```
trichoplax: Since this is easier to consider changing now rather than later when other things might have come to depend on it: Is there any benefit to making this an id rather than a class?

I don't know what the various arguments may be in each direction, but I'm asking because of one benefit I can think of: allowing a user to link to the position on the page where the answers start without linking to a specific answer (which would override that answer's position in the sort order).

I can't imagine this happening often, but I did once find use for it on SE, where I linked from the top of a King of the Hill contest with a particularly long specification so that people already familiar with the spec could skip straight to viewing the answers.

trichoplax (changes requested):
Functionality all working well. I've added a couple of review comments but please treat them as optional.

There is one problem with the layout. Now that the leaderboard is further up the page the checkboxes "Group by language" and "Show placements" show vertically, below the "Leaderboards by language" button. Previously they showed horizontally, both on the same line as the button.

public/assets/community/codegolf.js
trichoplax: Perhaps it's better to keep the `if`/`else if` structure rather than covering all cases, so the leaderboard is only inserted if one of the 4 specified tags is present. Maybe some challenge types that include none of the 4 tags would not be suitable for the current leaderboard. Perhaps popularity-contest (even though that has no challenges yet).

Oaphi:
> There is one problem with the layout. Now that the leaderboard is further up the page the checkboxes "Group by language" and "Show placements" show vertically, below the "Leaderboards by language" button. Previously they showed horizontally, both on the same line as the button.

@trichoplax they show up just fine for me (unless I misunderstood what you meant):

image

I presume this happens due to wrapping on narrower screens - IIRC, that's not related to the change:

image

Happy to fix that while we are at it, though, of course

app/views/posts/show.html.erb
Oaphi: We can do both, I suppose - I am not in favor of removing the class as this is our convention for persistent JavaScript hooks across the project (`.js-*` classes), but that doesn't mean we can't have the best of both worlds.

trichoplax: That sounds like a good reason to keep the class. In that case, adding an id is something that can be done any time, so no reason to include it in this pull request (I just wanted to avoid excluding it for future, which doesn't apply if we can just have both - I hadn't thought of this).

Oaphi: I went ahead and added an `answers` id to the header - if that sounds good to you, we can leave it at that for now (we'd want to add a visible and clickable anchor at some point to make the id more useful)

trichoplax: Yes that's all I was thinking of - making it discoverable is unrelated to this pull request. Thank you!

trichoplax (approved):
The layout problem I mentioned in my original review turned out to be my error (I forgot to include the community CSS file in the head file). All working perfectly now.

Thanks very much for helping me spot this in Discord.

PR #1920, opened 2025-12-03
Split ThreadFollower into separate models for posts and threads

trichoplax:
## Potentially irreversible change
**This code change must not be merged without running the migration, because the code refers to a table that does not yet exist. The movement of data makes this change irreversible (at least for Rails' automatic rollback), and the fact that we use MySQL means that if the migration fails part way through it will be left in an intermediate state (for example, with a table created but no data moved). Separate `up` and `down` methods have been added to the migration, designed to handle rerunning or undoing the migration after an only partial success. However, please ensure backups are up to date before running this migration.**

---

The [ThreadFollower model was created to keep track of users following comment threads](https://github.com/codidact/qpixel/blob/develop/db/migrate/20210102162706_create_thread_followers.rb). Later it had a [post_id field added to keep track of users following new threads on posts](https://github.com/codidact/qpixel/blob/develop/db/migrate/20220913183826_add_post_to_thread_followers.rb). This pull request splits out this later addition into a separate model NewThreadFollower.

Following [discussion in Discord](https://discord.com/channels/634104110131445811/668890340199235613/1445896279900291275) the migration file now covers all of:
- Create the `new_thread_followers` table .
- Move all rows containing a `post_id` from `thread_followers` to the new `new_thread_followers` table.
- Remove the now unused `post_id` column from the `thread_followers` table.

I'm making this change so that #1548 can add a large number of rows (one per post) to the `new_thread_followers` table, rather than adding to the current combined table `thread_followers`. Waiting until after that large number of rows is added before splitting the table would require a much larger data movement between tables later (unless it is decided to never split the table, which I'm uncomfortable with since there's no reason for it to serve two independent purposes).

I'm not experienced in migrations so if there are obvious things that don't need to be mentioned in review, please mention them anyway.


db/migrate/20251203164131_create_new_thread_followers.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      t.timestamps
+    end
+    add_index :new_thread_followers, [:user_id, :post_id]
+  end
+
+  def move_rows_with_non_nil_post_id
+    NewThreadFollower.insert_all(
+      ThreadFollower.select(:user_id, :post_id, :created_at, :updated_at).where.not(post_id:nil)
+    )
```
Oaphi, line 35: Unfortunately, `insert_all` will fail upon being passed a non-empty relation of `ThreadFollower`s as it expects an array of hashes. It *is* able to handle relation conversion but not the models. I suggest just simply making sure the method call gets the expected array (the solution is tested):
```ruby
ThreadFollower.select(:user_id, :post_id, :created_at, :updated_at)
              .where.not(post_id:nil)
              .to_a
              .map(&:attributes)
```

Oaphi (dismissed):

cellio (commented):
There's an issue with the migration that @Oaphi will describe.  Meanwhile, I've tested these changes (after locally making the migration fix) and behavior looks good -- the correct tables get updated on follow/unfollow on both posts and threads, and notifications get delivered as expected.

db/migrate/20251203164131_create_new_thread_followers.rb
```diff
@@ -0,0 +1,28 @@
+class CreateNewThreadFollowers < ActiveRecord::Migration[7.2]
+  def change
+    create_table_new_thread_followers
+    move_rows_with_non_nil_post_id
+    remove_post_id_column_from_thread_followers
+  end
+
+  def create_table_new_thread_followers
+    create_table :new_thread_followers do |t|
```
Oaphi: Non-blocking: since we can't make sure both the schema and data changes are in a single transaction, let's harden the migration a tiny bit and add `if_not_exists` to `create_table` and `add_index` (`create_table :new_thread_followers, if_not_exists: true` and `add_index :new_thread_followers, [:user_id, :post_id], if_not_exists: true` respectively) - this will ensure we can simply attempt to migrate again in case of a hard failure of subsequent data changes.

db/migrate/20251203164131_create_new_thread_followers.rb
trichoplax: Thank you for explaining this. I've made this change now.

I really should have known to use more varied data in my testing, ensuring that there were rows with a post_id, especially since the whole point of making this pull request is to support #1548 which will ensure such rows exist for all posts.

db/migrate/20251203164131_create_new_thread_followers.rb
trichoplax: In my testing I was simply resetting the database for each new attempt - checking for existence is much more convenient! I've now made this suggested change, and also added a check for the existence of the `post_id` column in the existing `thread_followers` table before attempting the data migration and column removal. It seems happy to be run a second time now, although it still isn't reversible. Is this sufficient?

cellio:
I tested `down` and `up` after the latest changes and it seemed to work, but also, my dev environment got a little messy from the earlier changes, so somebody else should provide final approval.

Oaphi (approved):
LGTM to me after the latest changes (tested, manual tests included), however, I am deferring to someone else to provide a final review given that I've made a couple of commits to the branch.

cellio:
LGTM but want @ArtOfCode- to review.  Also, hold merge until after we deploy; we have some other small things ready to ship, so let's get those out and *then* merge this just in case there are any issues that would require rollback.  (I don't think there are, but this isn't user-facing so it doesn't need to be deployed right away.)

db/migrate/20251226185531_create_new_thread_followers.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def create_table_new_thread_followers
+    create_table :new_thread_followers, if_not_exists: true do |t|
+      t.bigint :user_id
+      t.bigint :post_id
+
+      t.timestamps
+    end
+    add_index :new_thread_followers, [:user_id, :post_id], if_not_exists: true
```
ArtOfCode-, line 25: I would add indexes here for `user_id` and `post_id` separately as well. We're likely to be querying for thread followers by post or by user as well as by both.

db/migrate/20251226185531_create_new_thread_followers.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def move_rows_with_non_nil_post_id
+    NewThreadFollower.insert_all(
+      ThreadFollower.select(:user_id, :post_id, :created_at, :updated_at)
+        .where.not(post_id:nil)
+        .to_a
+        .map(&:attributes)
+    )
+    ThreadFollower.where.not(post_id:nil).delete_all
```
ArtOfCode-: Let's wrap this in a transaction so we don't only complete half of this operation and end up losing data one way or another.

db/migrate/20251226185531_create_new_thread_followers.rb
```diff
@@ -0,0 +1,61 @@
...
+    add_reference :thread_followers, :post, index: true, foreign_key: true, if_not_exists: true
+  end
+
+  def move_rows_back_from_new_thread_followers
+    ThreadFollower.insert_all(
+      NewThreadFollower.select(:user_id, :post_id, :created_at, :updated_at)
+        .to_a
+        .map(&:attributes)
+    )
+    NewThreadFollower.delete_all
```
ArtOfCode-, line 53: Likewise, let's transactionify this

ArtOfCode- (changes requested):

db/migrate/20251226185531_create_new_thread_followers.rb
ArtOfCode-: Revision after discussion in Discord: let's add an additional index for `:post_id` - the composite index will work for searches on `:user_id`.

db/migrate/20251226185531_create_new_thread_followers.rb
ArtOfCode-: Disregard - as this is a data change, the automatic migration transaction should take care of it.

db/migrate/20251226185531_create_new_thread_followers.rb
trichoplax: Thanks for explaining that the composite index works differently in MySQL. I've now added the index on `post_id` (and the corresponding removal for rollback).

ArtOfCode- (approved):

trichoplax:
# ~~I'm avoiding merging~~
~~I'm waiting until potential problems in the `develop` branch are resolved / ruled out:~~

https://discord.com/channels/634104110131445811/668890340199235613/1452154459223228518

---

This has now been resolved by #1931 so this pull request just needs rereview following adjustments made since then.

cellio (commented):
I've done a bunch of manual testing and as far as I can tell, this is working properly.  (Still want a code review from someone else.)  I tested with a mix of users (admin, post owners, regular users), testing following and unfollowing of both threads and new threads.  After each follow change I checked the table counts in the DB and they reflected the change.  The users who should have gotten notified did and the ones who shouldn't didn't, except that post owners always get notified even if they unfollowed new threads, which I believe is expected behavior that will be address with the linked issue.

My testing, being manual, leaves room for error, hence the request for code review specifically to focus on the test cases here.  Things look right; I don't think this will break anything; I'm not confident in my own abilities enough to click the "approve" button without another reviewer.

app/controllers/comments_controller.rb
```diff
@@ -60,11 +60,11 @@ def create_thread
         @comment.post.user.create_notification(notification, helpers.comment_link(@comment))
       end
 
-      ThreadFollower.where(post: @post).each do |tf|
-        unless tf.user == current_user || tf.user == @comment.post.user
-          tf.user.create_notification(notification, helpers.comment_link(@comment))
+      NewThreadFollower.where(post: @post).each do |ntf|
+        unless ntf.user == current_user || ntf.user == @comment.post.user
```
Oaphi: Non-blocking: let's make the check more robust with our identity helper:
```ruby
ntf.user.same_as?(current_user) || ntf.user.same_as?(@comment.post.user)
```

Oaphi (approved):
Current state of the PR LGTM code-wise

ArtOfCode- (approved):

app/controllers/comments_controller.rb
trichoplax: Good idea. I've now applied that to some other places in the same file too.

app/controllers/comments_controller.rb
```diff
@@ -395,7 +395,7 @@ def check_thread_access
   end
 
   def check_privilege
-    unless current_user&.at_least_moderator? || current_user == @comment.user
+    unless current_user&.at_least_moderator? || current_user.same_as?(@comment.user)
```
Oaphi: I believe we need safe navigation here, @trichoplax - `current_user` can be `nil` in the RHS branch (seems to be an edge case but I am a bit paranoid about `check_privilege`)

Oaphi: whoops, forgot to take a look at the next commit, my apologies

trichoplax: You're right, I overlooked this. Thankfully rubocop had the same concern you did.

Oaphi:
@trichoplax I've applied a couple of minor Rubocop autofixes as we lint migrations too now - the PR is good to go as soon as you are ready

PR #1922, opened 2025-12-08
keyboard next and prev were reversed

### LINKED ISSUE #1921: keyboard 'next' and 'previous' are flipped
meta:295125

According to the help for keyboard navigation, 'j' is down (next) and 'k' is up (previous).  The implementation is reversed.

cellio:
Fixes https://github.com/codidact/qpixel/issues/1921

I find it a little puzzling that in the JS code that implements the keyboard options, 'k' came before 'j' and the values were flipped.  Perhaps a previous edit flipped them from what was written originally?  Git blame did not bring me enlightenment.


Oaphi (approved):
All good - it's been like this from the start, just a minor oversight

PR #1923, opened 2025-12-11
Fix for QPixel#user race conditions

Oaphi:
Our current approach to fetching the user has an additional flaw that I overlooked in #1120 a couple years ago.
We have code that can (specifically, keyboard shortcuts on category index pages) resolve _before_ `QPixel._user` is cached (because `_pendingUserResponse` only stores the `fetch` promise). This PR ensures that is no longer possible


ArtOfCode- (approved):

cellio (approved):
Works for me now.

PR #1928, opened 2025-12-13
Account for text fragments when autoexpanding post history items

### LINKED ISSUE #1925: `/history`'s `history-event`s do not `open` when their relevant `id=` has a fragment directive appended.
meta:295150

#### What The Problem Is

[`linux.codidact.com/posts/294865/history#3`](https://linux.codidact.com/posts/294865/history#3) renders the relevant revision, expanded. by contrast, when a fragment directive is appended, like [`linux.codidact.com/posts/294865/history#3:~:`](https://linux.codidact.com/posts/294865/history#3:~:), it does not:

<blockquote>

~~~ASM
17:06:22.253 Uncaught Error: Syntax error, unrecognized expression: #3:~:text=it.%20They%20must%2C-,instead%2C%20merely%20wait,-for%20it%20to
    jQuery 7
        error
        tokenize
        select
        fa
        find
        init
        n
    t https://linux.codidact.com/assets/application-f633374372c26fba22ed96cbc21568af874c89721872e975217d544f37266dbc.js:74
    EventListener.handleEvent* https://linux.codidact.com/assets/application-f633374372c26fba22ed96cbc21568af874c89721872e975217d544f37266dbc.js:74
    jQuery 8
        i
        fireWith
        ready
        J
        promise
        <anonymous>
        <anonymous>
        <anonymous>
jquery.min.js:2:12556
    jQuery 7
        error
        tokenize
        select
        fa
        find
        init
        n
    t https://linux.codidact.com/assets/application-f633374372c26fba22ed96cbc21568af874c89721872e975217d544f37266dbc.js:74
    (Async: EventListener.handleEvent)
    <anonymous> https://linux.codidact.com/assets/application-f633374372c26fba22ed96cbc21568af874c89721872e975217d544f37266dbc.js:74
    jQuery 8
        i
        fireWith
        ready
        J
    (Async: EventListener.handleEvent)
        promise
        <anonymous>
        <anonymous>
        <anonymous>
~~~

</blockquote>

This is problematic when the user wants to link to a text fragment.

#### How To Reproduce It

1. ~~~JavaScript
   #!/usr/bin/env xdg-open
   window.location.href = 'https://linux.codidact.com/posts/294865/history#3:~:text=it.%20They%20must%2C-,instead%2C%20merely%20wait,-for%20it%20to';
   ~~~

1. Select the `//*[@id="toolbox-tab-webconsole"]`.

1. Expand the relevant `div.message:nth-child`.

#### Expected Behaviour

The relevant `details.history-event` should `open`.

#### My Environment

-	#### QPixel

	> Version [`1ff2d8b7`](https://github.com/codidact/qpixel/commit/1ff2d8b7f27a961d91ad62415e3a1d43e3e7efc6) (2025-11-19 11:32:27Z)

-	#### My Browser

	1.	~~~sh
		#!/usr/bin/env sh
		rpm -qi $(rpm -qf $(command -v firefox))
		~~~

	1.	> ~~~YAML
		> Name        : firefox
		> Version     : 145.0.1
		> Release     : 1.fc43
		> Architecture: x86_64
		> Install Date: Sun 23 Nov 2025 03:06:22 GMT
		> Size        : 267898587
		> Signature   :
		>               RSA/SHA256, Thu 20 Nov 2025 08:37:07 GMT, Key ID 829b606631645531
		> Source RPM  : firefox-145.0.1-1.fc43.src.rpm
		> Build Date  : Wed 19 Nov 2025 09:22:06 GMT
		> Build Host  : buildhw-x86-08.rdu3.fedoraproject.org
		> Packager    : Fedora Project
		> Vendor      : Fedora Project
		> Bug URL     : https://bugz.fedoraproject.org/firefox
		> ~~~

Oaphi:
closes #1925

### DIFF
--- a/app/assets/javascripts/post_histories.js
+++ b/app/assets/javascripts/post_histories.js
@@ -1,19 +1,12 @@
 document.addEventListener('DOMContentLoaded', () => {
   const openRelevantEditOnly = () => {
+    const [[, historyId]] = location.hash.split(':~:');
+
     document.querySelectorAll('details.history-event').forEach((el) => {
       if (el instanceof HTMLDetailsElement) {
-        el.open = false;
+        el.open = el.id === historyId;
       }
     });
-
-    const [hash] = location.hash.split(':~:');
-    const historyId = hash.slice(1);
-
-    const historyItem = document.getElementById(historyId);
-
-    if (historyItem instanceof HTMLDetailsElement) {
-      historyItem.open = true;
-    }
   };
 
   window.addEventListener('hashchange', openRelevantEditOnly);
--- a/app/assets/javascripts/post_histories.js
+++ b/app/assets/javascripts/post_histories.js
@@ -1,9 +1,21 @@
-$(() => {
+document.addEventListener('DOMContentLoaded', () => {
   const openRelevantEditOnly = () => {
-    $("details.history-event").prop('open', false);
-    $(location.hash).prop('open', true);
-  }
+    document.querySelectorAll('details.history-event').forEach((el) => {
+      if (el instanceof HTMLDetailsElement) {
+        el.open = false;
+      }
+    });
 
-  window.addEventListener("hashchange", openRelevantEditOnly);
+    const [hash] = location.hash.split(':~:');
+    const historyId = hash.slice(1);
+
+    const historyItem = document.getElementById(historyId);
+
+    if (historyItem instanceof HTMLDetailsElement) {
+      historyItem.open = true;
+    }
+  };
+
+  window.addEventListener('hashchange', openRelevantEditOnly);
   openRelevantEditOnly();
 });


app/assets/javascripts/post_histories.js
```diff
@@ -1,9 +1,21 @@
...
+document.addEventListener('DOMContentLoaded', () => {
   const openRelevantEditOnly = () => {
-    $("details.history-event").prop('open', false);
-    $(location.hash).prop('open', true);
-  }
+    document.querySelectorAll('details.history-event').forEach((el) => {
+      if (el instanceof HTMLDetailsElement) {
+        el.open = false;
+      }
+    });
```
trichoplax, line 9: This works fine, and is very easy to understand. I'd personally still find it readable if combined with the other `if` block so that only the relevant details sections are closed, rather than closing all and then reopening one. This makes no meaningful difference for performance - I'm just thinking it would be good to have less code to read, provided it is still as understandable. I was thinking something along the lines of:

```js
    document.querySelectorAll('details.history-event').forEach((el) => {
      if (el instanceof HTMLDetailsElement) {
        el.open = el === historyItem;
      }
    });
```

(This would replace the `if` block on line 14, so lines 3-7 could be omitted.)

app/assets/javascripts/post_histories.js
```diff
@@ -1,9 +1,21 @@
...
 
-  window.addEventListener("hashchange", openRelevantEditOnly);
+    const [hash] = location.hash.split(':~:');
+    const historyId = hash.slice(1);
```
trichoplax:
```js
const [hash] = 
```

I love this. Come join us on https://codegolf.codidact.com

More seriously, feel free to skip the rest of this optional comment. You taught me something new and I went exploring.

---

I had no idea you could omit the `, ...rest` in `const [hash, ...rest] = ` and have JavaScript just automatically assume you want to throw the rest away. I do love this, but I wonder if its meaning will be as clear as:

```js
const hash = location.hash.split(':~:')[0];
```

If we're confident that future developers will be comfortable with destructuring, we could go further and change the order of operations to condense 2 lines into 1.

Instead of:

```js
    const [hash] = location.hash.split(':~:');
    const historyId = hash.slice(1);
```

We'd just have:

```js
    const [historyId] = location.hash.slice(1).split(':~:');
```

If we decide against the destructuring, we could still have:

```js
    const historyId = location.hash.slice(1).split(':~:')[0];
```

To be clear, I'm perfectly happy as it is already (it's beautiful). These are just options because I was curious what was possible.

trichoplax (approved):
Great to see more jQuery being retired.

The functionality of each previous jQuery line is replicated without the vulnerability to text fragment directives in the URL, so this looks complete to me, and I've approved.

The only comments I've made are a matter of personal preference where I can't guess what will be easier to read for different people, so I've just listed some thoughts. I'm fine with this being merged as is though.

app/assets/javascripts/post_histories.js
Oaphi: Agreed, let's make it a bit terser. That said, I am not a big fan of DOM node identity comparison, so I opted to keep the id check.

app/assets/javascripts/post_histories.js
Oaphi: Let's keep the array destructuring assignment - it's been a part of ECMAScript for a long time, I doubt there any devs worth their salt left who don't know what it does or can't quickly look this up when encountered. FYI, you can even nest it as demonstrated by the update 🙂

app/assets/javascripts/post_histories.js
trichoplax: Looks good. Keeping the id check actually made it terser...

app/assets/javascripts/post_histories.js
trichoplax: I tried to think of a way using nesting, but I didn't realise you could throw away just the first element that way. Will bear this in mind.

cellio (approved):
I tested on my dev server with the URL `http://localhost:3000/posts/46/history#2:~:` and it went to the correct history entry.  I don't know what`:~:` is supposed to do (I got that from the bug report), but it doesn't do something obviously bad now.  If @trichoplax (who understands the problem more than I do) is happy, I'm happy.

Oaphi:
> I don't know what`:~:` is supposed to do

As a quick FYI, `:~:` is a part of URI text fragments (that allow browsers to highlight specific parts of a given resource's text content) that delimits their start - it's called [fragment directive](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments#sect). The syntax can be a bit hard to parse, but on the bright side, it's unlikely that such a sequence can be encountered as part of the actual hash (a.k.a. "fragment").

ArtOfCode- (approved):

cellio:
Confirmed on dev server: https://a.dev.codidact.org/posts/725/history#2:~:

PR #1931, opened 2025-12-22
Follow / unfollow thread actions should bump its last activity date & time + minor general fixes

### LINKED ISSUE #1930: "follow" and "unfollow" actions should invalidate cache for inline thread views
Due to our aggressive caching configuration, clicking on the follow toggle button from a given thread's _inline_ view results in cached responses as following / unfollowing doesn't count towards invalidating cache. This is very confusing for end users as unless they have cache disabled or it otherwise expires, they are lead to believe that the action silently failed whereas it, in fact, succeeded.

### LINKED ISSUE #1932: Multiple thread follower rows can be added
**Describe the bug**
If a page is in a different state to the server (for example, if multiple tabs are open and one changes the state) it is possible to follow a thread multiple times, each time adding a separate row to the `thread_followers` table. Unfollowing then requires pressing "unfollow" that many times, since each time only removes 1 of the rows.

**To Reproduce**
1. Open a post with a thread under it that the user is not following.
2. Open it again in another tab.
3. In the first tab, press "follow".
4. In the second tab, press "follow".
5. There are now 2 rows in the `thread_followers` table for that thread and that user, and "unfollow" will need to be pressed twice before it will change to show "follow" again.

**Expected behavior**
There should only ever be zero or one rows in the database for a given combination of user and thread. Compare with the behaviour for following new threads on a post. See [app/controllers/comments_controller lines 332-350](https://github.com/codidact/qpixel/blob/develop/app/controllers/comments_controller.rb#L332-L350).

### LINKED ISSUE #1933: Inconsistent behaviour between thread unfollowing and post unfollowing
**Describe the bug**
If the page state is inconsistent with the server state, so a user presses "unfollow" or "unfollow new" when not following, the result is different between unfollowing a thread and unfollowing new threads on a post:
- For a thread, the request silently fails, changing nothing on the server (correctly), but leaving the page incorrectly saying "unfollow".
- For a post, the request succeeds but changes nothing on the server (correctly), and changes the page to correctly say "follow".

**To Reproduce**
Steps to reproduce the behavior for a thread:
1. Go to a post with a comment thread under it that is being followed by the user.
2. Open it again in a separate tab.
3. Change the state in one tab by pressing "unfollow".
4. Switch to the other tab and press the same again.
5. See nothing happens from a user perspective.
6. See a 400 error in the browser's developer console.

Steps to reproduce the behavior for a post:
1. Go to a post for which the user is following new threads.
2. Open it again in a separate tab.
3. Change the state in one tab by pressing "unfollow new".
4. Switch to the other tab and press the same again.
5. See the link change from "unfollow new" to "follow new".
6. See a 200 response in the browser's developer console.

**Expected behavior**
The behaviour should be the same for a thread as for a post. Which behaviour should be adopted for both? From a user perspective, having the link update to say the correct thing seems better. For the response status, I don't know whether it's better to return a 400 (as is currently the case for a thread) and make the page update to say "follow" due to the failure, or to return a 200 (as is currently the case for a post) and make the page update to say "follow" since that will be the case whether the user was already following or not.

Should a 400 be reserved for the case where the server is not already in the state being requested, and there is a failure preventing the state from being changed? Should the case where there is no change on the server because it was already in the state the user was requesting result in a 200?

Oaphi:
closes #1930
closes #1932
closes #1933

This PR makes sure following / unfollowing a comment thread bumps the thread's last activity date & time and thus prevents stale content from being served afterwards.

It also fixes confusing inversion of the `User#anonymize` method's `persist_changes` parameter (turns out it wasn't the best idea to rename `dirty` to `persist_changes` while tired - thankfully, the logic was sound, this change is just a matter of making it not confusing).

Additionally, it ensures attempting to make the same action multiple times does not create duplicate entries (as well as automatically cleans up any duplicate entry on unfollow).

Finally, it also makes sure we don't need to reload the whole page upon successfully unfollowing a thread (and gets us closer to removing `thread_unrestrict` entirely).

Note for reviewers: to be able to test this change, you need to [enable caching in development](<https://guides.rubyonrails.org/v7.2/caching_with_rails.html#caching-in-development>) (should be on by default, but it doesn't hurt to check) *and* set the `PERFORM_CACHING` env variable to `true`.

### DIFF
--- a/app/models/comment_thread.rb
+++ b/app/models/comment_thread.rb
@@ -109,7 +109,7 @@ class CommentThread < ApplicationRecord
   # @param user [User] user to remove from followers
   # @return [Boolean] status of the operation
   def remove_follower(user)
-    ThreadFollower.find_by(comment_thread: self, user: user)&.destroy || false
+    ThreadFollower.where(comment_thread: self, user: user).destroy_all.any?
   end
 
   private
--- a/app/models/comment_thread.rb
+++ b/app/models/comment_thread.rb
@@ -99,6 +99,13 @@ class CommentThread < ApplicationRecord
     self.last_activity_at = DateTime.now
   end
 
+  # Removes a given user from the thread's followers
+  # @param user [User] user to remove from followers
+  # @return [Boolean] status of the operation
+  def remove_follower(user)
+    ThreadFollower.find_by(comment_thread: self, user: user)&.destroy || false
+  end
+
   private
 
   # Comment author and post author are automatically followed to the thread. Question author is NOT
--- a/app/models/comment_thread.rb
+++ b/app/models/comment_thread.rb
@@ -53,9 +53,9 @@ class CommentThread < ApplicationRecord
 
   # Gets last activity date and time on the thread
   # @return [DateTime] last activity date and time
-  def last_activity_at
-    last_comment_activity_at = comments.map(&:last_activity_at).compact.max
-    [created_at, locked_at, updated_at, last_comment_activity_at].compact.max
+  def last_activity
+    last_comment_activity = comments.map(&:last_activity).compact.max
+    [created_at, updated_at, last_activity_at, last_comment_activity].compact.max
   end
 
   # Gets a list of user IDs who should be pingable in the thread.
--- a/app/models/comment_thread.rb
+++ b/app/models/comment_thread.rb
@@ -92,7 +92,9 @@ class CommentThread < ApplicationRecord
   # @param user [User] user to register as a follower
   # @return [Boolean] status of the operation
   def add_follower(user)
-    ThreadFollower.create(comment_thread: self, user: user)
+    if ThreadFollower.where(comment_thread: self, user: user).none?
+      ThreadFollower.create(comment_thread: self, user: user)
+    end
   end
 
   # Directly bumps the thread's last activity date & time
--- a/app/models/comment_thread.rb
+++ b/app/models/comment_thread.rb
@@ -92,9 +92,9 @@ class CommentThread < ApplicationRecord
   # @param user [User] user to register as a follower
   # @return [Boolean] status of the operation
   def add_follower(user)
-    if ThreadFollower.where(comment_thread: self, user: user).none?
-      ThreadFollower.create(comment_thread: self, user: user)
-    end
+    return true if ThreadFollower.where(comment_thread: self, user: user).any?
+
+    ThreadFollower.create(comment_thread: self, user: user)
   end
 
   # Directly bumps the thread's last activity date & time
@@ -111,6 +111,8 @@ class CommentThread < ApplicationRecord
   # @param user [User] user to remove from followers
   # @return [Boolean] status of the operation
   def remove_follower(user)
+    return true if ThreadFollower.where(comment_thread: self, user: user).none?
+
     ThreadFollower.where(comment_thread: self, user: user).destroy_all.any?
   end
 
--- a/app/models/comment_thread.rb
+++ b/app/models/comment_thread.rb
@@ -88,11 +88,11 @@ class CommentThread < ApplicationRecord
     end
   end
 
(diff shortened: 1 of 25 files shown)


cellio:
Would it be easy to also, while we're here, make unfollow leave the in-page thread open, same as follow does?  This difference was discussed in chat (pre-existing issue), but I don't know if making them consistent is easy or too much scope creep (and we should file a separate issue for it).

app/models/user.rb
```diff
@@ -483,13 +483,13 @@ def active_flags(post)
   end
 
   # Anonymizes the user (f.e., for the purpose of soft deletion)
-  # @param persist_changes [Boolean] if set to +false+, will persist the changes
+  # @param persist_changes [Boolean] if set to +true+, will persist the changes
```
cellio, line 486: What happened here?  We had a bug with true/false inversion that you noticed while making this change?  I'd like someone else to review this code just as a sanity-check.

cellio (commented):
I tested (with the env var per the description) and it's working as expected.  The code overall looks good to me; I have one question and would like someone else to review that part.

app/models/user.rb
Oaphi: No, thankfully, there is no bug - as I mentioned in the PR's description, seems like I chose a... peculiar way of applying Art's suggestion to rename the `dirty` parameter to `persist_changes`. As a result, I ended up with the comment stating that setting `persist_changes` to `false` persists the changes (which *is* correct, but unexpected). Aligning the parameter's name, default value, and logic with each other simply required a couple of extra changes. Feel free to defer to @ArtOfCode-, they've seen the method before

app/models/comment_thread.rb
```diff
@@ -86,6 +88,30 @@ def maximum_title_length
     end
   end
 
+  # Registers a given user as a follower of the thread
+  # @param user [User] user to register as a follower
+  # @return [Boolean] status of the operation
+  def add_follower(user)
+    ThreadFollower.create(comment_thread: self, user: user)
```
trichoplax, line 100: Is it worth checking for an existing row before adding this one?

trichoplax (approved):
Having switched on caching in development as suggested (`rails dev:cache` from bash inside the container, followed by adding `PERFORM_CACHING=true` to `docker/env`, followed by restarting the container), I can reproduce the problem on the `develop` branch (both the inline thread view and the full thread page can be in a state where clicking "unfollow" appears to have no effect and even a hard refresh doesn't change it to show "follow" (or the other way around)).

On this new branch, the caching problem is fixed so if a page is not matching the server state, refreshing the page fixes it. I see that as fixing the identified problem, so I have approved this pull request, with one minor optional comment.

---

During testing I noticed a couple of pre-existing problems, so to avoid scope creep I've raised them as separate issues #1932 (which I see you already fixed while I was typing up the other one) and #1933.

app/models/comment_thread.rb
Oaphi: Yeah, it is - I am just in the middle of deciding what approach to take here (I like the simplicity of just checking if one already exists, but it has the same drawbacks as find_or_create_by, so not sure yet).

trichoplax: I see no reason for the decision to hold up this pull request unless you particularly want to do it here. The high priority problem is fixed already, and even if there are duplicate rows they won't cause problems anymore with your addition of `destroy_all`.

cellio (approved):
LGTM now.  Thanks for adding that route so the in-page behavior is the same for follow and unfollow.

cellio (approved):
Cleanup job LGTM by inspection. Re-approving pending some testing that @trichoplax is going to do.  (Test code looks good; I haven't set up the bug conditions for manual testing.)

trichoplax (approved):

ArtOfCode- (approved):

PR #1935, opened 2026-01-04
Improve and shorten wordings in right panel of profile

trichoplax:
The changes suggested in meta:287189:
- Remove redundancy in wordings.
- Make some wordings links to relevant pages.
- Change the 3 usages of "Count" to "Total".

Also:
- Change "User since" to "Joined" to avoid the time zone wrapping onto the next line.

Notice that the height of the panel is now significantly reduced.

## Appearance before (left) and after (right)
Screenshot 2026-01-04 at 06-24-47 DEV User user - QPixel

Screenshot 2026-01-04 at 06-25-08 DEV User user - QPixel

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -184,7 +184,7 @@
         <td class="overflow-ellipsis" title="<%= @user.metric 'E' %>"><%= @user.metric 'E' %></td>
       </tr>
       <% if current_user&.id == @user.id || current_user&.at_least_moderator? %>
-        <tr><td colspan="3">User since <%= @user.community_user.created_at %></td></tr>
+        <tr><td colspan="3">Joined <%= @user.community_user.created_at %></td></tr>
       <% end %>
     </table>
 
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -204,13 +204,13 @@
         <% end %>
         <div class="widget--footer">
           <% if current_user&.id == @user.id %>
-           <%= link_to abilities_path, class: 'has-font-weight-bold', 'aria-label': 'View your abilities' do %>
-             Abilities &raquo;
-           <% end %>
+            <%= link_to abilities_path, class: 'has-font-weight-bold', 'aria-label': 'View your abilities' do %>
+              Abilities &raquo;
+            <% end %>
           <% else %>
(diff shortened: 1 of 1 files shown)


app/views/users/show.html.erb
```diff
@@ -146,26 +146,48 @@
...
+            <% if current_user&.id == @user.id %>
+              <%= link_to search_path(search: "user:#{@user.id} post_type:2"), 'aria-label': 'View your answers' do %>
+                Answers
+              <% end %>
+            <% else %>
+              <%= link_to search_path(search: "user:#{@user.id} post_type:2"), 'aria-label': "View answers from #{rtl_safe_username(@user)}" do %>
+                Answers
+              <% end %>
+            <% end %>
+        </td>
```
Oaphi: Since all these widgets only differ in what `aria-label` (and sometimes what link param is used), let's make these less verbose. We also have a convenience `same_as?` method on `User` that is both `nil`-safe and easier to read than `current_user&.id == @user.id` (@user.same_as?(current_user)).

Oaphi (changes requested):

app/views/users/show.html.erb
Oaphi: Something along the lines of (where `is_me` can be shared with all the other widgets):

```erb
<% is_me = @user.same_as?(current_user) %>

<%= link_to search_path(search: "user:#{@user.id} post_type:2"),
                    'aria-label': is_me ? 'View your answers' : "View answers from #{rtl_safe_username(@user)}" do %>
    Answers
<% end %>
```

(somehow GitHub's editor got even worse, so apologies for the single-line `aria-label`)

Oaphi:
LGTM to me otherwise, feel free to override my RC (not critical by any means)

app/views/users/show.html.erb
ArtOfCode-: I'd agree with this - these feel like they're missing some ternaries or helper methods, rather than full `if` statements for a relatively simple wording change.

trichoplax: Thanks for the feedback. I've applied this change so the new code and a few existing places in the same file now use `is_me`. Rather than split a ternary across multiple lines I've put the value for `'aria-label:'` on the next line down from the property name (for the one line that exceeded 120 characters otherwise). Let me know if there's a preferred way to split long lines for ERB files rather than this.

Oaphi (approved):
tested the updated version out - LGTM

PR #1938, opened 2026-01-04
Fix for accidental insertions of "[]()" into fields upon hitting Enter in various post fields + general cleanup

### LINKED ISSUE #1058: Hitting enter in a synonyms field adds a `[]()` to the tag wiki
I tried to add `pie` as a synonym of `Proto-Indo-European` on L&L, but by reflex hit the enter key when I finished typing the name into the field. This, for some reason, caused the wiki field to be filled in with `[]()`.

image

Currently tested using Edge on Windows 11; I haven't tested other browsers yet.

Oaphi:
closes #1058

The PR also:
- enables TypeScript checks for our community assets (and fixes remaining type-related errors in them);
- moves remaining markdown-related helpers to our `QPixel.MD` common API global object;
- fixes runtime error when using the `judasim.js` community asset without the sidebar notice setting filled in;
- fixes Sefaria Linker clobbering post source in edit mode (it links every occurrence of the source text by default);

Since the PR touches upon community assets, DO NOT MERGE until both @trichoplax and @cellio approve (that is, unless you check that those assets work fine with changes yourself).

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -97,7 +97,13 @@ interface QPixelMD {
    * @param start text to insert at selection start
    * @param end text to insert at selection end, if any
    */
-   insertIntoField($field: JQuery<HTMLInputElement | HTMLTextAreaElement>, start: string, end?: string | null): void;
+  insertIntoField($field: JQuery<HTMLInputElement | HTMLTextAreaElement>, start: string, end?: string | null): void;
+  /**
+   * Replace the selected text in an input field with a provided replacement.
+   * @param $field the field in which to replace text
+   * @param text the text with which to replace the selection
+   */
+  replaceSelection($field: JQuery<HTMLInputElement | HTMLTextAreaElement>, text: string): void;
    /**
     * Inserts text at a given {@link idx} in a given {@link str}
     * @param str text to insert into
@@ -482,12 +488,6 @@ interface QPixel {
    */
   preference?: (name: string, community?: boolean) => Promise<string>;
 
-  /**
-   * Replace the selected text in an input field with a provided replacement.
-   * @param $field the field in which to replace text
-   * @param text the text with which to replace the selection
-   */
-  replaceSelection?: ($field: JQuery<HTMLInputElement | HTMLTextAreaElement>, text: string) => void;
   setFilter?: (name: string, filter: QPixelFilter, category: string, isDefault: boolean) => Promise<void>;
 
   /**
--- a/global.d.ts
+++ b/global.d.ts
@@ -91,6 +91,13 @@ interface StripMarkdownOptions {
 }
 
 interface QPixelMD {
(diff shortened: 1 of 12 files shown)


public/assets/community/judaism.js
```diff
@@ -150,7 +151,7 @@ THE SOFTWARE.
 
     /* Event handling for buttons and checkboxes*/
     kb.find('.hbkey').click(function () {
-      t = currentTextfield[0];
+      var t = currentTextfield[0];
```
Oaphi, line 154: Preemptive note: I opted to use `var` for now as it's best we update the whole asset in one go when we can. This change needs to be tested - I am fairly sure I understood the author's intent here, but initializing a variable without any declaration (`var`/`let`/`const`) has a side-effect of declaring a _global_ variable (should not matter here, but I am worried about breaking something I missed that relied on that quirk as it's oddly the only variable initialized in such a manner).

global.d.ts
```diff
@@ -688,6 +702,10 @@ declare var hljs: any;
 declare var MathJax: any;
 // DOMPurify lib, TODO: types
 declare var DOMPurify: any;
+// Sefaria Linker (no known types)
+declare var sefaria: {
+  link: () => void
+} | undefined;
```
Oaphi: AFAIK, [Sefaria Linker](https://www.sefaria.org/linker.js) is just a script loaded directly from the project's website and doesn't even have an NPM package, let alone types - if anyone's aware of the opposite, please do let me know.

ArtOfCode- (approved):

public/assets/community/judaism.js
trichoplax: If @cellio is going to be testing this anyway, would now be a good time to set this to `const` and see if the testing shows up any reason not to?

trichoplax (approved):
What I understand looks good to me, but I don't have an understanding of TypeScript yet so I'll leave it to someone else to approve those changes.

So this is just to confirm that I've tested that the code golf assets still work and that pressing <kbd>Enter</kbd> in the Edit Comment field while editing a post no longer adds an empty Markdown link to the Body field. Pressing <kbd>Enter</kbd> in the Edit Comment field does now save the edit, which I assume is intentional (seems reasonable behaviour to me).

cellio (approved):
I've tested the Sefaria linker with the latest changes and all looks good now.  I assume that other reviewers did a code review and tested Code Golf.

public/assets/community/judaism.js
Oaphi: Let's do this separately, @trichoplax (feel free to file an issue if I don't get to it soon so as we don't forget) - I ended up fixing a bunch of extra stuff in the script, so I don't want to accidentally break those fixes while refactoring.

PR #1947, opened 2026-01-20
Judaism community JS asset cleanup

### LINKED ISSUE #1942: Find out whether var can be removed from Judaism community JS file
Following a [comment on a pull request](https://github.com/codidact/qpixel/pull/1938#discussion_r2659823072), can we test to find out whether the usages of `var` in `public/assets/community/judaism.js` can be safely replaced with `let` or `const`?

This is being raised specifically for the usages of `var t` on lines 154 and 168, but ideally other usages of `var` would also be considered for replacement.

Oaphi:
closes #1942 

This is a minor PR that replaces all instances of `var` declarations with `const` / `let`, improves JSDoc for most of the functions in the script + ensures we don't redefine `pad` shim on every call to `toIsoDate`. No "real" changes - to test, just confirm that the expected functionality is not broken.

Also fixes the Hebrew keyboard's settings as it used to try to set cookies for the `stackexchange.com` domain, which we obviously cannot (and should not) do:

### DIFF
--- a/public/assets/community/judaism.js
+++ b/public/assets/community/judaism.js
@@ -54,8 +54,8 @@ THE SOFTWARE.
 
   /** @type {JQuery<HTMLTextAreaElement | HTMLInputElement>} */
   let currentTextfield = $('textarea, input[type=text]');
-  $(document).ready(function(){
-    $(document).on('focus', 'textarea, input[type=text]', function(){
+  $(document).ready(function () {
+    $(document).on('focus', 'textarea, input[type=text]', function () {
       currentTextfield = $(this);
     });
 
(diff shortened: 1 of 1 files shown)


cellio (approved):
Tested this javascript and the Sefaria linker and calendar widget behave as expected.

public/assets/community/judaism.js
```diff
@@ -22,38 +22,54 @@ THE SOFTWARE.
 // Thanks to all who've helped debug and discuss, especially the Mac users, nebech.
 
 (function HBKeyboard() {
-  var docCookies = { //from developer.mozilla.org/en-US/docs/Web/API/document.cookie
+  // see https://developer.mozilla.org/en-US/docs/Web/API/document.cookie
+  const docCookies = {
+    /**
+     * @param {string} sKey
+     * @returns {string | null}
```
trichoplax, line 29: Non-blocking: Is the `null` serving a purpose here? If the function returned an empty string instead, would this result in the same functionality? If so, is it worth simplifying the return type?

trichoplax (approved):
All looks good. I've added one optional comment in case it's a change you'd like to make while we're here, but happy to approve either way.

public/assets/community/judaism.js
Oaphi: I presume that's for consistency with Web Storage API's `getItem` method that returns either `string` or `null` (many DOM APIs do the same too). I do like it when the return type is an object, not so much when it's a string, yeah. That said, I am inclined to keep the signature as is for consistency.

Oaphi: Oh, upon inspecting the object again to check the method signature, I noticed that `setItem` (probably as a carryover from Somewhere Else) sets cookies with `domain=stackexchange.com`, we should probably handle that in this PR because it (obviously) leads to an expected `Cookie “layoutSetting” has been rejected for invalid domain.` error

Oaphi:
Temporarily blocking this as @trichoplax's question made me notice that the `layoutSetting` cookie never worked as it tries to set cookies for the `stackexchange.com` domain, which is obviously not allowed and errors out (easily reproducible in production, just open devtools to confirm).

cellio:
I forgot until I saw the most recent comments that the Hebrew keyboard is also part of this Javascript file.  Just tested that and it's working properly (inserts characters, is draggable and dismissible).

cellio (approved):

public/assets/community/judaism.js
trichoplax: Oh well spotted with the SE domain! And keeping the signature consistent makes sense - thanks for pointing that out.

trichoplax (approved):

PR #1948, opened 2026-01-20
Minor suspension notice fix

### LINKED ISSUE #1940: Suspension public comment is displayed on user profile even if empty
**Describe the bug**
This is very minor. A space appears before the full stop at the end of the standard suspension message shown at the top of a user's profile. This shows as:

> This user has been **temporarily suspended** .

**To Reproduce**
1. Go to a user page as a mod or admin
2. Click on "Moderator Tools"
3. Click on "warn or suspend user"
4. Suspend the user without choosing a public notice
5. See space before full stop as described above

**Expected behavior**
When there is no public notice to display, there should be no space added.

**Additional context**
In `app/views/users/show.html.erb` on line 13:

```erb
  <% if @user.community_user.suspension_public_comment.nil? %>
```

This appears to never evaluate to true, even if no public notice is selected.

Oaphi:
closes #1940 

This PR fixes extra whitespace appearing in a suspended user's notice if no public reason is provided. Also adds a proper partial for the notice as I don't like PRs that take longer to describe than actually make.

Screenshot from 2026-01-20 04-09-08

With the reason:

Screenshot from 2026-01-20 04-13-10

### DIFF
--- a/app/views/users/notices/_suspension.html.erb
+++ b/app/views/users/notices/_suspension.html.erb
@@ -1,0 +1,17 @@
+<%#
+   "Renders suspension notice for a given user
+
+    Variables:
+      user: User to show the notice for
+"%>
+
+<div class="notice is-danger h-m-b-4">
+  <p>
+    <% if user.community_user.suspension_public_comment.blank? %>
+      This user has been <strong>temporarily suspended</strong>.
+    <% else %>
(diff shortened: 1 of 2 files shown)


trichoplax (approved):
Code changes look good. Message is now correct when there is no suspension public comment, and remains correct when there is a public comment. Full datetime still shows correctly on hover.

PR #1951, opened 2026-01-21
Fix premature confirmation of sign in for users with app 2FA set up

### LINKED ISSUE #1119: Premature confirmation of login with 2FA
https://meta.codidact.com/posts/288069

When you log in, you first enter a name and password, and then, if you're using 2FA, you get a second page where you enter the confirmation code.  On that second page, we display a banner saying "signed in successfully", even though you haven't yet.  Can we not show that until the second factor is validated and you're actually signed in?

I thought this might be a simple matter of front-end HTML, but my search led me to `app/controllers/users/sessions_controller.rb` and I'm not fluent enough to handle it myself.

While in there, please also make the code field active so you can start typing immediately, resolving https://meta.codidact.com/posts/288070.

Oaphi:
closes #1119 (second part of the issue has already been handled in #1428)

2FA verification page after successful sign in with this PR (see the linked issue for the current behavior):

Screenshot from 2026-01-21 02-42-09

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -106,6 +106,10 @@ class Users::SessionsController < Devise::SessionsController
 
   def handle_2fa_login(user, remember_me = false)
     sign_out user
+
+    # prevents premature sign in confirmation
+    flash[:notice] = nil
+
     case user.two_factor_method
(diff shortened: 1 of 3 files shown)


cellio (approved):
Tested, LGTM.

PR #1952, opened 2026-01-21
Fix vote summary badge (tag) misalignment

### LINKED ISSUE #1100: Badges on vote summary page are misaligned
https://meta.codidact.com/posts/287784

I don't know if this is a qpixel issue or a co-design one.

On the user profile "vote summary" tab, if the "badges" for the post type and category don't fit on the same line as the title, they wrap -- and are left-justified, unlike all the others that are right-justified.  Example:

screenshot

Possibly relevant: even though some of those badges would fit to the right of the second line of the title, they instead go to a new line (left-justified).  Ideally, if there's room on the same line as the end of the title, they'd go on that line just like for short titles.  But this might not work because it looks like the badges are slightly taller than the text.  I don't know what the best solution would be.

Oaphi:
closes #1100

Additionally, this PR improves the vote summary's list responsiveness to variable device screen widths:

https://github.com/user-attachments/assets/e6a51af6-f310-4528-b409-8bc7d0eb1167

### DIFF
--- a/app/assets/stylesheets/_variables.scss
+++ b/app/assets/stylesheets/_variables.scss
@@ -62,4 +62,9 @@ $font-sizes: (
   xl: 1.5em
 );
 
+$gap-sizes: (
+  sm: 0.5em,
+  md: 1em
+);
+
 $text-aligns: left, center, right, justify;
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -230,14 +230,31 @@ $sizes: (16, 32, 40, 48, 64, 128, 256);
 
 
   .item-list--item {
+    flex-wrap: nowrap;
     padding: 0 !important;
 
     .item-list--number-value {
-      width: 60px !important;
+      width: max-content !important;
     }
 
     .grid {
       align-items: center;
     }
+  }
+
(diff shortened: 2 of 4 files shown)


cellio:
This is much better!  I was about to say "works for me" when I stumbled over a weird edge case.

Here is a page with a browser width of 600:

looks good

And here is that page just a smidge smaller (594):

original behavior

I'm not sure exactly where in that 6-pixel range it's happening (I'm adjusting window sizes by hand here and using a "how wide am I?" tool to get the values).

If this is really just some weirdness in an extreme edge case, I don't want to block on it.  But I wonder why it happens and if anything else behaves oddly for just a few pixels.

Oaphi:
> I'm not sure exactly where in that 6-pixel range it's happening (I'm adjusting window sizes by hand here and using a "how wide am I?" tool to get the values).

I was worried something like this would happen - our grid is actually a flexbox with `flex-wrap` set to `wrap` which, even with the fix, leads to weirdness like this. Let's switch to a proper grid layout - setting the PR in "draft" mode for now

cellio:
It's better than it was, but I couldn't figure out the cause and implications of this weirdness.

PR #1954, opened 2026-01-22
Keyboard shortcuts improvements

### LINKED ISSUE #646: Unable to comment using keyboard shortcuts
`T->c`, nothing is happening 

Error : Uncaught TypeError: Cannot read property 'scrollIntoView' of undefined

### LINKED ISSUE #1020: Minor indentation misalignment in keyboard shortcuts help
https://meta.codidact.com/posts/287473

In the help box for keyboard shortcuts, the entry for `t` is missing a space of indentation (compared to the other entries).  This is really minor and not worth fixing on its own, but if somebody is working on this help box anyway, please fix this while you're at it.

Oaphi:
closes #1020
closes #646

The PR also makes keyboard shortcut definition dynamic to make it easier to make changes to them going forward.

Here's how the shortcuts will look like with this PR applied (note that the gap between keys and descriptions is now a bit wider to avoid longer keys such as `esc` to look like they are part of the description, that the width is now variable to avoid mostly empty 350px-=wide dialogs, and that there's no longer a "Use a tool (on selection)" shortcut when in post list contexts):

image
Screenshot from 2026-01-22 08-32-16
Screenshot from 2026-01-22 08-32-22
Screenshot from 2026-01-22 08-32-30
Screenshot from 2026-01-22 08-42-18

### DIFF
--- a/app/assets/javascripts/keyboard_tools.js
+++ b/app/assets/javascripts/keyboard_tools.js
@@ -160,6 +160,61 @@ document.addEventListener('DOMContentLoaded', async () => {
     );
   };
 
+  const renderGoToMenu = () => {
+    /** @type {KeyboardShortcut[]} */
+    const shortcuts = [
+      { key: 'm', text: 'Main page' },
+      { key: 'u', text: 'User list' },
+      { key: 'h', text: 'Help' },
+      { key: 'd', text: 'Dashboard' },
+      { key: 'p', text: 'Your profile page' },
+      { key: 'c', text: 'Category ...' },
+      { key: 't', text: 'Tags of category ...' },
+      { key: 'e', text: 'Suggested Edits of ...' },
+      { key: 'f', text: 'Flags (mod only)', if: QPixel.Keyboard.is_mod }
+    ];
+
+    QPixel.Keyboard.dialog(
+      'Go to ...\n' +
+      delimitShortcutsGroup(26) +
+      formatShortcuts(shortcuts, 3)
+    );
+  };
+
+  const renderToolsMenu = () => {
+    /** @type {KeyboardShortcut[]} */
+    const shortcuts = [
+      { key: 'f', text: 'Flag' },
+      { key: 'e', text: 'Edit' },
+      { key: 'c', text: 'Comment' },
+      { key: 'l', text: 'Get permalink' },
+      { key: 'h', text: 'View history' },
+      { key: 'v', text: 'Vote ...' },
+      { key: 't', text: 'Use tools', if: QPixel.Keyboard.is_mod }
+    ];
+
+    QPixel.Keyboard.dialog(
+      'Use tool ...\n' +
+      delimitShortcutsGroup(17) +
+      formatShortcuts(shortcuts, 3)
+    );
+  };
+
+  const renderToolsVoteMenu = () => {
+    /** @type {KeyboardShortcut[]} */
+    const shortcuts = [
+      { key: 'u', text: 'Up' },
+      { key: 'd', text: 'Down' },
+      { key: 'c', text: 'Close' }
+    ];
+    
+    QPixel.Keyboard.dialog(
+      'Vote ...\n' +
+      delimitShortcutsGroup(9) +
+      formatShortcuts(shortcuts, 3)
+    );
+  };
+
   /**
    * Handles the "home" keyboard state
    * @param {JQuery.KeyboardEventBase} e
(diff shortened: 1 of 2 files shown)


cellio (approved):
LGTM.  I tested the various menus and they look fine.  As usual, someone who actually groks Javascript should review/approve because I'm not qualified.

app/assets/javascripts/keyboard_tools.js
```diff
@@ -1,5 +1,13 @@
 window.QPixel ||= {};
 
+/**
+ * @typedef {{
+ *   key: string,
+ *   text: string,
+ *   if?: boolean
```
trichoplax, line 7: Not an objection, just a question:

Are there any problems with using the keyword `if` as a property name? The functionality works fine, but I'm wondering if it could cause confusion for people reading the code in future. Is this a commonly used convention? If not, would it be worth using something that isn't also a keyword?

All I can think of is something like `when` or `condition`, or maybe `only_if`. I'm not particularly fond of any of those, although I like the fact that `condition` is a noun (like `key` and `text`).

This is non-blocking - `if` is short and self-explanatory, and I don't have reason to believe it would cause confusion, it just made me wonder.

app/assets/javascripts/keyboard_tools.js
```diff
@@ -83,54 +98,143 @@ document.addEventListener('DOMContentLoaded', async () => {
     }
   });
 
+  /**
+   * @param {number} len
+   * @returns {string}
+   */
+  const delimitShortcutsGroup = (len) => {
+    return `${'='.repeat(len)}\n`;
```
trichoplax, line 106: Is using `======` as a separator a problem for accessibility?

This approach isn't being introduced by this pull request, so I'm happy for this to be addressed in a separate issue.

If it doesn't cause any other problems, I recommend using `<hr>`, which is inherently accessible.

app/assets/javascripts/keyboard_tools.js
```diff
@@ -83,54 +98,143 @@ document.addEventListener('DOMContentLoaded', async () => {
...
+  };
+
+  /**
+   * @param {KeyboardShortcut} shortcut
+   * @param {number} [gap]
+   * @returns {string}
+   */
+  const formatShortcut = ({ key, text }, gap = 4) => {
+    return `${key}${' '.repeat(gap - (key.length - 1))}${text}`;
```
trichoplax, line 115: Even though this is compatible with all of our current key lengths, we could be defensive here to allow for new longer keys added in future (and in particular translations of keys by non-technical users who might not know to change the value of `gap` if the translation is longer).

The most direct approach would be to impose a minimum for the argument to `repeat` so it can never error by being less than zero (even if the `gap` is too small), and ideally never give an empty string by being equal to zero. For example (untested):

```js
return `${key}${' '.repeat(Math.max(1, gap - (key.length - 1)))}${text}`;
```

This would result in keys of length greater than or equal to `gap` pushing their corresponding text further right, rather than giving a `RangeError`. Is this preferable?

It might be worth instead considering a larger change that removes the need to manually align text, such as an HTML table (perhaps as a separate issue).

trichoplax (dismissed):

app/assets/javascripts/keyboard_tools.js
Oaphi: We can't use anything other than text here - as `dialog` only renders plain text. I'd like us to move away from this text-based formatting to proper HTML, but that's definitely out of scope for the PR.

app/assets/javascripts/keyboard_tools.js
Oaphi: > Are there any problems with using the keyword if as a property name?

Not that I know of, it's an allowed property name (just don't try to destructure it). That said, it's just my personal style (not sure whether it's commonly used), so happy to switch to something else if the consensus is to avoid `if`. @ArtOfCode-?

app/assets/javascripts/keyboard_tools.js
Oaphi: To be honest, I haven't put too much effort in those helpers _precisely_ because I expect us to move away from text-based rendering of the modal sooner or later. Since that thought crossed your mind too, I'd rather tackle it in a separate PR that'll do the switch in one go.

trichoplax: Makes senses. I've now raised #1958 so I don't see any need to make any changes in this pull request, so I'll approve it.

trichoplax (approved):

Oaphi:
When looking through a list of bugs to tackle, I noticed an old issue claiming starting new comments via keyboard shortcuts is completely broken. Bad news is, seems like it's been broken for at least 3 years. Good news is that the PR now fixes that issue as well. To test, open a post, open shortcuts and use the `t -> c` shortcut. If you do that without an open comment thread, the create thread button should be scrolled into view & clicked automatically (opening a new thread form), otherwise, the reply button should be scrolled to & clicked (opening a reply to thread form).

cellio (approved):
Commenting works as described.  AFAIK we don't have a way to expand a comment thread using the keyboard only, so my test was a little "hybrid" -- used the mouse to expand a comment thread on an answer, then used j/k from the keyboard to navigate to that answer and then used the tools to add a reply.  (In the nothing-expanded case it started a new thread, as expected.)

Oaphi:
> AFAIK we don't have a way to expand a comment thread using the keyboard only, so my test was a little "hybrid"

Yeah, had to do the same to test - we should probably add it to keyboard shortcuts (might also be a good idea to start a Meta discussion once this PR lands) - not opposed to just adding this as an additional improvement as well.

cellio:
> Yeah, had to do the same to test - we should probably add it to keyboard shortcuts (might also be a good idea to start a Meta discussion once this PR lands) - not opposed to just adding this as an additional improvement as well.

I would defer.  We'd need a way to navigate threads, and we don't want it to collide with navigating answers, and it's not clear what the best way is to make that work together.  Don't block on it; it's something to decide and handle separately IMO.

ArtOfCode- (approved):

PR #1956, opened 2026-01-22
Fix incorrect handling of rendered tag wikis

### LINKED ISSUE #1059: Tag synonyms break tag wiki
If you go to a normal tag that doesn't have a tag wiki, it looks like this:

image

if there isn't a tag wiki for the tag, and the tag has a synonym, it instead has this weird blank section at the bottom:

image

It seems like it does work when there actually is a tag wiki:

image

Oaphi:
closes #1059

If you want to test that the issue at hand is fixed, create a tag with the only content in its wiki being `[]()`.

### DIFF
--- a/app/helpers/tags_helper.rb
+++ b/app/helpers/tags_helper.rb
@@ -15,6 +15,13 @@ module TagsHelper
       end
   end
 
+  # Renders wiki for a given tag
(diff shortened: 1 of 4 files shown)


cellio (approved):
LGTM - tested and reviewed code.

PR #1957, opened 2026-01-23
Make pagination item disabled state clearer

### LINKED ISSUE #1465: Disabled pagination buttons don't look diabled
meta:291620

A row of pagination buttons includes "previous", numbered pages, and "next".  If one of those isn't actually available, for example "previous" when you're on page 1, you can't click on it -- but it *looks* like you can, because it's the same color or close enough as makes no difference.  Disabled buttons should be visually disabled -- more gray, for example -- and as a stretch goal, if you mouse over them the cursor could change to the "no" symbol.

(Not sure if this is qpixel or co-design.)

Oaphi:
closes #1465

This PR makes disabled state of pagination items clearer (aligned with that of Co-design's disabled state for buttons):

image

It's not shown in the screenshot above, but the cursor now clearly indicates that the action is inactive.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -84,24 +84,6 @@ a.header--site-name {
   }
 }
 
-/* Copied from Co-Design for compatibility purposes */
-ul.pagination li.active {
-  border-color:#293338;
-  background-color:#3f4e56;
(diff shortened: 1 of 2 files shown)


cellio (approved):
Tested & LGTM.

PR #1959, opened 2026-01-23
Allow pinned links (the tool) to be filtered by period

### LINKED ISSUE #1250: Mod tools: move expired featuring to the "inactive" section in the list
meta:289058

Posts can be featured either indefinitely or for a specific date range.  If you cancel the featuring of a post, it's listed in the table as "inactive" (we keep the history).  If it had an end date that has been reached, though, it stays listed as "active" in the table, though it correctly stops being featured.  Over time, these active-not-really posts make it harder to manage the actually-active posts.  This isn't a huge problem and I'm marking the priority as low, but it's something we should fix.

Oaphi:
closes #1250

I chose to add a new filter defaulting to showing only **non-timed or current** links as an alternative to making "expired" pinned links inactive that also allows for more flexible filtering (future only, past only, present only, any period).

Additionally, we now validate that a given link's end date is not before its start date (if both are set, of course) as it's an invalid state.

---

image
image
image
image

### DIFF
--- a/app/models/pinned_link.rb
+++ b/app/models/pinned_link.rb
@@ -12,8 +12,8 @@ class PinnedLink < ApplicationRecord
     timed.where('shown_before < ?', DateTime.now)
   }
 
-  # a present link is not timed or started in the past and ends in the future
-  scope :present, lambda {
+  # a current link is not timed or started in the past and ends in the future
+  scope :current, lambda {
     where(shown_after: nil, shown_before: nil).or(
       timed.where('shown_after <= ?', DateTime.now)
            .where('shown_before > ?', DateTime.now)
--- a/app/models/pinned_link.rb
+++ b/app/models/pinned_link.rb
@@ -7,23 +7,20 @@ class PinnedLink < ApplicationRecord
     includes(:post, post: [:community])
   }
 
-  # a past link is one that ended in the past
   scope :past, lambda {
     timed.where('shown_before < ?', DateTime.now)
   }
 
-  # a current link is not timed or started in the past and ends in the future
   scope :current, lambda {
     where(shown_after: nil, shown_before: nil).or(
-      timed.where('shown_after <= ?', DateTime.now)
-           .where('shown_before > ?', DateTime.now)
+      timed.where('shown_after is null or shown_after <= ?', DateTime.now)
+           .where.not('shown_before < ?', DateTime.now)
     )
   }
 
-  # a future link is one that both starts and ends in the future
   scope :future, lambda {
     timed.where('shown_after > ?', DateTime.now)
-         .where('shown_before > ?', DateTime.now)
+         .where('shown_before is null or shown_before >= ?', DateTime.now)
   }
 
   scope :timed, lambda {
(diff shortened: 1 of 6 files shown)


app/models/pinned_link.rb
```diff
@@ -7,6 +7,31 @@ class PinnedLink < ApplicationRecord
...
+  # a past link is one that ended in the past
+  scope :past, lambda {
+    timed.where('shown_before < ?', DateTime.now)
+  }
+
+  # a present link is not timed or started in the past and ends in the future
+  scope :present, lambda {
+    where(shown_after: nil, shown_before: nil).or(
+      timed.where('shown_after <= ?', DateTime.now)
+           .where('shown_before > ?', DateTime.now)
```
cellio: Why do we require a start time to count as "present" if there is a future end time?  I can't yet tell (because caching) whether an event with only a (future) end time is shown, but I would expect it to be.  Am I missing something?

Currently, I have an event that does not show up under any of past, present, future, but is there under "all".  I was expecting "all" to be the union of the other three.

app/models/pinned_link.rb
```diff
@@ -7,6 +7,31 @@ class PinnedLink < ApplicationRecord
...
+    )
+  }
+
+  # a future link is one that both starts and ends in the future
+  scope :future, lambda {
```
cellio, line 21: Similarly, if an event starts in the future, I expected that to be enough to treat it as a future event (show up under "future", appear as a featured link when the time comes).

cellio (commented):
Tested and the layout looks good.  (And the buttons work better than the table itself on a phone. :-) )  I have questions about the meaning of time -- specifically, I expected a past event to be one that's already ended and a future event to be one that hasn't started yet, but the code seems to care about both start and end dates.  It's possible to create an event with only one of those values and it feels like we should loosen up our definitions of past/future accordingly.

Oaphi:
A quick heads up, @cellio - I renamed the `present` filter to `current` as I realized its predicate form `present?` clashes with the built-in method of the same name in a very confusing way. No changes otherwise

cellio:
> A quick heads up, @cellio - I renamed the `present` filter to `current` as I realized its predicate form `present?` clashes with the built-in method of the same name in a very confusing way. No other changes otherwise

That's much better, thanks!

app/models/pinned_link.rb
Oaphi: That was an attempt to prevent misconfigurations like this (as, unfortunately, we don't have _any_ validation for this right now) from showing up in respective tabs (prompting the user to take another look at the config):

image

Judging by the other comment (which seems like a bug in my implementation) and this one, this approach is not that much better, let's do it the "proper" way. Will update the PR shortly

cellio: I didn't realize we didn't enforce start before end.  Improving event validation can be a separate matter; in a sense, if you create a malformed event, you should expect to see odd results here.  Really we should block the malformed event in the first place, but that's a lot harder and shouldn't block this.

Oaphi:
Finished with the rework (including basic validation for period upon trying to save a link), @cellio, should be ready for review

cellio (approved):
Tested and LGTM.  I'm glad adding the validation wasn't as much work as I'd thought it would be (I thought you'd have to do more front-end stuff).

Oaphi:
> 'm glad adding the validation wasn't as much work as I'd thought it would be (I thought you'd have to do more front-end stuff).

It _did_ help that I'd already done all the preparatory work in #1727 :)

ArtOfCode- (approved):

PR #1961, opened 2026-01-24
Remove all hardcoded Codidactyl references from QPixel

### LINKED ISSUE #1396: Reference to the Codidactyl in non-Codidact instances

Oaphi:
closes #1396

This PR removes all remaining hardcoded Codidactyl references from QPixel.

To that end, it adds several new site settings:

- `NetworkMascotName` (defaults to our current `Codidactyl`);
- `NetworkMascotLogoPath` (defaults to our current `/assets/codidactyl.png`);
- `TourBeginningDescription` - this is the hint displayed on the first page of the Tour (`/tour`);
- `FriendlyEmailFooter` - this is the footer that we add to Devise-related emails (the simplest way to test is to change a user's password to their current one);

The PR also replaces our old deleted user page (the one that users that CANNOT see deleted users see) with our common 404 page (it doesn't make sense to expose information that the user is deleted if the viewer is not supposed to see them anyway):

Screenshot from 2026-01-24 04-56-32

There are also a couple of minor tweaks to ensure we don't hardcode Codidactyl anywhere.


db/seeds/site_settings.yml
```diff
@@ -681,6 +703,22 @@
     The content of a post is shown in short in lists (e.g. category post overview or in search).
     This setting controls how many characters of a post are shown.
 
+- name: NetworkMascotName
+  value: Codidactyl
+  value_type: string
+  category: SiteDetails
+  community_id: ~
+  description: >
+    Network mascot name tto use where appropriate.
```
cellio: Could be cryptic to people running other networks (people who found us via GitHub etc, rather than via our own network).  I'd add something like "(for example, at the beginning of the [tour]" and make that a relative link.

config/application.rb
```diff
@@ -36,6 +36,7 @@ class Application < Rails::Application
     config.i18n.default_locale = :en
 
     config.to_prepare do
+      Devise::Mailer.helper :application
```
cellio, line 39: What does this do?

cellio (commented):
Tested and works as advertised.  After changing these settings I ran through the tour checking that the name/icon were updated, generated email, and tested deleted users and deleted posts for anonymous users.  One very minor suggestion for documentation of a setting, and one question 'cause I don't really know Devise, but LGTM otherwise.

config/application.rb
Oaphi: A bit of Rails magic, as usual :) `helper` basically includes the module - in this case, `ApplicationHelper` (denoted as a symbol without the `helper` part). It's there to make `ApplicationHelper` methods accessible to the Devise mailer (as we now use `render_markdown` in some of the templates)

db/seeds/site_settings.yml
Oaphi: Agreed - done (also noticed I made a typo there - `tto`, oops)

config/application.rb
cellio: Thanks.  Part of my confusion was "why now? what changed?" (since we only changed footer text), but I hadn't considered that it would now need access to more methods too.

cellio (approved):

config/application.rb
Oaphi: Yeah, it's not exactly obvious unless you know exactly what's going on (our setup is a bit weird here - it's the only mailer that is configured in this manner)

ArtOfCode- (approved):

PR #1962, opened 2026-01-25
Show community logo & name for inbox notifications

### LINKED ISSUE #1599: Ability notifications should say which community they're from
meta:293757

When you get a new ability, including the initial ones when you join a community, we send a notification.  When a new user joins several communities, the notifications are confusing -- you see a bunch of "you've earned Participate" notices, and people think these are duplicates, but what's actually happening is that abilities are community-specific and you're getting a notice for each place you earn it.

We could avert a lot of confusion if the notice included the community name, something like "you've earned the Participate ability on Software Development".

Oaphi:
closes #1599 

This PR adds community icons & names to inbox notifications. We don't have (at least as far as I am aware) proper small square icons of our communities for this list, so I opted to simply use `SiteLogoPath` - it doesn't look ideal, but better than nothing (if we _do_ have proper icons or have some designer resources to spare, please let me know):

image

Also fixes CSS specificity shenanigans breaking site header if both `MobileLogoPath` and `SiteLogoPath` are set.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -43,14 +43,6 @@ a {
   overflow-wrap: break-word;
 }
 
-a.header--site-name {
-  display: flex !important;
-}
-.header .header--item.is-complex:not(.is-mobile-menu) {
-  align-items: center !important;
-  display: flex !important;
-}
-
 @media all and (max-width: 73.5em) {
   .grid {
     margin-left: 0;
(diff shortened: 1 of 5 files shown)


cellio (approved):
LGTM after the scoping fix:

screenshot w/different logos for different communities

cellio (approved):
Tested and everything works as expected.  I'm not too good with CSS, so probably want to have someone else review those changes.

ArtOfCode- (approved):

PR #1976, opened 2026-02-03
General fixes & improvements

### LINKED ISSUE #1671: Silent error when upload size limit does not match server configuration
**Describe the bug**
Not a problem for Codidact in production, as the file upload limit is set to 2MB and works correctly, but if another instance of QPixel has the server configuration set to a value smaller than specified in the admin interface then the user will be given no error message, but files of sizes between the 2 different limits will silently fail. To a user who doesn't look in the developer window, it will look like the file is in the process of uploading, but it will never finish.

**To Reproduce**
At the time of raising this, the dev server has this mismatch, so you can see the lack of error message there.

1. Go to the edit page for a new/existing post.
2. Click on the file upload button.
3. Select a file that is a little under 2MB.
4. See no indication that anything is wrong.
5. The developer console shows an error message `Uncaught (in promise) SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
` and the upload line in the Network tab shows a response of `413 Request Entity Too Large` from nginx.

**Expected behavior**
If the response is not JSON there should be a message to the user indicating that the upload failed, and the `Uploading, please wait...` should be removed from the post body.

### LINKED ISSUE #1884: Remove livestampjs
We use [Livestamp.js](https://github.com/mattbradley/livestampjs) for auto-updating timestamps. Unfortunately, it:

- hasn't been maintained in about a _decade_;
- doesn't have types;
- relies on jQuery that we intend to eventually phase out;

Steps we need to take to remove it:

- [ ] find a replacement library;
- [ ] switch notifications to the replacement;

### LINKED ISSUE #1965: Make notification times visible in drop-down
meta:294223

The notifications drop-down shows relative times (1 day ago, etc).  Usually when we show relative times, the actual times are available on hover -- but that's not the case here.  I also noticed that those times flash briefly before being replaced by the relative times, so the notification dropdown has the information.

Can we show the timestamp on hover like we do elsewhere?

### LINKED ISSUE #1966: Inbox page uses non-standard date formats
meta:294223

Most of the time we show either ISO time or relative time, but the inbox page has dates like "January 29, 2026".  In addition to being different, this removes the fact that the times are UTC, not local time zone.  For consistency we should have the inbox page show the full timestamps.

### LINKED ISSUE #1968: Inconsistent naming of comments
Under a post it is possible to add a comment to an existing thread or create a new thread. When creating a new thread the comment field is labelled "Your comment" but when adding to an existing thread the comment field is labelled "Your message":

Comment section with both types of field showing, with Your message and Your comment highlighted

Should both of these say "Your comment" for consistency?

### LINKED ISSUE #1969: Improving the wording when creating a new comment thread
When creating a new comment thread under a post that has existing comment threads, a message is displayed:

> **Start a new comment thread**
>
> You can create a new comment thread. Please check first whether an existing comment thread covers already what you need.

> The message wording in context, highlighted

Since the first sentence duplicates the heading, could this be omitted to reduce clutter? The second sentence could also be made shorter without losing its meaning. This would give:

> **Start a new comment thread**
>
> Please consider whether an existing comment thread covers what you need.

Would this new wording be acceptable?

### LINKED ISSUE #1970: Slight misalignment in stats table on profile page
On the right hand panel of the user profile page, immediately below the avatar, there is a table of statistics. The first row ("Reputation") is slightly misaligned, starting a little to the left of all of the other rows:

The user avatar with the stats table below, showing the first column having a misalignment in the first row

### LINKED ISSUE #1971: Formatting problem with search field
On the search page (including the initial search page and the results page) the search field is restricted to a small section of the right hand side of the available space when the window width is 779px or less. This can affect narrow windows, zoomed in views, and phones in vertical (portrait) orientation.

> Search field taking up less than half of the available width

### LINKED ISSUE #1975: Sort order is not maintained when showing more threads
Clicking "Show more" under a post that has more than 5 comment threads shows more comment thread titles, but the list of comment thread titles is no longer in the same order:

> Image
> Image

Should the longer list still show in the same order as the shorter list?

### LINKED ISSUE #1977: Uploading a 6 MiB avatar returns “413 Request Entity Too Large”.
meta:295396

#### What The Problem Is

When I attempt to upload a new avatar, if the submitted file is ⪆ 4 MiB:

1.	~~~sh
	#!/usr/bin/env sh
	mediainfo --Output=JSON "$HOME/JPEG.JPEG" | yq -P
	~~~

1.	<blockquote>

	~~~YAML
	creatingLibrary:
	  name: MediaInfoLib
	  version: "25.10"
	media:
	  '@ref': /home/RokeJulianLockhart/JPEG.JPEG
	  track:
	    - '@type': General
	      ImageCount: "1"
	      FileExtension: JPEG
	      Format: JPEG
	      FileSize: "6404703"
	      StreamSize: "17260"
	    - '@type': Image
	      Format: JPEG
	      Width: "3072"
	      Height: "3072"
	      ColorSpace: YUV
	      ChromaSubsampling: 4:4:4
	      BitDepth: "8"
	      Compression_Mode: Lossy
	      StreamSize: "6387443"
	      colour_primaries: BT.709
	      transfer_characteristics: sRGB/sYCC
	      extra:
	        ColorSpace_ICC: RGB
	        colour_primaries_ICC_Description: GIMP built-in sRGB
	~~~

	</blockquote>

…I observe a nondescript error message:

<blockquote>

~~~HTML
<html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx/1.18.0 (Ubuntu)</center>
<script defer src="[https://static.cloudflareinsights.com/beacon.min.js/vcd15cbe7772f49c399c6a5babf22c1241717689176015](view-source:https://static.cloudflareinsights.com/beacon.min.js/vcd15cbe7772f49c399c6a5babf22c1241717689176015)" integrity="sha512-ZpsOmlRQV6y907TI0dKBHq9Md29nnaEIPlkf84rnaERnq6zvWvPUqr2ft8M1aS28oN72PdrCzSjY4U6VaAw1EQ==" data-cf-beacon='{"version":"2024.11.0","token":"fa0c13a4688940dfbd0ead60ec559af0","server_timing":{"name":{"cfCacheStatus":true,"cfEdge":true,"cfExtPri":true,"cfL4":true,"cfOrigin":true,"cfSpeedBrain":true},"location_startswith":null}}' crossorigin="anonymous"></script>
</body>
</html>
~~~

</blockquote>

#### How To Reproduce It

1.	~~~pwsh
	#!/usr/bin/env pwsh
	Start-Process 'https://powerusers.codidact.com/users/edit/profile#user_avatar:~:text=user%20avatar-,Avatar,-An%20optional%20profile'
	~~~

1.	~~~JS
	#!/usr/bin/env xdg-open
	(() => {
	  const run = () => {
	    const fileInput = document.querySelector('#user_avatar');
	    const onChange = () => {
	      if (fileInput.files && fileInput.files.length > 0) {
	        document.querySelector('input.button').click();
	      }
	      fileInput.removeEventListener('change', onChange);
	    };
	    fileInput.addEventListener('change', onChange, { once: true });
	    fileInput.click();
	  };
	  if (document.readyState === 'loading') {
	    document.addEventListener('DOMContentLoaded', run, { once: true });
	  } else {
	    run();
	  }
	})();
	~~~

If that fails:

1. Visit [`powerusers.codidact.com/users/edit/profile#user_avatar`](https://powerusers.codidact.com/users/edit/profile#user_avatar:~:text=user%20avatar-,Avatar,-An%20optional%20profile).

1. Select “Choose”, under “Avatar”.

1. Select an image > 6 MiB.

1. Select “Save”.

To demonstrate this:

https://github.com/user-attachments/assets/82dc40e8-99b7-458c-af5c-372060b609ea

#### The Expected Behaviour

I should be informed that the image is too large, after which QPixel should allow me to re-upload a new avatar file.

Oaphi:
closes #1968 (the label is now configurable via locale strings):

Screenshot from 2026-02-03 17-39-50

closes #1971:

Screenshot from 2026-02-03 18-17-51
Screenshot from 2026-02-03 18-18-02
Screenshot from 2026-02-03 18-18-15

closes #1975 - comment threads now use the `priority_order` common scope regardless of whether they are initially loaded or expanded via the "show more" button (note that we still order by `updated_at` and not `last_activity_at` - we should switch to the latter when we can).

closes #1970:

image

closes #1965 (see below);
closes #1966 (see below);
closes #1884 (we now simply use Moment.js to format timestamps - note the new format and preserved relative time):

closes #1977 (also exposes `MaxUploadSize` to client-side code as `QPixel.MAX_UPLOAD_SIZE`):
Screenshot from 2026-02-05 16-19-05

closes #1969 (with existing threads and no threads respectively):

image
image

closes #1671 (new site setting: `MaxRequestBodySize`):
image


cellio:
Tested:

- [x] comment label (1968)
- [x] search width (1971)
- [x] thread order (1975)
- [x] profile stats alignment (1970)
- [x] notification times (1965, 1884)
- [x] inbox date format (1966)
- [x] large avatar images (1977)
- [x] new comment wording (1969)
- [x] silent error for too-large image (1671)

~~What user testing, if any, applies for https://github.com/codidact/qpixel/issues/1884 and https://github.com/codidact/qpixel/issues/1671?~~

db/seeds/site_settings.yml
```diff
@@ -189,6 +189,15 @@
     set by your web server configuration. This setting should be set to the same value so that the correct limit
     can be displayed to users.
 
+- name: MaxRequestBodySize
+  value: 2097152
```
Oaphi, line 184: 2MB (1024 * 1024 * 2)

trichoplax: Is this worth a comment in the file?

Oaphi: Yeah, I'll add a note in the setting's description

db/seeds.rb
```diff
@@ -199,6 +212,29 @@ def seed_community_assets
...
+  end
+
+  return unless to_remove.any?
+
+  ActiveRecord::Base.transaction do
+    to_remove.each(&:destroy!)
+  end
+
+  puts "#{SiteSetting.model_name}: removed #{to_remove.size}"
+end
```
Oaphi, line 238: After this PR is merged, we'll start automatically removing `SiteSetting`s that are no longer present in seeds. My hope is that it'll evolve into a proper cleanup step in the future and then into a full solution for dynamically changing seeds.

cellio: Question: if a setting is no longer in seeds but is in the DB, then code that still checks for that setting would continue to work in existing deployments and fail in new ones.  How do we make sure that there's no referring code before removing, and especially auto-removing, a setting?  I love having cleanup but do want to make sure we only clean up truly unused stuff.

Oaphi: We can't - well, we *technically* can, but that would be an unreasonable thing to automate. For the cleanup to trigger, one has to manually remove the setting from the seed file - at which point it's entirely on them to ensure there's nothing referencing the setting in code, we can't do anything about that. And if _we_ remove a setting (as is the case with `MaxUploadSize` here), the onus is on us to update the codebase to not use it anymore.

Oaphi:
> What user testing, if any, applies for https://github.com/codidact/qpixel/issues/1884 and https://github.com/codidact/qpixel/issues/1671?

For #1884, none is needed as long as the new timestamp format works. For #1671 one could tweak the new site setting (`MaxRequestBodySize`), then attempt to upload a file of a size bigger than the new limit. When pasting the file directly to the new post body field, there should be a notification mentioning the limit. When uploading via the form, it should be mentioned in the caption + correctly error out. NGINX can be tested once the change is on the dev server

db/seeds.rb
cellio: Ah, ok.  I assume grepping the code for the name of the setting you're removing is sufficient to catch any remaining references?

Oaphi: Yup, there should be no code referencing the old setting (I thoroughly purged it before dropping, of course) but double-checking won't hurt

cellio: My question wasn't about *this* change, but future ones. :-)  Is there a comment in the YAML file to the effect that removing an entry will purge it (not just stop seeding it)?  Belt and suspenders isn't bad when it comes to comments for future editors.

Oaphi: I have hard time imagining someone _removing_ a baked in site setting without checking how it's used and expecting it to not break anything, but I suppose a mention in the docs won't hurt. We might want the cleanup to be locked behind an environment variable similar to `UPDATE_POSTS` just to be on the safe side (it's not like having a dangling site setting is the end of the world). @ArtOfCode-, what do you think about adding one?

cellio: I'm probably being needlessly paranoid -- that PR would get reviewed, after all, and we'd have a conversation like this if it didn't mention calling code or the lack thereof.  I think we can leave it as-is.  (Feel free to resolve this conversation.)

db/seeds.rb
```diff
@@ -199,6 +212,31 @@ def seed_community_assets
   end
 end
 
+def cleanup_posts
```
cellio, line 215: Question (possibly for @ArtOfCode- ): are there any Posts that are community-agnostic and could be caught by this?  (Help?  Abilities pages?  Any site boilerplate?)  Should we do any sort of logging/output (here and for site settings below), just in case any manual recovery were ever needed?

cellio (commented):
Tested everything listed (and checked off) in a previous comment and it all looks good.  I have a (paranoid?) question about deleted posts, and someone more qualified should review new JS and the JSON & config changes in the `db` directory.  (I looked at all the other code changes and they look fine to me.)

db/seeds.rb
Oaphi: Do note that the post removal logic has existed for a very, very long time - the PR only wraps it into a method. IIRC, we don't have community-agnostic posts in the first place, help seeds are simply created / updated for every existing community.

cellio: Oh, I thought from the diff that this was new.  Works for me.  (My main concern with deleting help wouldn't be deleting the help that we could re-seed, but rather, deleting something that was locally modified.  We'd have to go to backups for that if it got damaged.)

app/assets/javascripts/relative_time.js
```diff
@@ -0,0 +1,47 @@
...
+
+    if (elapsed < updateInterval && lastRunAt !== -1) {
+      requestAnimationFrame(updateRelativeTime);
+      return;
+    }
+
+    document.querySelectorAll('[data-relstamp]').forEach(processNode);
+
+    lastRunAt = timestamp;
+    requestAnimationFrame(updateRelativeTime);
```
trichoplax: Using `requestAnimationFrame` results in this function being called every frame. This is 60 times per second on my machine, and will be more on some machines. I don't expect this to be a performance issue even on older machines since it's not a demanding function, but I'd still prefer to avoid unnecessary work, particular for battery powered devices such as phones, tablets, and laptops.

Would [setTimeout](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout) be a suitable alternative here? This would allow only calling the function once per minute (to reflect `updateInterval`). If there is concern about this sometimes misaligning and updating the timestamps up to a minute late, we could call it once per second instead, rather than 60 times per second.

app/views/layouts/_expose.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+  Object.defineProperty(QPixel, 'LOCALE', {
+    configurable: false,
+    value: "<%= I18n.locale %>",
+    writable: false,
+  });
+
+  Object.defineProperty(QPixel, 'MAX_UPLOAD_SIZE', {
+    configurable: false,
+    value: <%= SiteSetting['MaxRequestBodySize'] %> ?? 2 * 1024 * 1024,
```
trichoplax: When the setting does not exist, `<%= SiteSetting['MaxRequestBodySize'] %>` appears to resolve to an empty string instead of `undefined`. This results in `value: ?? 2 * 1024 * 1024,` in the JavaScript provided to the page, which gives a syntax error in the browser's developer tools.

app/views/layouts/_header.html.erb
```diff
@@ -174,7 +174,7 @@
       <span>votes:4</span> <span class="has-color-tertiary-600">posts with 4+ votes</span>
     </div>
     <div class="grid--cell is-6 is-12-sm">
-      <span>created:<1w</span> <span class="has-color-tertiary-600">created < 1 week ago</span>
+      <span>created:&lt;1w</span> <span class="has-color-tertiary-600">created < 1 week ago</span>
```
trichoplax, line 177: Not a problem, just a question:

The less than sign in `<1w` appears to already be displaying correctly in the `develop` branch, before this change to an HTML entity.

The less than sign in `created < 1 week ago` that hasn't been changed to an HTML entity seems to be displaying correctly too.

Is there a reason to escape one but not the other? Is this required for some browsers when there are no surrounding spaces?

trichoplax (changes requested):
I have a couple of requests and a question.

app/views/layouts/_header.html.erb
Oaphi: The difference is that the first instance is followed by a word character, which usually indicates an HTML tag, while the second is not - I only noticed because my IDEs highlighting got very confused due to this and thus fixed it in a "while I am at it" manner. Not opposed to changing the other instance too (in fact, we probably should do that across the project someday.

trichoplax: I was just curious - thanks for explaining. Changing just the one you have seems fine, and I see no reason to change any others in this pull request - "someday" sounds good for changes that don't affect the users.

app/views/layouts/_expose.html.erb
Oaphi: Whoops, good catch, fixed

app/assets/javascripts/relative_time.js
Oaphi: Do note that the only thing the callback executes every animation frame is checking whether the interval has elapsed - which is very lightweight. That said, I opted for RAF before choosing to only update about once per minute - `setTimeout` should do just fine here, yeah.

trichoplax (approved):
All looks good.

db/seeds.rb
Oaphi: Changes moving code around can be a bit hard to spot, yeah

Issue #1995, opened 2026-02-27
Claiming SE content doesn't seem to work anymore + its failures shouldn't be silent


Oaphi:
@trichoplax's attempt to merge in their imported user silently failed with no way for us to even trace the error. The process is likely broken, but we should also add at least an error log if it fails. See the `transfer_se_content` method on the users controller for details on how the actual transfer is performed.

PR #1996, opened 2026-02-27
Add job-specific logging

ArtOfCode-:
Related to #1995.

Adds a job-specific logger tagged with the job class and randomised ID, and adds logging to our existing jobs.


app/jobs/application_job.rb
```diff
@@ -1,7 +1,16 @@
 class ApplicationJob < ActiveJob::Base
   # Automatically retry jobs that encountered a deadlock
-  # retry_on ActiveRecord::Deadlocked
+  retry_on ActiveRecord::Deadlocked
 
   # Most jobs are safe to ignore if the underlying records are no longer available
```
trichoplax, line 5: Can the word "Most" be dropped from this comment? If not, do we need a check here for whatever jobs are not safe to ignore in this context?

app/jobs/auto_close_complaints_job.rb
```diff
@@ -4,8 +4,10 @@ class AutoCloseComplaintsJob < ApplicationJob
   def perform(*_args)
     complaints = Complaint.where(status: 'reviewed')
                           .where(Arel.sql('status_updated_at <= ?', 14.days.ago))
-    complaints.each do |complaint|
+    statuses = complaints.map do |complaint|
       complaint.update_status 'closed'
     end
+    successful = statuses.map { |x| (x && 1) || 0 }.sum
```
Oaphi: Non-blocking: wouldn't it be better if we used standard API here and just [`compact_blank`](https://api.rubyonrails.org/v7.1.2/classes/Enumerable.html#method-i-compact_blank) instead? Like this:
```ruby
statuses.compact_blank.size
```

app/jobs/application_job.rb
ArtOfCode-: Unknown - this comment came from the Rails framework, it's their recommended default

app/jobs/auto_close_complaints_job.rb
ArtOfCode-: Yep - that's a nice Active Support trick I didn't realise existed.

app/jobs/application_job.rb
trichoplax: Oh I see. I have no insight here - happy to go with your judgement call.

config/initializers/job_logger.rb
```diff
@@ -0,0 +1,15 @@
+module Rails
+  class << self
+    @job_logger = nil
+
+    def job_logger
+      if @job_logger.nil?
+        logger = ActiveSupport::Logger.new(Rails.root.join('log/jobs.log'))
```
Oaphi, line 7: Non-blocking: since we are adding a singular log file, wouldn't it make sense to add a shared tag for every job logger indicating current `Rails.env` (unless it's done under the hood, of course)?

ArtOfCode-: I considered it, but we really only ever run one environment on one system - prod never runs dev, dev never runs prod etc, so I wasn't too worried about it.

app/jobs/cleanup_votes_job.rb
```diff
@@ -3,28 +3,33 @@ class CleanupVotesJob < ApplicationJob
...
+              event_type: 'vote_delete',
+              related: v,
+              user: system_user
+            )
+          else
+            logger.warn "Failed to remove vote #{v.id}. Validations follow."
+            v.errors.full_messages.each do |msg|
+              logger.warn msg
+            end
+            v.errors.each { |e| puts e.full_message }
```
Oaphi: Question: since we are adding a proper logger, do we even need to `puts` errors anymore?

config/initializers/job_logger.rb
Oaphi: Up to you then - we can always add the tag later if it turns out to be relevant

config/initializers/job_logger.rb
```diff
@@ -0,0 +1,16 @@
+module Rails
+  class << self
+    @job_logger = nil
+
+    def job_logger
+      if @job_logger.nil?
+        logger = ActiveSupport::Logger.new(Rails.root.join('log/jobs.log'))
+        logger.level = ActiveSupport::Logger::INFO
```
Oaphi: Non-blocking: should we respect configuration `log_level` here? [Source for reference](https://github.com/ruby/logger/blob/master/lib/logger/severity.rb)

```ruby
logger.level = ActiveSupport::Logger::Severity.coerce(Rails.configuration.log_level)
```
Or even simpler (as I've just learned [`level=`](https://github.com/ruby/logger/blob/aa47f3a9e0c6245bb6fabbdb32f25017712e4a9a/lib/logger.rb#L399C1-L401) calls `coerce` under the hood):
```ruby
logger.level = Rails.configuration.log_level
```

app/jobs/cleanup_votes_job.rb
ArtOfCode-: Missed that one

config/initializers/job_logger.rb
ArtOfCode-: I was in two minds over whether we follow config or whether we actually _always_ want the jobs information regardless. Not tied to this so since you had the same thought I'll change it

app/jobs/clean_up_thread_followers_job.rb
```diff
@@ -15,10 +15,15 @@ def perform
...
-      unless result
-        puts "failed to destroy thread follower duplicate \"#{duplicate.id}\""
-        duplicate.errors.each { |e| puts e.full_message }
+      next if result
+
+      logger.warn "Failed to destroy thread follower duplicate \"#{duplicate.id}. Validations follow."
+      duplicate.errors.full_messages.each do |msg|
+        logger.warn msg
       end
+      duplicate.errors.each { |e| puts e.full_message }
```
Oaphi: We can also drop (the `puts` one) this line (as we decided those are no longer needed)

app/jobs/clean_up_new_thread_followers_job.rb
```diff
@@ -15,10 +15,15 @@ def perform
...
-      unless result
-        puts "failed to destroy new thread follower duplicate \"#{duplicate.id}\""
-        duplicate.errors.each { |e| puts e.full_message }
+      next if result
+
+      logger.warn "Failed to destroy new thread follower duplicate \"#{duplicate.id}. Validations follow."
+      duplicate.errors.full_messages.each do |msg|
+        logger.warn msg
       end
+      duplicate.errors.each { |e| puts e.full_message }
```
Oaphi: We can drop this (the `puts` one) line too (as we decided those are no longer needed)

Oaphi (approved):
LGTM - not worried about coverage here, it's mostly because we barely have any for jobs in the first place.

PR #2001, opened 2026-03-02
Fix scheduled jobs

ArtOfCode-:
Changes our scripts to use `perform_now` instead of `perform_later` so that jobs actually run. Also removes a few old one-time maintenance scripts that are no longer required, and moves a couple of old scripts to jobs instead.


app/jobs/mail_uncaptured_donations_job.rb
```diff
@@ -0,0 +1,29 @@
+class MailUncapturedDonationsJob < ApplicationJob
+  queue_as :default
+
+  def perform(*_args)
```
Oaphi, line 4: Non-blocking: I don't think we need to parametrize it all, do we? We don't use the param anyway, so let's avoid creating mental overhead for those less familiar with the syntax.

Oaphi (approved):
LGTM - low-ish coverage is fine here as there are (a) no changes in how the jobs function, only in how they are executed (+ some of the scripts are wrapped into proper jobs) and (b) it's due to low initial coverage for jobs in general.

PR #2004, opened 2026-03-03
Online Safety Act updates

ArtOfCode-:
* Add options for complaints about process adherence or failure in statutory duty


app/views/complaints/training/overview.html.erb
```diff
@@ -0,0 +1,64 @@
...
+<p>
+  The Act defines two types of service: search services and user-to-user services. User-to-user services are those where
+  users may interact with one another; this is where we fall. There are different requirements imposed on each kind of
+  service, which for user-to-user services primarily focus on preventing and removing harmful content, and protecting
+  users from related harms.
+</p>
+
+<h3>Our responsibilities</h3>
+<p>
+  Responsibility for compliance with the requirements of the Act obviously falls on us (meaning the Codidact Foundation
```
cellio, line 28: Is it practical to make this part configurable (org name and contact as global settings, whole paragraph as a setting, something else)?  You probably don't want your name leaking out as the contact person for other instances.

Or, alternatively, put the entire training behind a global setting, so that instances that don't care (such as private ones outside the UK) don't need to edit contacts etc at all.

app/views/complaints/training/overview.html.erb
```diff
@@ -0,0 +1,64 @@
...
+  escalated appropriately.
+</p>
+<p>
+  To be clear: we're <strong>not</strong> expecting you to handle harmful or illegal content alone. Our ask of you is
+  simple: if you identify something that you think would be covered by the Act, please:
+</p>
+<ul>
+  <li>Delete it to remove it from public view</li>
+  <li>
+    Escalate it to the Community Team straight away: either ping us in Discord, or flag it yourself and escalate your
```
cellio, line 54: Should they use the reporting form for this?  If it *is* OSA-grade content we'd need to file a report ourselves anyway, so why not have the mods do it and use our existing notification channels?  Plus, this gives us a way to have back-and-forth discussion on the record if we need to.

app/views/complaints/training/definitions.html.erb
```diff
@@ -0,0 +1,274 @@
...
+      work, and (b) consenting adult sex workers who are controlled or exploited by another person.
+    </div>
+  </div>
+  <div class="widget">
+    <div class="widget--body">
+      <h4>Human trafficking</h4>
+      Human trafficking encompasses a wide range of harmful activities. It can involve modern slavery, and victims and
+      survivors include adults and children. It is estimated that there were 122,000 people living in modern slavery in
+      the UK in 2021. Notable forms of human trafficking where harm can manifest online include sexual exploitation and
+      abuse, forced labour, and criminal exploitation such as county lines exportation of illegal drugs.
```
cellio: I'm having trouble parsing "criminal exploitation such as county lines exportation of illegal drugs" around "county lines".

app/views/complaints/training/definitions.html.erb
```diff
@@ -0,0 +1,274 @@
...
+      to engage in harmful or illegal behaviours and activities themselves.
+    </div>
+  </div>
+</div>
+
+<h3>Non-Priority Illegal Content</h3>
+<div class="has-font-size-caption">
+  <div class="widget">
+    <div class="widget--body">
+      <h4>Epilepsy trolling</h4>
```
cellio, line 192: I expect we'll get questions about this, since what's problematic for epilepsy sufferers is not well-known beyond rapidly flashing lights.  Can we say a little more about what characteristics such content would have?  (I mean a sentence or two, not a paragraph.)

app/views/complaints/training/handling.html.erb
```diff
@@ -0,0 +1,52 @@
+<p class="has-font-size-caption has-color-tertiary-600">Last updated 22 March 2026</p>
+
+<h1 class="is-brand-comp with-subtitle">Online Safety Act</h1>
+<h2 class="subtitle">Handling Illegal Content</h2>
+
+<p>
+  It's important that illegal content is handled consistently in order to ensure that we are able to meet the statutory
+  requirements of us, to protect our users, and to enable us to be transparent with our communities. The way in which
+  illegal content is handled is set out in our Online Safety & Illegal Content Policy & Procedure, OP02 &mdash; you can
+  find a copy of this <a href="https://codidact.org/OP02">here</a>.
```
cellio: Not yet.  Also, might be another case where we want configuration rather than a baked-in link.

app/views/complaints/training/handling.html.erb
```diff
@@ -0,0 +1,52 @@
...
+
+<h3>As a volunteer moderator</h3>
+<p>
+  The two ways in which you are likely to come across illegal content as a volunteer moderator are either through
+  encountering it as you're browsing, or through flags from community members. Either way, our ask of you is the same.
+  If you believe the content may be illegal:
+</p>
+<ul>
+  <li>
+    Delete it, using the "delete" button for the post or comment, to remove it from public view. Deletion doesn't
```
cellio: I would remove the "doesn't completely remove" part, as we're not asking mods to make legal determinations anyway -- we're asking them to remove it from (most) view and raise it with us.

app/views/complaints/training/handling.html.erb
```diff
@@ -0,0 +1,52 @@
...
+    completely remove a post and leaves it visible for some community members, so in addition:
+  </li>
+  <li>
+    Escalate it to the Community Team. If a community member has flagged the post, you can escalate the flag to us; this
+    will send an email notifying us. If not, you can ping any of us in Discord, or flag the post yourself and escalate
```
cellio, line 30: As mentioned in another comment, why not ask them to use our reporting tool for consistency and better tracking?

app/views/complaints/training/handling.html.erb
```diff
@@ -0,0 +1,52 @@
+<p class="has-font-size-caption has-color-tertiary-600">Last updated 22 March 2026</p>
+
+<h1 class="is-brand-comp with-subtitle">Online Safety Act</h1>
+<h2 class="subtitle">Handling Illegal Content</h2>
+
+<p>
+  It's important that illegal content is handled consistently in order to ensure that we are able to meet the statutory
+  requirements of us, to protect our users, and to enable us to be transparent with our communities. The way in which
```
cellio, line 8: I would add (after the first sentence) something like "We are not asking our volunteer moderators to make legal determinations; we are asking you to help us meet our legal obligations by making us aware of content you encounter that could violate the law."  I want to reassure them that, amidst all this legal language about so many kinds of illegal content, *they* are not on the hook -- we are.  They are volunteers who are assisting us.

app/views/complaints/training/conclusion.html.erb
```diff
@@ -0,0 +1,17 @@
+<p class="has-font-size-caption has-color-tertiary-600">Last updated 22 March 2026</p>
+
+<h1 class="is-brand-comp with-subtitle">Online Safety Act</h1>
+<h2 class="subtitle">Conclusion</h2>
+
+<p>
+  Thank you for taking the time to go through this training. We appreciate that these are often difficult and sensitive
+  topics, but your understanding and assistance help us to create safe communities for all our users.
```
cellio: Consider adding something like: As moderators you are likely to see problems before we do, and we appreciate your help in bringing them to our attention.

app/views/complaints/training/conclusion.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  Thank you for taking the time to go through this training. We appreciate that these are often difficult and sensitive
+  topics, but your understanding and assistance help us to create safe communities for all our users.
+</p>
+<p>
+  Please click the button below to mark your training complete. You can come back to these pages at any time from the
+  Safety Center if you need a reminder. If you need any support, please contact the Community Team.
+</p>
+
+<%= form_tag osa_training_complete_path, method: :post do %>
+  <%= submit_tag 'Mark as complete', class: 'button is-filled is-primary' %>
```
cellio, line 17: I expected this to show up in the audit log, but after completing training as a mod, I didn't see an entry as an admin.  Where are we recording it?

cellio (commented):
Thanks for putting this together.  I reviewed by going through the training in the UI and leaving comments in the HTML source here, so order of comments in the diff is not the order I left them in (mentioning because you might see some things that came out as forward-references but weren't written that way).

A few small comments on content, and two other issues:

- Should the network-specific parts, like your name, be behind global site settings so other deployments don't pick them up?
- Can we log completion somewhere beyond in the DB?  (I looked in the audit log.)

app/controllers/complaints_controller.rb
```diff
@@ -202,6 +203,25 @@ def reporting
...
+      not_found!
+    end
+  end
+
+  def training_complete
+    if current_user.update(osa_training: DateTime.now)
+      flash[:success] = 'Thank you. Your training has been marked as complete.'
+    else
+      flash[:danger] = 'Something went wrong. Please tell a staff member about this.'
+    end
```
Oaphi: We should make those locale strings while we are at it (non-blocking)

app/views/complaints/training/overview.html.erb
```diff
@@ -0,0 +1,64 @@
...
+  service, which for user-to-user services primarily focus on preventing and removing harmful content, and protecting
+  users from related harms.
+</p>
+
+<h3>Our responsibilities</h3>
+<p>
+  Responsibility for compliance with the requirements of the Act obviously falls on us (meaning the Codidact Foundation
+  as the organisation running the platform). The Foundation designates one of the Board of Directors as a named
+  individual with ultimate responsibility for compliance with the Act, which is currently
+  <a href="https://meta.codidact.com/users/8045">ArtOfCode</a>.
```
Oaphi: This should probably be a configuration option too to avoid having to do the whole pipeline if the nickname / person responsible for compliance / contact URL change (assuming we don't just go for the site setting approach).

app/views/layouts/osa_training.html.erb
```diff
@@ -0,0 +1,60 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+<head>
+  <%= render 'layouts/head' %>
+</head>
+<body class="<%= Rails.env.development? ? 'development' : '' %>"
```
Oaphi, line 6: This should probably be (Rubocop for ERB files is pending till I get enough time to redo the ERB linting PR):
```ruby
<%= 'development' if Rails.env.development %>
```
Also non-blocking, just saves us a linting issue

test/controllers/complaints_controller_test.rb
```diff
@@ -346,6 +346,38 @@ class ComplaintsControllerTest < ActionDispatch::IntegrationTest
...
+    get osa_training_path('home')
+    assert_response(:success)
+  end
+
+  test 'should be able to complete training' do
+    sign_in users(:staff)
+    post osa_training_complete_path
+    assert_response(:found)
+    assert_redirected_to safety_center_path
+    assert_equal 'Thank you. Your training has been marked as complete.', flash[:success]
```
Oaphi: If `training_complete` action's flash messages are made locale strings, this test should be updated too

app/views/complaints/training/overview.html.erb
ArtOfCode-: No, because legal-delete automatically files a report anyway, so no need.

app/views/complaints/training/definitions.html.erb
ArtOfCode-: This is a UK thing. I'll clarify.

app/views/layouts/osa_training.html.erb
ArtOfCode-: This was copied directly from the `without_sidebar` layout, so is presumably in both—I'll leave this until we pick it up through the ERB linting sweep.

app/views/complaints/training/overview.html.erb
```diff
@@ -0,0 +1,64 @@
...
+  service, which for user-to-user services primarily focus on preventing and removing harmful content, and protecting
+  users from related harms.
+</p>
+
+<h3>Our responsibilities</h3>
+<p>
+  Responsibility for compliance with the requirements of the Act obviously falls on us (meaning the Codidact Foundation
+  as the organisation running the platform). The Foundation designates one of the Board of Directors as a named
+  individual with ultimate responsibility for compliance with the Act, which is currently
+  <%= user_link(User.find(SiteSetting['OSADirector'])) %>.
```
Oaphi: That'll result in an `ActiveRecord::RecordNotFound` error for those that don't have the user `8045` if they try to open the page - let's do a `find_by(id: ...)` instead to get `nil` when the user is not found ("which is currently deleted user" is a bit funny, but better than a server error).

app/views/complaints/training/higher-risk.html.erb
```diff
@@ -0,0 +1,100 @@
...
+  <a href="https://meta.codidact.com/posts/288194">fairly clear policy</a> prohibiting sole AI use, this regularly gets
+  missed or ignored, and work into appropriately citing sources used in generative AI models is still very young.
+</p>
+
+<div class="has-text-align-right">
+  <%= link_to osa_training_path('conclusion') do %>
+    <strong>Next</strong><br/>
+    Conclusion &raquo;
+  <% end %>
+</div>
```
Oaphi: Minor note: end of file marker's missing

app/views/complaints/training/handling.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    disclose who has reported or escalated the content report to us.
+  </li>
+</ul>
+
+<div class="has-text-align-right">
+  <%= link_to osa_training_path('higher-risk') do %>
+    <strong>Next</strong><br/>
+    Higher-Risk Content &raquo;
+  <% end %>
+</div>
```
Oaphi: Minor note: end of file marker's missing

db/seeds/site_settings.yml
```diff
@@ -862,3 +862,27 @@
...
+- name: OSATrainingEnabled
+  value: true
+  value_type: boolean
+  category: SafetyCenter
+  community_id: ~
+  description: >
+    Should Online Safety Act moderator training be available on this instance?
+    
+- name: OSADirector
+  value: 8045
```
cellio: Should we default to 1?  In a new deployment, 1 is going to be the admin user who then sets up everything else.  I don't think we want to bake in 8045 (we should instead set that in prod).

db/seeds/site_settings.yml
```diff
@@ -862,3 +862,27 @@
...
+  value_type: integer
+  category: SafetyCenter
+  community_id: ~
+  description: >
+    The user ID of the Director with responsibility for OSA compliance.
+
+- name: ContentPolicyURL
+  value: https://codidact.org/OP02
```
cellio: Similarly, let's not bake this in; it's better to have an obviously-wrong example.com (or whatever) than something that looks legit but doesn't actually apply elsewhere.

app/controllers/complaints_controller.rb
```diff
@@ -257,7 +257,8 @@ def set_complaint
   end
 
   def training_access
-    unless user_signed_in? && (current_user.staff? || current_user.at_least_moderator?)
+    osa_training_enabled = SiteSetting['OSATrainingEnabled']
```
cellio, line 263: Let's also add this check to the main safety page so we don't show a link that 404s.

app/views/complaints/training/handling.html.erb
```diff
@@ -7,7 +7,7 @@
   It's important that illegal content is handled consistently in order to ensure that we are able to meet the statutory
   requirements of us, to protect our users, and to enable us to be transparent with our communities. The way in which
   illegal content is handled is set out in our Online Safety & Illegal Content Policy & Procedure, OP02 &mdash; you can
```
cellio: Small thing: also delete OP02, which is the code for *our* policy but probably no one else's.

db/seeds/site_settings.yml
```diff
@@ -862,3 +862,27 @@
   description: >
     Formatting types allowed in post titles.
     By default, only <code>code</code>, <kbd>keyboard</kbd>, and <em>italic</em> are enabled.
+
+- name: OSATrainingEnabled
+  value: true
```
cellio: Because an instance is going to have to do several other things including creating a policy to link, probably editing the overview seed that mentions the Codidact Foundation, and updating these settings, let's consider making it off by default.  We could expand the description here to say what else they'll need to do, or add a section to the installation doc, or both.

cellio (commented):
Much better, thanks.  I ran through the training again after looking at the diff (this time in a small window to simulate a phone).  The training looks good and I see the entry in the audit log now.

Thanks for adding the site settings.  I left some comments about default values there.

Even with these settings, another network would need to edit some of the seeds -- references to the Codidact Foundation, our AI policy, and Discord jumped out.  Let's not make the perfect the enemy of the good here; instead of trying to configure everything, we can (in a follow-on task) add some documentation to the installation guide and have this be off by default so nobody's surprised by these things in a live instance.  Turning it all on in prod after deploy will not be hard for us, and others will need to do more work before they'll want to turn it on anyway.

Oaphi (approved):
@ArtOfCode- I took the liberty to fix up the controller's tests - other than that, the changes LGTM code review-wise

cellio (approved):

PR #2006, opened 2026-03-04
Fix automatic logout of deleted users

### LINKED ISSUE #591: `You are already signed in` shouldn't show for deleted user
`You are already signed in.` The text should be `Your profile on this community has been deleted.` as written [here](https://github.com/codidact/qpixel/commit/5f5778f8601836d4489a250c1ae0b7a727a6ac74). But, unfortunately some time that context shows. Some time another one shows.
Screenshot from 2021-08-02 02-40-18

Currently, when selecting the button (ask question) it also returns `You are already signed in`

Oaphi:
closes #591

To test the change:
1. Log in as any user (important note: the "Stay signed in on this device" option **must** be checked);
2. While logged in, mark the user as deleted (the easiest way is to set `deleted` to `true` for that user directly in the database);
3. Reload the page - the header should now look as if you were logged out;
4. Click on the "sign in" button;
5. Observe being redirected to the sign in page with the danger notice (incorrectly) saying "You are already signed in";
6. Checkout the PR, reload the page, and click on the "sign in" button again;
7. Observe being redirected to the sign in page with the danger notice (correctly) saying:
    - "Invalid email or password" if the user is marked as deleted network-wide;
    - "Your profile on this community has been deleted" if only the user's community user is marked as deleted;

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,7 +1,5 @@
 # Provides helper methods for use by views under <tt>ApplicationController</tt> (and by extension, every view).
 module ApplicationHelper
-  include Warden::Test::Helpers
-
   ##
   # Is the current user a moderator or admin on the current community?
   # @return [Boolean]
@@ -330,7 +328,7 @@ module ApplicationHelper
     @current_user ||= warden.authenticate(scope: :user)
     if @current_user&.deleted? || @current_user&.community_user&.deleted?
       scope = Devise::Mapping.find_scope!(:user)
-      logout scope
+      warden.logout(scope)
       @current_user = nil
     end
     @current_user


ArtOfCode- (approved):

cellio:
On `develop` and following the setup instructions, I didn't get the notice at step 5 so I couldn't repro the problem before proceeding.  But when I tried to log in it failed, and (as admin) if I try to visit the profile page I get a Rails error because of a data error presumably because this isn't really how to delete users.

The specific change I made in the db was: `update users set deleted=true where id=61;` (where 61 is one of my cannon-fodder users, nothing else special about it).  Is that how I was supposed to do it?

Oaphi:
>  I get a Rails error because of a data error

Just in case we have one more bug on our hands - what's the exact error you get?

cellio (commented):
Couldn't repro yesterday's behavior, weird.  I'm sorry; it was something about a nil or null that seemed like it might have been about a time value (`deleted_at`?), but I didn't record it since I expected to be able to repro on demand, oops.

This behaves correctly for me on the branch.  However, `develop` also behaves correctly for me, so I'm not sure what that means.  If Art is happy then I'm fine with merging.

Oaphi:
> Couldn't repro yesterday's behavior, weird.

My apologies, I forgot to include the part that actually makes the import reproducible - the remember me checkbox **must** be set when logging in. Please do try again - this is a critical path, so I would like to be extra sure the fix works as expected (and that I've correctly identified the root problem).

cellio:
> My apologies, I forgot to include the part that actually makes the import reproducible - the remember me checkbox **must** be set when logging in. 

Ah!  That made the difference -- was able to reproduce the problem on `develop` and then confirmed that it doesn't happen on the branch.

cellio (approved):

PR #2007, opened 2026-03-06
Fix copy code button size when over oneline code blocks

### LINKED ISSUE #2002: Copy code button doesn't align well with single-line code blocks
The button should probably just be a bit smaller in such cases (although determining that a block is a single-line one is not as straightforward as one might think):

Image

Oaphi:
With this PR, multiline blocks will still show the big button as before: 

Screenshot from 2026-03-07 00-51-06

Single-line blocks, however, will show a smaller better-fitting version of the button:

Screenshot from 2026-03-07 00-51-18

closes #2002

### DIFF
--- a/app/assets/javascripts/codeblocks.js
+++ b/app/assets/javascripts/codeblocks.js
@@ -1,29 +1,62 @@
 document.addEventListener('DOMContentLoaded', () => {
-  const buttonTemplate = `<button class="copy-button button is-muted is-outlined has-margin-2">Copy</button>`;
+  /**
+   * @param {string} content
+   * @returns {HTMLButtonElement}
+   */
+  const createCopyButton = (content, isSmall = false) => {
+    const button = document.createElement('button');
+    button.classList.add('copy-button', 'button', 'is-muted', 'is-outlined', 'has-margin-2');
+    button.textContent = 'Copy';
+
+    if (isSmall) {
+      button.classList.add('is-small');
+    }
+
+    button.addEventListener('click', async () => {
+      const originalButtonText = button.textContent;
+
+      try {
+        await navigator.clipboard.writeText(content);
+        button.textContent = 'Copied!';
+      }
+      catch (e) {
+        console.warn(e);
+        button.textContent = 'Failed!';
+      }
+      finally {
+        setTimeout(() => {
+          button.textContent = originalButtonText;
+        }, 2000);
+      }
+    });
+
+    return button;
+  };
+
+  /**
+   * @param {HTMLElement} element
+   * @returns {HTMLDivElement}
+   */
+  const wrapRelative = (element) => {
+    const wrapper = document.createElement('div');
+    wrapper.style.position = 'relative';
+    wrapper.append(element.cloneNode(true));
+    element.replaceWith(wrapper);
+    return wrapper;
+  };
+
+  if (!window.isSecureContext) {
+    return;
+  }
 
   $('.post--content pre > code')
     .parent()
-    .each(function () {
-      const $button = $(buttonTemplate);
+    .each(function (_, element) {
       const $content = $(this).text();
       const numLines = $content.trim().split(/\r?\n/).length;
 
-      if (numLines <= 1) {
-        $button.addClass('is-small');
-      }
-
-      $(this)
-        .wrap('<div style="position:relative;"></div>')
-        .parent()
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/codeblocks.js
```diff
@@ -1,16 +1,29 @@
...
+  $('.post--content pre > code')
     .parent()
-    .each(function() {
-      const content = $(this).text()
+    .each(function () {
+      const $button = $(buttonTemplate);
+      const $content = $(this).text();
+      const numLines = $content.trim().split(/\r?\n/).length
+
+      if (numLines <= 1) {
```
trichoplax: Using `<= 1` suggests we are considering not only the case where `numLines` is `1`, but also the case where `numLines` is `0`. Should that case avoid having the Copy button (of either size)?

This seems a very rare case. The only example I can think of where avoiding having the Copy button might be a problem is Code Golf. In the esolang (esoteric programming language) [Whitespace](https://en.wikipedia.org/wiki/Whitespace_%28programming_language%29) it's possible to have a valid program composed entirely of newlines, tabs, and spaces, which would all be removed by the `.trim()` leaving zero length, but still needing a Copy button.

If we wanted to cover both of these rare cases we could also test the length of the content:
- For zero length content, show no button.
- For zero lines after trimming, but non-zero length content, show the small or large Copy button depending on number of newlines.

Is there a simpler approach where we find the height of the code block without trimming? Since both `\r\n` and `\n` contain one `\n`, could we just count how many times `\n` occurs in the content to determine the number of lines, and choose a button accordingly:
- 0 lines: No button
- 1 line: Small button
- Anything else: Large button

Maybe something like (untested):

```js
numLines = $content.length == 0 ? 0 : $content.match(/\n/g).length + 1
```

app/assets/javascripts/codeblocks.js
```diff
@@ -1,16 +1,29 @@
 $(() => {
-  $(".post--content pre > code")
+  const buttonTemplate = `<button class="copy-button button is-muted is-outlined has-margin-2">Copy</button>`;
+
+  $('.post--content pre > code')
```
trichoplax: Non-blocking: would this be a good opportunity to phase out jQuery for this file?

trichoplax (commented):
I'm away from my development machine today so I haven't tested either the code or my suggestion, so my comments here are just thoughts.

cellio (commented):
Tested and looks fine to me.  Deferring to others for code review.  During testing I wondered if the smaller button would actually be fine everywhere, but I didn't test on a phone and maybe the smaller button is a more difficult target there?

app/assets/javascripts/codeblocks.js
Oaphi: Yeah, let's do that - I'll do it in 2 commits, the first one's done (also ensured we only add the button in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Secure_Contexts) as `navigator.clipboard` is not accessible otherwise)

app/assets/javascripts/codeblocks.js
Oaphi: Yeah, the initial implementation was a quick and dirty solution to the "every code block ends with a newline" problem - I've improved it to not require `trim` to work.

cellio (commented):
Tested and behaves as expected: small button for one-liners, large button for multi-liners (including all blank), no button for `<pre>` block.  LGTM but someone else should review the code.

trichoplax:
> no button for `<pre>` block

Is there a reason to avoid a Copy button on a `<pre>` block? I can imagine cases where it might be useful to be able to copy them, such as text diagrams or puzzles. If we decide to have them that can be a separate issue - just wanted to ask the question.

cellio:
> Is there a reason to avoid a Copy button on a `\<pre\>` block? I can imagine cases where it might be useful to be able to copy them, such as text diagrams or puzzles. If we decide to have them that can be a separate issue - just wanted to ask the question.

We probably should (separately), yeah.  I just meant that these changes are about (explicit) code blocks so I didn't *expect* it to happen for `<pre>`, but I tested to make sure it didn't do anything unfortunate there.

trichoplax:
I've added #2014 separately to avoid distraction here.

ArtOfCode- (approved):

PR #2011, opened 2026-03-11
Make include_tags & exclude_tags params more robust

### LINKED ISSUE #2010: include_tags search filter sometimes fails to be parsed as an array
We've already fixed #1902 that should've been enough to avoid the issue, but it seems like something else can cause the problem too. See this [Discord discussion](https://discord.com/channels/634104110131445811/676843271762214932/1481031373396377743) for additional context.

Oaphi:
closes #2010

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -114,11 +114,11 @@ module SearchHelper
       filter_qualifiers.append({ param: :source, value: params[:source].to_sym })
     end
 
-    if params[:include_tags]&.all? { |id| id.match? valid_value[:integer] }
+    if valid_tags_list?(params[:include_tags])
       filter_qualifiers.append({ param: :include_tags, tag_ids: params[:include_tags] })
     end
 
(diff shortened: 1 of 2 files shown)


cellio (approved):
A valid filter URL is like this: 

`http://localhost:3000/categories/1?sort=&predefined_filter=&min_score=&max_score=&min_answers=&max_answers=&status=any&include_tags%5B%5D=8`

I hand-edited that make `include_tags` have an invalid value, like this:

`http://localhost:3000/categories/1?sort=&predefined_filter=&min_score=&max_score=&min_answers=&max_answers=&status=any&include_tags=%5B8%5D`

On `develop` that causes a server error; on this branch it does not.

I don't know if we care, but with the invalid URL, it renders the category page as expected, but it shows with a filter applied but the tag is missing:

screenshot

If there is no on-site way to create that malformed URL then *I* don't care, but I mention it in case there are other implications.

ArtOfCode- (approved):

PR #2012, opened 2026-03-11
Applying a filter when not using an explicit sort type should preserve the default one

### LINKED ISSUE #2003: Applying a filter in the post list completely resets sort parameter in some cases
When the _default sorting_ option is used (for example, after navigating to a community's homepage), applying any filter leads to it being set to nothing (`sort=`), causing none of the options to appear selected (even though backend sorting seems fine):

Image

Pretty harmless all things considered, but you can't unsee it once you notice

Oaphi:
closes #2003

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -158,10 +158,15 @@ class CategoriesController < ApplicationController
   end
 
   def set_list_posts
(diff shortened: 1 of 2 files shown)


cellio (approved):
Tested, LGTM.

ArtOfCode- (approved):

PR #2013, opened 2026-03-11
Fix undefined secondary sort for user posts breaking pagination

### LINKED ISSUE #2008: A user's post list, when sorting by score, can have duplicate or missing entries.
meta:295628

We use the `will_paginate` gem, which has a longstanding issue when the order is not well-defined (https://github.com/mislav/will_paginate/issues/553) (in this case, when posts have the same score, and we are sorting by score).

I believe it should be mostly a matter of defining an additional order here, for instance by sorting posts with the same score by they're creation time or some other unique value.

https://github.com/codidact/qpixel/blob/4deef91944f95e831eb74596c881e25c93db845d/app/controllers/users_controller.rb#L218-L228

Oaphi:
closes #2008 

This PR adds a _secondary_ order by creation timestamp to user's posts (newest first) - this should be safe enough for 99.99% cases (creation timestamp collision is highly unlikely). In addition, the PR lays the groundwork for a proper action callback by making the logic for setting posts in the `show` and `posts` actions shared (that's as far as I am willing to go in scope for this particular fix).

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -45,12 +45,13 @@ class UsersController < ApplicationController
 
   def show
     @abilities = Ability.on_user(@user)
+    @limit = params[:limit]&.to_i || 15
 
     @posts = set_posts.user_sort({ term: params[:sort], default: :score },
                                  age: :created_at, score: :score)
-                      .first(15)
 
     @total_post_count = @posts.count
+    @posts = @posts.first(@limit)
     render layout: 'without_sidebar'
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -46,19 +46,11 @@ class UsersController < ApplicationController
   def show
     @abilities = Ability.on_user(@user)
 
-    all_posts = if current_user&.privilege?('flag_curate') || @user == current_user
-                  @user.posts
-                else
-                  @user.posts.undeleted
-                end
-                .list_includes
-                .joins(:category)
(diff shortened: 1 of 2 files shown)


cellio:
This changed the number in the "see all" button on the profile from the true number back to 15 (for someone with more than 15, i.e. someone with pagination).

Oaphi:
> This changed the number in the "see all" button on the profile from the true number back to 15 (for someone with more than 15, i.e. someone with pagination).

ah, whoops, a sec

app/controllers/users_controller.rb
```diff
@@ -46,19 +46,11 @@ def index
...
-                .user_sort({ term: params[:sort], default: :score },
-                           age: :created_at, score: :score)
-
-    @posts = all_posts.first(15)
-    @total_post_count = all_posts.count
+    @posts = set_posts.user_sort({ term: params[:sort], default: :score },
+                                 age: :created_at, score: :score)
+                      .first(15)
+
+    @total_post_count = @posts.count
```
cellio, line 53: Ah, here's the bug I think -- the previous line limits to 15 and then we count that...

Oaphi: Yeah, glanced over that part - fixed & added a test that should prevent future changes from introducing a regression there.

cellio (approved):
LGTM now; tested and it's behaving correctly.  Thanks for adding the test.

ArtOfCode- (approved):

PR #2015, opened 2026-03-11
Order tag search matches by their length

### LINKED ISSUE #2009: Make it easier to choose the existing tag you meant rather than the fuzzy match you didn't
https://software.codidact.com/posts/295664

In the tags portion of the post editor, when you start to type a tag name you get a list of suggested completions.  If instead of choosing one of those you just hit Return, it uses the first one in the list.  However, the first one in the list is not necessarily the best one.  If you type `c` and there is a tag `c`, it should use that, not (say) `css`.

I can think of two ways to approach this.  The possibly easier one is, on Return, to check if the string you've typed so far is an existing tag and if so use it.  A better solution would be for that tag to show up first in the suggestions list -- prefer a complete match (even on a one-character tag) over the best fuzzy match.  I assume this is primarily a problem for very short tag names.

If you hit return and what you have typed so far is the name of an existing tag, use it.

Oaphi:
closes #2009 

With the PR applied, matching tags should prefer full or close to full matches over longer partial ones:

Screenshot from 2026-03-12 02-27-33

credit to @MoshiKoi for suggesting the solution in our [Discord discussion](<https://discord.com/channels/634104110131445811/676843271762214932/1481385629467148478>)

As opposed to the current sorting that only sorts tag names alphabetically:

image

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -49,7 +49,7 @@ class Tag < ApplicationRecord
 
     # Select from the union of the above queries, select only the tag columns such that we can distinct them
     from(Arel.sql("((#{q1.to_sql}) UNION (#{q2.to_sql})) tags"))
-      .order(Arel.sql(sanitize_sql_array(['name LIKE ? DESC, name', value])))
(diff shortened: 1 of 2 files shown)


cellio (approved):
Tested & LGTM.

PR #2016, opened 2026-03-12
Hotfix: Preload community and user for error report

ArtOfCode-:
Admin error reports pages are currently taking ~20s to load on prod because we're hitting an N+1 query where we don't preload `community`. I've added this preload (and one for `user` because we also use that, although it doesn't seem to be causing an issue).


Oaphi:
@ArtOfCode- I've fixed up the runtime error caused by preloading communities before error types are collected.
Other than that, LGTM, the N+1 query is gone with the PR applied.

PR #2022, opened 2026-03-17
fixed at-mention suggestions at line start in multiline comments

### LINKED ISSUE #2021: Mention completions do not appear if typed at the beginning of a line that isn't the first line
Mention completions do not appear if typed at the beginning of a line that isn't the first line

For example, this doesn't give a popup for my username (Moshi on Codidact)

```
some text
@mo
```

These do

```
@mo
```

```
some text
some more text @mo
```

Oaphi:
closes #2021

Screenshot from 2026-03-17 17-57-07

To reproduce on `develop`:

1. Start a new comment to an existing comment thread.
2. Type in a partial @-mention at the start of the first line - note the suggestions popup appearing.
3. Add a new line and repeat step 2 - note the suggestions popup **not** appearing at all.

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -387,7 +387,7 @@ $(() => {
 
     const $tgt = $(ev.target);
     const content = $tgt.val();
-    const splat = content.split(' ');
+    const splat = content.split(/\s+/);
     const caretPos = $tgt[0].selectionStart;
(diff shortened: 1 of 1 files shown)


cellio (approved):
Tested with both inline threads and thread pages.  Works now.

ArtOfCode- (approved):

PR #2024, opened 2026-03-17
Fix pressing "Enter" in the "add link" modal submitting the parent form

### LINKED ISSUE #2023: Pressing Enter when trying to add a link to a post submits it instead
meta:295748

Oaphi:
closes #2023

This is a "quick and dirty" version of the fix as I don't have full access to my usual tools at the moment (a proper fix would be moving the modal outside the `markdown_tools` partial as `form` tags can't be nested, wrapping the inputs in a `<form>` tag, and cancelling [`preventDefault`] any `submit` event on it - see the `image_upload` partial for how it should be done). Nevertheless, it should do the trick.

### DIFF
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -29,10 +29,9 @@ $(() => {
     }
   });
 
-  $('#markdown-link-name, #markdown-link-url').on('keydown', (ev) => {
-    if (ev.keyCode === 13) {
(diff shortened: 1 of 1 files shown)


cellio (approved):
Works now -- pressing enter in either field (anchor text or link) does nothing, so you have to click the "insert" button to return to the editor.  No premature post submissions.

ArtOfCode- (approved):

PR #2026, opened 2026-03-18
prevent nuking a tag that has children

### LINKED ISSUE #2025: Deleting a tag that has children causes a DB error
If you try to delete a tag that has children, you might expect that to work, with the children being reparented up the chain.  Or you might expect some sort of "what do you want to do about the children?" UI.  What you actually get, though, is a server error, which turns out to be a DB validation error:

```
ActiveRecord::InvalidForeignKey (Mysql2::Error: Cannot delete or update a parent row: a foreign key constraint fails (`qpixel_dev`.`tags`, CONSTRAINT `fk_rails_aff1790b59` FOREIGN KEY (`parent_id`) REFERENCES `tags` (`id`))):

Causes:
Mysql2::Error (Cannot delete or update a parent row: a foreign key constraint fails (`qpixel_dev`.`tags`, CONSTRAINT `fk_rails_aff1790b59` FOREIGN KEY (`parent_id`) REFERENCES `tags` (`id`)))
```

cellio:
If a tag has children, replace the nuke page with its warnings and "are you sure?" button with a message saying you need to handle the children first.  I opted for a simple message rather than disabling the button because with that whole page full of red boxes, doing that felt like burying the lede.

While it shouldn't be possible to reach the `nuke` method in the controller, I also added a check there just in case.

Minimally fixes https://github.com/codidact/qpixel/issues/2025.  A more advanced fix would be to do something reasonable with the children, but that requires some design discussion as I'm not sure we want to simply reparent them.  Maybe we want a human to decide.  In the meantime, this at least prevents the server error.

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -228,7 +228,6 @@ class TagsController < ApplicationController
     if @tag.children.any?
       flash[:error] = 'Cannot delete a tag that has children.'
     else
-
       Post.transaction do
         AuditLog.admin_audit event_type: 'tag_nuke', related: @tag, user: current_user,
                              comment: "#{@tag.name} (#{@tag.id})"
@@ -253,8 +252,8 @@ class TagsController < ApplicationController
       end
 
       flash[:success] = "Deleted #{@tag.name}"
-      redirect_to category_tags_path(@category)
     end
+    redirect_to category_tags_path(@category)
   end
 
   def nuke_warning; end
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -226,34 +226,34 @@ class TagsController < ApplicationController
 
   def nuke
     if @tag.children.any?
-      flash[:error] = "Cannot delete a tag that has children."
+      flash[:error] = 'Cannot delete a tag that has children.'
     else
 
-    Post.transaction do
-      AuditLog.admin_audit event_type: 'tag_nuke', related: @tag, user: current_user,
-                           comment: "#{@tag.name} (#{@tag.id})"
+      Post.transaction do
+        AuditLog.admin_audit event_type: 'tag_nuke', related: @tag, user: current_user,
+                             comment: "#{@tag.name} (#{@tag.id})"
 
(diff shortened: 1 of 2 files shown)


cellio:
I see the codecov complaint.  I don't know what it means in the context of this PR or what I should do about it.

app/controllers/tags_controller.rb
```diff
@@ -225,31 +225,36 @@ def merge
...
+        end
+
+        # Nuke it
+        @tag.destroy
       end
 
-      # Nuke it
-      @tag.destroy
+      flash[:success] = "Deleted #{@tag.name}"
+      redirect_to category_tags_path(@category)
```
ArtOfCode-: Unless an explicit `render` or `redirect` call is made, Rails' default behaviour is to render a view with the same name as the action (i.e. `app/views/tags/nuke.html.erb`). Since you've moved this `redirect` call into the `else` path, the `if` path for when a tag has children doesn't render or redirect, so this is what it will try to do. As that view doesn't exist, it will 404.

Easiest fix is to move the `redirect` call back outside the `if`/`else` block, leaving the `flash` inside it.

app/views/tags/nuke_warning.html.erb
```diff
@@ -6,6 +6,10 @@
 
 <h1>Delete tag <%= @tag.name %></h1>
 
+<% if @tag.children.any? %>
+<p><strong>Tags with children cannot be deleted. Please reparent the children first.</strong></p>
```
ArtOfCode-: I'd suggest a warning notice here rather than a plain paragraph - take an example from the `else` block:


```suggestion
<div class="notice is-warning has-color-yellow-900">
  <p>
    <strong>Tags with children cannot be deleted. Please reparent the children first.</strong>
  </p>
</div>
```

ArtOfCode- (changes requested):

app/views/tags/nuke_warning.html.erb
cellio: Much better, thanks.

ArtOfCode- (approved):

PR #2030, opened 2026-03-20
Make user profile query more efficiently

ArtOfCode-:
Strips the count query back without all the joins. Also adds a NOT NULL to `categories.min_view_trust_level` because the way we were querying that was inefficient, and adds an index for it.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -50,7 +50,11 @@ class UsersController < ApplicationController
     @posts = set_posts.user_sort({ term: params[:sort], default: :score },
                                  age: :created_at, score: :score)
 
-    @total_post_count = @posts.count
+    @total_post_count = if @user == current_user
+                          @user.posts.undeleted.count
+                        else
+                          @user.posts.count
+                        end
     @posts = @posts.first(@limit)
     render layout: 'without_sidebar'
   end
@@ -651,14 +655,16 @@ class UsersController < ApplicationController
   end
 
   def set_posts
-    @posts = if current_user&.privilege?('flag_curate') || @user == current_user
-               @user.posts
-             else
-               @user.posts.undeleted
-             end
-             .list_includes
-             .joins(:category)
-             .where('IFNULL(categories.min_view_trust_level, 0) <= ?', current_user&.trust_level || 0)
+    Rack::MiniProfiler.step('set_posts') do
+      @posts = if current_user&.privilege?('flag_curate') || @user == current_user
+                 @user.posts
+               else
+                 @user.posts.undeleted
+               end
+               .list_includes
+               .joins(:category)
+               .where('categories.min_view_trust_level <= ?', current_user&.trust_level || 0)
+    end
   end
 
   def set_user
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -50,7 +50,15 @@ class UsersController < ApplicationController
     @posts = set_posts.user_sort({ term: params[:sort], default: :score },
                                  age: :created_at, score: :score)
 
-    @total_post_count = Post.all.by(@user).count
+    @total_post_count = if @user == current_user || current_user&.at_least_moderator?
+                          Post.all.by(@user).count
+                        else
+                          Post.all.by(@user)
+                              .undeleted
+                              .joins(:category)
+                              .where('categories.min_view_trust_level <= ?', current_user&.trust_level || 0)
+                              .count
+                        end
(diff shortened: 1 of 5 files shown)


app/controllers/users_controller.rb
```diff
@@ -50,7 +50,7 @@ def show
     @posts = set_posts.user_sort({ term: params[:sort], default: :score },
                                  age: :created_at, score: :score)
 
-    @total_post_count = @posts.count
+    @total_post_count = Post.all.by(@user).count
```
Oaphi: I believe this will expose posts that are in categories inaccessible for the current user - IIRC, this is why the same query was used to both get the 15 posts and the count. While not the end of the world, I'd prefer us to have a slower query in this case (that said, I don't think `.join(:category).where(...)` is the cause for the slowdown - most likely it's the `list_includes` part of it)

Oaphi (changes requested):

app/controllers/users_controller.rb
ArtOfCode-: Joining categories would be quicker than the full `list_includes`, but slower than not joining at all - albeit faster now because of the changes on `min_view_trust_level`. This doesn't reveal post _content_, just a count (i.e. "post(s) exist") - are we okay with revealing that there are posts the current user can't see? cc @cellio

cellio: I'd actually prefer that we find a way for users to always have access to their posts even if in now-restricted categories.  We show them their own *deleted* posts; from a user perspective it makes sense to also let them see their *live but restricted* posts.  However, I recognize that this is a larger issue that shouldn't block this work.  In the meantime I'm ok with revealing the existence of these posts.

trichoplax: How important is it for the "See all N" on the profile page to have the same N as the "N posts" that shows at the top of the user's paginated posts page? Personally I would think something was broken if the number of posts in my list didn't match both of those numbers.

cellio: Oh, I didn't realize there'd be a mismatch.  Yes, the two numbers should match, either including or excluding those hidden posts in both places.

trichoplax: That was in response to:

> are we okay with revealing that there are posts the current user can't see?

which made it sound like the total might include more posts than are displayed.

cellio: Right.  So if we report a number that is higher than the actual number of posts on the paginated pages, that's a problem but needn't be a blocking problem here since that's a rare case and we can fix it separately.  But if the number is *different* on the main-page button and in the text count on the pagination page, that'd be more of a problem.

ArtOfCode-: I'm not at my PC - does anyone have test data with posts in private categories that they can test this with to see if the two are different? If not I'll test later.

cellio: In prod, I just checked my Judaism profile.  Logged in I see 173 posts; incognito I see 165.  The difference is six deleted posts and two posts in Purim Torah, a restricted category.  So currently we're excluding those posts in the list and in the count, unless you can see the category.  

I've just set up a test case in my dev environment to test with -- results are the same as for prod, as expected.

trichoplax: As a non-admin user, I see "See all 165" on your Judaism profile page, and "165 posts" on your posts page, and the list of posts contains 165 posts (6 pages of 25 plus 15 on the 7th page).

app/controllers/users_controller.rb
```diff
@@ -50,7 +50,15 @@ def show
...
-    @total_post_count = @posts.count
+    @total_post_count = if @user == current_user || current_user&.at_least_moderator?
+                          Post.all.by(@user).count
+                        else
+                          Post.all.by(@user)
+                              .undeleted
+                              .joins(:category)
+                              .where('categories.min_view_trust_level <= ?', current_user&.trust_level || 0)
+                              .count
+                        end
```
Oaphi, line 61: I believe we can move `count` to after `if` as both branches call `.count`:

```ruby
    @total_post_count = if @user == current_user || current_user&.at_least_moderator?
                          Post.all.by(@user)
                        else
                          Post.all.by(@user)
                              .undeleted
                              .joins(:category)
                              .where('categories.min_view_trust_level <= ?', current_user&.trust_level || 0)
                        end.count
```
Certainly non-blocking and, frankly, I am not sure if I like the current convention of chaining calls like this

Oaphi (commented):
I left a small note, but other than that the current version LGTM

Oaphi (approved):

PR #2031, opened 2026-03-20
Allow ability descriptions to be edited

Oaphi:
Only ability descriptions are editable for now (the user has to at least be a moderator for the button to show up):

https://github.com/user-attachments/assets/219b5d72-8825-4b17-802f-c447c407c80e

Global mods & admins will also be able to push the changes network-wide:
image


ArtOfCode- (changes requested):
Two semi-related requests:
- [x] I think it's fine to allow moderators to edit these - gives communities some agency and reduces the admin burden. If we're doing that -
- [x] while we're here anyway, can we add a checkbox for global mods/admins to push the edit to network, which will update matching abilities where the description has not locally diverged? It's fine for that check to just be "body matches previous state" as there's no history for abilities.

db/schema.rb
```diff
@@ -115,6 +115,7 @@
     t.index ["community_id"], name: "index_categories_on_community_id"
     t.index ["default_filter_id"], name: "index_categories_on_default_filter_id"
     t.index ["license_id"], name: "index_categories_on_license_id"
+    t.index ["min_view_trust_level"], name: "index_categories_on_min_view_trust_level"
```
Oaphi, line 118: Apparently, we forgot to update the schema in the recent updates - adding a comment here to avoid confusion

ArtOfCode- (approved):

PR #2033, opened 2026-03-25
Fix category change select dropdown

### LINKED ISSUE #2032: Change Category does not show list of categories (preventing changes)
Post -> Tools -> Change Category shows a selector for the category, but no categories.  Neither clicking in it nor typing does anything.  I reproduced this in my dev environment, so it's not a Cloudflare problem in fetching the list.

It feels like the problem could be in this line of code (in `_post_tools.html.erb`), but I'm a little lost here:

```
<%= select_tag :target_id, options_for_select([]), include_blank: true,
         class: 'form-element js-category-select', id: "target_id_#{post.id}" %>
```

`options_for_select` with an empty array looks suspicious to me, but I've been surprised by Rails magic before as I am *far* from an expert.  So maybe this is a red herring, but this is as far as I was able to get on my own before filing this bug.

This tool has worked in the past, so this is a regression.  I don't know how long ago it broke; the last logged category change in prod was 2025-06-25.

Oaphi:
closes #2032

image

The PR also actually enables category search by name (that it doesn't work is, ironically, only noticeable when trying to narrow down a category change dropdown's options):

image

### DIFF
--- a/app/assets/javascripts/categories.js
+++ b/app/assets/javascripts/categories.js
@@ -45,7 +45,10 @@ $(() => {
 
   $('.js-category-select').each((_i, el) => {
     const $tgt = $(el);
+    /** @type {HTMLElement|undefined} */
+    const modal = el.closest('.modal--container') || void 0;
     $tgt.select2({
+      dropdownParent: modal,
       ajax: {
         url: '/categories',
         headers: { 'Accept': 'application/json' },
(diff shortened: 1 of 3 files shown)


Oaphi:
Let's make another small deploy on merge (no need to wait for me to do the latter either if everything looks good)

cellio (approved):
Tested and LGTM.  Code looks fine with my usual caveat about JS.

ArtOfCode- (approved):

PR #2035, opened 2026-04-01
fixed setting SiteLogoPath to a missing asset causing server errors

### LINKED ISSUE #2034: Community advertisements should gracefully handle incorrect SiteLogoPath
Currently, if `SiteLogoPath` is set to a non-existing asset, attempts to generate an advert result in an `ImageMagickError`, specifically:
```
unable to open image `<image path here>`: No such file or directory
```
There's no actual requirement for communities to have logos either, so let's handle this gracefully instead of a server error.

Oaphi:
closes #2034

To reproduce on develop, set the `SiteLogoPath` site setting to a path starting with `/assets/` and pointing to a non-existent asset (any gibberish should be fine). Observe a server error upon visiting, for example, `<your origin here>/ca/posts/random.png` (will attempt to generate a community ad for a random post). With the PR applied doing the latter should result in a valid image response with the community name instead of the logo.

Example community ad with a valid logo path:

And the same ad with an invalid logo path:

### DIFF
--- a/app/helpers/advertisement_helper.rb
+++ b/app/helpers/advertisement_helper.rb
@@ -78,18 +78,25 @@ module AdvertisementHelper
   # images.
   # @param icon_path [String] A path or URI from which to load the image. If using a path this should be the asset path
   #   as it would be accessible through the Rails server - see example.
-  # @return [Magick::ImageList] An ImageList containing the icon.
+  # @return [Magick::ImageList, nil] An ImageList containing the icon if found
   # @example Load an image from app/assets/images:
   #   # This uses the path from which the image would be accessed via HTTP.
   #   helpers.community_icon('/assets/codidact.png')
   def community_icon(icon_path)
-    if icon_path.start_with? '/assets/'
-      icon = Magick::ImageList.new("./app/assets/images/#{File.basename(icon_path)}")
+    return nil unless icon_path.present?
+
+    expanded_path = File.expand_path(icon_path)
+
+    if expanded_path.start_with?('/assets/')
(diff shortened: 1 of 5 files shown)


Oaphi:
Lacks tests and ~~needs #2004 to be merged for the TypeScript check to be fixed~~, so I am keeping it as a draft for the time being

cellio:
I reproduced the error on `develop`, switched to this branch and got the behavior described -- and then switched the logo *back* to what it had been, but I still get "dev community" instead of the logo in the ad.  Does that happen for you too?

Oaphi:
> and then switched the logo back to what it had been, but I still get "dev community" instead of the logo in the ad

Have you cleared the cache with `Rails.cache.clear` afterwards?

cellio:
> Have you cleared the cache with `Rails.cache.clear` afterwards?

Ah.  No I hadn't realized it was cached, and yes that fixed the issue.

cellio (approved):
Tested and LGTM.  Someone more fluent in magick and our caching should review the code.

Oaphi:
> Ah. No I hadn't realized it was cached, and yes that fixed the issue.

Yea, ads are cached the moment they are created, so you have to reset either the whole cache or the corresponding key every time you change a dependency such as a site setting (we should probably add some plumbing to do expire the cache automatically in such cases)

cellio:
> (we should probably add some plumbing to do expire the cache automatically in such cases)

I would expect setting a new value to flush the cache entry, yeah.  That's something we should look into, separately.

ArtOfCode- (approved):

PR #2037, opened 2026-04-02
Fix comment thread expansion if its title contains clickable elements

### LINKED ISSUE #2036: Comment thread inline expansion breaks if the title contains clickable content such as MathJax
https://math.codidact.com/posts/295860

Reproduction steps and the reason why it happens can be found in the linked post.

There's a bonus issue that MathJax content is stripped out of the title upon collapsing the thread again

Oaphi:
closes #2036 

Also makes sure we do not replace the inline thread with the fetched content unless the response indicates success (most likely solves meta:295715, but let's make sure every case is covered in a separate PR).

As a treat, it also ensures MathJax stays properly rendered upon collapsing previously opened comment threads.

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -35,6 +35,11 @@ $(() => {
   async function openThread(wrapper, threadId, options) {
     const data = await QPixel.getThreadContent(threadId, options);
 
+    if (!data) {
+      QPixel.createNotification('danger', 'Failed to open thread');
+      return;
+    }
+
     wrapper.innerHTML = data;
 
     window.MathJax && MathJax.typeset();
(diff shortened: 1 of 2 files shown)


Oaphi:
Let's do a production deployment upon review & merge

cellio (approved):
Tested and works for me.

Before the fix, in a dev environment I got what looks like a stack trace rather than a 404, so I'm not sure how to test the "maybe also fixes" issue.  But we can check that case after deploying this and see if it's still an issue.

Oaphi:
> Before the fix, in a dev environment I got what looks like a stack trace rather than a 404

That's the exact same error, @cellio, it's just that in development mode the `ErrorsController` is not invoked by default (for the dev to be able to see the trace and other info). In production, the same error results in the 404 error page being served.

ArtOfCode- (approved):

PR #2038, opened 2026-04-17
Update bundle & Ruby version

ArtOfCode-:
Update the bundle to address some of the outstanding dependency alerts. Drops support for ruby 3.1 and 3.2, adds 3.3 and 3.4. 3.3 is in security-maintenance only, so we'll need to look at 4.0 fairly soon.


ArtOfCode-:
Sorry, the diff on this is horrific because Rubocop complained about line endings. The important changes are in the Gemfile and lock.

cellio:
Will anything here affect the Docker setup?  (I don't know enough to review this, so leaving a question for reviewers who do.)

ArtOfCode-:
Can't say I know for sure - I don't use it.

PR #2044, opened 2026-04-25
Fix user list avatar size at 48 by 48 pixels

### LINKED ISSUE #462: Size of user images varies on users list

trichoplax:
Fixes #462 by preventing avatars from shrinking when the text of a user card gets too wide. This forces text wrapping to happen a few characters sooner, but keeps all of the avatars the same size.

## Electrical Engineering user list currently
User card in 3 columns, with the avatars at various sizes

## Electrical Engineering user list with this fix
User cards in 3 columns, with the avatars all with maximum dimension 48

Square avatars are now always displayed at 48 by 48 pixels. Portrait avatars have height 48 pixels. Landscape avatars have width 48 pixels.

### DIFF
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -41,10 +41,10 @@ img {
   }
 
   img {
-    min-height: 32px;
-    min-width: 32px;
-    width: 32px;
-    height: 32px;
+    min-height: 48px;
+    min-width: 48px;
+    width: 48px;
+    height: 48px;
     object-fit: contain;
   }
 


cellio (approved):
Tested; LGTM.

PR #1950, opened 2026-01-20
Data dumps for QPixel

Oaphi:
closes #1918

PR #2047, opened 2026-05-12
Add data dumps

### LINKED ISSUE #1918: Add a system for regular and/or on-demand data dumps
meta:295094
meta:293597

Adding a tracking issue since it's been requested a couple of times and is a general expectation of our community members. It won't be an easy feat to implement properly, but we shouldn't lose sight of the request.

Related: meta:285446

ArtOfCode-:
Add data dumps. Weekly export of the entire database minus anything sensitive, uploaded to S3 and made available via a new page linked in the footer. Also adds an option for manually-created dump records, intended for quarterly uploads to Archive.org.

Incorporates #1950 by cherry-pick.


cellio:
After switching to the branch and running the migrations, and before running any job to generate dumps, I hit the "data dumps" link in the footer and got:

```
ActionView::Template::Error (undefined method `title' for nil):

Causes:
NoMethodError (undefined method `title' for nil)
    29:   </thead>
    30:   <tbody>
    31:     <tr>
    32:       <td><%= @latest.title %></td>
    33:       <td><%= @latest.file.filename %></td>
    34:       <td><%= @latest.created_at.strftime('%Y-%m-%d') %></td>
    35:       <td><%= link_to 'Download', rails_blob_path(@latest.file, disposition: 'attachment') %></td>
  
app/views/dumps/index.html.erb:32
```

Let's defend against that; even if we intend to generate a dump immediately after deploying, (a) there might be a delay and (b) we can't assume other instances will do so.

cellio:
Running the job from the Rails console produced an access error -- looks like we're missing a grant for `perform`.

```
:001> DataDumpJob.perform_later
Performing DataDumpJob (Job ID: 173732ce-812a-4dae-bd88-b6e324228c46) from Inline(default) enqueued at 2026-05-15T03:01:18.998364424Z
Error performing DataDumpJob (Job ID: 173732ce-812a-4dae-bd88-b6e324228c46) from Inline(default) in 17.49ms: ActiveRecord::ConnectionFailed (Mysql2::Error::ConnectionError: Access denied for user 'qpixel'@'%' to database 'qpixel_dump'):
```
(Stack trace follows; I assume you don't need all that.)

ArtOfCode-:
> Running the job from the Rails console produced an access error -- looks like we're missing a grant for perform.

I deliberately didn't include this as part of the job, as the DB user running the export shouldn't have `GRANT` permissions. I did it manually - I'll add some error handling for that specifically that puts a warning in the job log.

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,463 @@
...
+    - community_id
+    - user_id
+    - is_moderator
+    - is_admin
+    - reputation
+    - created_at
+    - updated_at
+    - trust_level
+    - deleted
+    - deleted_at
```
cellio: How about omitting `deleted_at` and `deleted_by_id`?  We're not dumping deleted profiles so we don't expect including the columns here to be a problem, *but* if a profile was deleted and then undeleted, and if that process leaves any residue in the data (by intention or by bug), that'd leak stuff we don't want to leak.

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,463 @@
...
+    - deleted_by_id
+    - post_count
+  query: >
+    WHERE deleted = FALSE
+
+filters:
+  columns:
+    - id
+    - name
```
cellio: This includes every user-written filter, not just the system ones.  Users' filters aren't public, but merely stripping the user ID here doesn't seem like enough.  Maybe we want a WHERE for system filters only?

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,463 @@
...
+    - min_answers
+    - max_answers
+    - status
+    - include_tags
+    - exclude_tags
+    - created_at
+    - updated_at
+    - source
+
+flags:
```
cellio, line 170: I don't think we want to dump data about individual flags.  "Confidential" means mod-only as opposed to also curators; even "non-confidential" flags are not public.

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,463 @@
...
+reactions:
+  columns:
+    - id
+    - user_id
+    - reaction_type_id
+    - post_id
+    - comment_id
+    - created_at
+    - updated_at
+
```
cellio, line 356: Should we exclude reactions on deleted posts?

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,463 @@
...
+    - ever_archived_before
+    - deleted
+    - deleted_by_id
+    - created_at
+    - updated_at
+    - community_id
+    - locked_at
+    - last_activity_at
+  query: >
+    WHERE deleted = FALSE
```
cellio: Should we also exclude if the post the thread is attached to is deleted?

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,463 @@
...
+    - deleted
+    - user_id
+    - community_id
+    - comment_thread_id
+    - has_reference
+    - reference_text
+    - references_comment_id
+    - last_activity_at
+  query: >
+    WHERE deleted = FALSE
```
cellio: Similar to the previous, should we also exclude if the thread is deleted or if the thread's post is deleted?  (We've nuked some pretty rude comment fights by just nuking the thread...)

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,463 @@
...
+    - parent_id
+
+user_abilities:
+  columns:
+    - id
+    - community_user_id
+    - ability_id
+    - is_suspended
+    - suspension_end
+    - suspension_message
```
cellio: I know it says `WHERE is_suspended = FALSE`, but I'm still twitchy about including the message here.

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,463 @@
...
+    - staff
+    - trust_level
+  query: >
+    WHERE deleted = FALSE
+
+votes:
+  columns:
+    - id
+    - vote_type
+    - created_at
```
cellio: Observation: by including created and updated timestamps, we give the dump more data than is available through the UI, where people can see the aggregate votes but not the timing.  I don't have an opinion on whether we *should*; I just noticed it.

cellio (commented):
I tested the dump process and the page for viewing them and all looks good there.  I focused my review on `dump_permitted_columns.yml` and have some comments there.  

I also looked at a dump.  DDL for all tables (even non-permitted ones) is there, which is fine -- the schema is public, after all.  In cases where a table does not appear in the YAML file, the dump has the DDL but no data -- good.  In cases where a table is included but some columns are not, those values are NULL in the dump.  That all makes sense to me, and I assume the NULLs won't confuse anyone who would be looking at dumps in the first place.

db/scripts/dump_permitted_columns.yml
ArtOfCode-: I debated this as I wrote it, but I wonder if there might be some interest in looking into how filters are being used. They don't contain anything identifying, and the only free-text field is `name`, which I don't think poses any significant privacy concerns. Thoughts?

db/scripts/dump_permitted_columns.yml
ArtOfCode-: Fair point, although non-confidential flags _are_ shown to curators so not quite like some "never shown to users" data. However, there might also be some legitimate interest (as for filters) in querying into flags. What if we dropped `post_id` as well, so you can see _how_ flags are being used without seeing _what's_ being flagged?

db/scripts/dump_permitted_columns.yml
trichoplax: My preference is that things that are public are obviously so. That is, either part of something that will be deliberately published (like a post or comment), or labelled as public if that might be surprising to some.

As an example, for a long time favourited posts on SE were public, while giving no indication of being so. Unless they knew where to look, it was easy for users to assume their favourites were private, and they had no reason to suspect otherwise so no reason to go looking. In the majority of cases there was no reason to need favourites to be private, but for some users knowing they were public would have affected their decision on whether to favourite a post about a medical condition or political opinion.

I feel this case is similar. Most people will have no reason to care whether the `name` field is seen by others, but the expectation of privacy means some might word things differently than if they expected it to be public. This could be partially mitigated by not including the user who made the filter, but this would still leave some problems. For example, someone could create a filter named "Block user7's irritating posts". Since we can't currently filter users in or out, this would be an inaccurate name (perhaps using a combination of tags and score, but named based on the underlying motivation). Even if the user who created the filter is not divulged, this would still be embarrassing for user7.

I would expect this and worse cases to be very rare, but because the potential is there my preference would be for the name of the filter to be omitted (at least for user created filters) unless there is a clear indication for people who don't use the data dump that the name they give their filter will not be private.

db/scripts/dump_permitted_columns.yml
trichoplax: This makes me wonder whether there could be additional data dumps hidden from users who are not signed in. For example, a curator data dump that includes flags that curators can see, and a moderator data dump that includes flags that only moderators can see, with no flag information showing in the public data dump.

This would only be sharing information that the user already has access to, but it would make accidental or deliberate sharing of that data much easier, so I don't know if it should be done.

ArtOfCode-: That sounds like additional complexity for not very much gain - I think we could do without managing the complexity of multiple different data dumps.

db/scripts/dump_permitted_columns.yml
ArtOfCode-: @trichoplax Helpful to hear your thoughts, thank you. I'll take `name` out as well.

trichoplax (commented):
There are a few places where data is excluded if it is attached to a deleted post. As a potential edge case: I vaguely recall it being rare but possible for an answer to a deleted question to end up being not deleted. Although I don't think the user interface will display such an answer to users who can't view deleted posts (since you can't view its deleted parent question), will comments/reactions attached to such an answer still be included in the data dump? Do we need to also exclude items attached to an answer with a deleted parent question?

db/scripts/dump_permitted_columns.yml
cellio: Dropping `post_id` so people can only see how flags are happening but not on what works for me.

cellio:
> There are a few places where data is excluded if it is attached to a deleted post. As a potential edge case: I vaguely recall it being rare but possible for an answer to a deleted question to end up being not deleted. Although I don't think the user interface will display such an answer to users who can't view deleted posts (since you can't view its deleted parent question), will comments/reactions attached to such an answer still be included in the data dump? Do we need to also exclude items attached to an answer with a deleted parent question?

This was caused by data imports.  (Not always; I don't know if we ever characterized the bug.)  We could defend against it here, or add something that fixes the data problems so we don't need to.  (If we take the latter approach, let's add a task here, not just fix it in prod; other instances might have done imports.)

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,461 @@
...
+    - reputation
+    - created_at
+    - updated_at
+    - trust_level
+    - deleted
+    - post_count
+  query: >
+    WHERE deleted = FALSE
+
+filters:
```
cellio, line 156: All filters are now showing a user ID of -1 (System) because that's the default, but that's incorrect for user-written filters.  Is there something we can do in the dump to handle this, without having to go rework the schema and filter code?  It's ok (even nice) to distinguish system and user-written filters in the dump, without revealing any more about the latter.  If we're stuck with treating them all as system even when they're not, we can live with that, but can we do better?

db/scripts/dump_permitted_columns.yml
```diff
@@ -0,0 +1,461 @@
...
+    - id
+    - label
+    - url
+    - position
+    - user_id
+
+users:
+  columns:
+    - id
+    - created_at
```
cellio: I just realized that this exposes identifiable user creation times and update times, which we chose not to expose in the UI (only mods and owners see it).  It's probably safer to exclude the columns here; *most* users wouldn't care, but if even one would care that we've exposed private info, that'd look bad.

Ditto for `community_users`.

trichoplax:
> > There are a few places where data is excluded if it is attached to a deleted post. As a potential edge case: I vaguely recall it being rare but possible for an answer to a deleted question to end up being not deleted. Although I don't think the user interface will display such an answer to users who can't view deleted posts (since you can't view its deleted parent question), will comments/reactions attached to such an answer still be included in the data dump? Do we need to also exclude items attached to an answer with a deleted parent question?

> This was caused by data imports. (Not always; I don't know if we ever characterized the bug.) We could defend against it here, or add something that fixes the data problems so we don't need to. (If we take the latter approach, let's add a task here, not just fix it in prod; other instances might have done imports.)

I don't recall any details now, but I thought there might have been a case that resulted from timings (maybe a user pressing save on an answer after the question was deleted?), so unless we can rule that out we might get very rare new cases in future.

If that isn't already ruled out, would a potential solution be to have a job to delete such answers, and have it run just before the data dump each week?

db/scripts/dump_permitted_columns.yml
ArtOfCode-: Not realistically without changing how the data dump works. We could change the default value, but I'm not sure if we rely on that - perhaps this is something to revisit if it gets asked for?

cellio:
> I don't recall any details now, but I thought there might have been a case that resulted from timings (maybe a user pressing save on an answer after the question was deleted?), so unless we can rule that out we might get very rare new cases in future.
> 
> If that isn't already ruled out, would a potential solution be to have a job to delete such answers, and have it run just before the data dump each week?

Oh, I think you're right.  A cleanup job sounds like a good idea (and is separable from this PR).

app/views/dumps/index.html.erb
```diff
@@ -0,0 +1,72 @@
...
+      <% @others.each do |dump| %>
+        <tr>
+          <td><%= dump.title %></td>
+          <td><%= dump.created_at.strftime('%Y-%m-%d') %></td>
+          <td>
+            <% if dump.file.attached? %>
+              <%= link_to 'Download', rails_blob_path(@latest.file, disposition: 'attachment') %>
+            <% elsif dump.link.present? %>
+              <%= link_to 'View', dump.link %>
+            <% end %>
```
cellio, line 66: Do we want checksums for manual dumps too?

cellio (approved):
Tested and works great for me.  I read the code changes and understood most but not all of them, so treat this as a functionality review more than a code review.

Edit: Art walked me through the code I was having trouble understanding.  LGTM!

app/views/dumps/index.html.erb
ArtOfCode-: No, because those will just be links to IA rather than files - nothing to checksum.

ArtOfCode-:
I have no idea why a sign-in rate limit test has suddenly started failing and have run out of time to work on this for a few days - if anyone else has time to look please feel free.

cellio:
> I have no idea why a sign-in rate limit test has suddenly started failing and have run out of time to work on this for a few days - if anyone else has time to look please feel free.

It's because of the new default creation/update times for users.  Don't merge until we resolve.

cellio (approved):
Retested with the latest changes and all looks good now: new users have the correct (real) timestamps, and data dumps show 1970 for users, community users, and votes.  Changing the default in the dump DB only from within the dump job is clever!

PR #2049, opened 2026-05-14
Add PII correlation tool

### LINKED ISSUE #2000: Add a PII-correlation tool for moderators
Add a tool which enables moderators to identify where there are PII correlations between users. This should be a user-specific mod tool (i.e. accessible via user mod tools rather than the community-wide mod tools menu). This should:

* allow moderators to select a specific other user against whom to correlate the current user's PII
* protect user PII by not revealing exact PII values
   * where it's necessary to display _a value_ to moderators, this should be a hash of the stored value rather than the information itself
   * information may be broken down if this helps to identify connections (i.e. email address into handle and domain, IP address into octets) but each part should still be hashed
   * it may help to display only the first _n_ characters of a hash for readability, but the full hash should be available and/or the UI should clearly display if two hashes are identical
* simply display a measure of correlation (as hashes for very similar values will be completely different)
   * an IP address where 3 of 4 octets are identical is a stronger correlation than just the first octet (consider also IPv6)
   * an email address on the same domain may mean nothing for gmail.com, but could mean everything for a personal domain; consider some way to measure how commonly used an email domain is
* [optional] separately from specific-user selection, may suggest other users whose PII has some correlation with the current user
   * this will require additional querying - consideration needed to optimisation; potentially gated behind an additional click rather than always requested

ArtOfCode-:
Closes #2000.

Adds a tool for moderators to correlate user PII, as described in the linked issue.


cellio (commented):
Tested on my dev server.  The tool looks good, a user with a different email domain shows as different (good), all the IPs look the same.  I tried using a VPN to log in one of my users but it still showed up all red; I'm not sure if localhost overrides VPN or the weekly job needs to run or something else.  We should test this on the dev server before deploying to make sure we aren't unintentionally implicating a lot of people who aren't really on the same IPs.

For email addresses, I only saw "unknown number of users", even when I made up fake domains.  What's an example of a domain that should produce some other description there?

Most of the code here is above my head; someone else should definitely review.

ArtOfCode-:
> For email addresses, I only saw "unknown number of users", even when I made up fake domains. What's an example of a domain that should produce some other description there?

These stats are recalculated every week and cached in between. To generate the initial cache, run `UpdateUserStatsJob.perform_now` in a Rails console.

> I'm not sure if localhost overrides VPN or the weekly job needs to run or something else.

Localhost probably overrides VPN. You can tell if it's a localhost address - for IPv4 that's 127.0.0.1; although it's hashed you'll be able to see that the middle two octets are the same and the first and last are different. For IPv6 it's `::1`, which expands to a lot of zeroes and a 1; again, the hash for every hexadecet (favourite new word) except the last will be the same.

cellio:
These two users have the same email domain; should that have been highlighted?  A domain isn't as strong an indicator as IP is (particularly with popular providers), but we say we highlight matches, so seems like we should either move that notice to before the IP block or do it for email too.

[Screenshot 2026-05-19 at 10 36 41 AM](screenshot: matching domains with no red, matching IPs with red)

ArtOfCode-:
> These two users have the same email domain; should that have been highlighted?

Yep, apparently I completely missed that

cellio (commented):
Tested and LGTM.

PR #2056, opened 2026-05-19
Include post author in comment pings

### LINKED ISSUE #2046: Lack of post author notification of a comment in a thread under their post
Raised in [Discord](https://discord.com/channels/709567671087136829/709570137791135744/1502468728191713411):

> In the post https://codegolf.codidact.com/posts/296063 there is a comment thread labeled Output format.  User wizzwizz4 left me a comment in that thread but I didn’t receive an inbox notification for their comment.  Why not?

Note that the lack of notification is for another user joining the thread. There had already been comments from the user who initiated the thread and the post author. The post author confirms that they are a follower of the thread (it says "unfollow" at the top), and that they received notifications for the comments from the thread author, just not for the new commenter.

In addition to the comment that lead to no notification being from a user new to the thread, it was also the first comment in the thread in which the post author was `@` pinged. I don't know if this is relevant.

Note that the following screenshot is from my user, so says "follow" at the top, unlike for the post author, where it says "unfollow" at the top. The new commenter that triggered no notification is wizzwizz4:

The thread showing comments from the thread author and post author, then a comment from a new user pinging the post author

trichoplax:
Fixes #2046.

In #1548 I neglected to remove 1 special case treatment of the post author, resulting in no notification when the post author is mentioned in a comment under their own post. This pull request adds a fix and tests.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -92,9 +92,8 @@ class CommentsController < ApplicationController
                                         .where('link LIKE ?', "#{thread_url}%")
         next if existing_notification.exists?
 
-        title = @post.parent.nil? ? @post.title : @post.parent.title
         follower.user.create_notification("There are new comments in a followed thread '#{@comment_thread.title}' " \
-                                          "on the post '#{title}'",
+                                          "on the post '#{@comment.root.title}'",
                                           helpers.comment_link(@comment))
       end
     else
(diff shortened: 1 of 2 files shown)


trichoplax:
For prioritisation decisions, until this fix is deployed:
- A post author `@` mentioned in a new thread will be notified of the new thread (but not the mention - this pull request fixes that).
- A post author `@` mentioned in a comment in an existing thread ***will not be notified at all*** (neither of the new comment nor of the mention - this pull request fixes that).
- A post author not `@` mentioned will be correctly notified of new threads and new comments.

cellio (approved):
LGTM.  Tested to confirm the fixes listed in https://github.com/codidact/qpixel/pull/2056#issuecomment-4490007783 .  I noticed that a post author pinged in a new comment thread now receives a notification for the pinging comment but not also the thread; this seems reasonable to me, as you're automatically being notified of the new thread indirectly and it would be too hard to say "new thread" in the notification in this case (and we don't want double notifications there).

ArtOfCode- (approved):

PR #2057, opened 2026-05-19
Make ping notification reflect new thread or comment in a thread

trichoplax:
When a follower of new threads on a post is `@` mentioned in a comment that starts a new thread, the notification now refers to "a new thread" rather than "a comment in a thread", for consistency with the notification received when there is no `@` mention.


ArtOfCode- (approved):

cellio (approved):
LGTM

PR #2063, opened 2026-06-11
User moderator tools user interface redesign

trichoplax:
This converts the user moderation tools from a pop up panel on the user page to a tab at the top of the user page. On desktop this has a left hand panel listing all of the user moderation tools. On mobile this list is available by pressing "Show User Moderation Tools" near the top of the tab.

Note that this change also includes some tidying: code formatting to match the style guide and use of the new user model methods such as `admin?` instead of `is_admin`. Please let me know if you would prefer any of these to be a separate pull request. My aim has been to get these out of the way now to minimise such changes in later user moderator tools pull requests.


app/views/shared/_user_mod_sidebar.html.erb
```diff
@@ -0,0 +1,94 @@
...
+        <div class="menu--heading">Global Mod Tools</div>
+        <%= link_to mod_delete_network_account_path(user),
+                    class: "menu--item #{current_page?(mod_delete_network_account_path(user)) ? 'is-active' : ''}" do %>
+          <i class="fas fa-skull fa-fw"></i>
+          Delete Network Account
+        <% end %>
+        <%= link_to mod_failban_path(user),
+                    class: "menu--item #{current_page?(mod_failban_path(user)) ? 'is-active' : ''}" do %>
+          <i class="fas fa-eye-slash fa-fw"></i>
+          Fail-ban
```
ArtOfCode-: Can we change this to "Feed to STAT" so that we keep the phrasing unchanged?

app/views/users/_tabs.html.erb
```diff
@@ -28,4 +33,25 @@
...
+      current_page?(user_annotations_path(user)) ||
+      current_page?(mod_vote_summary_path(user)) ||
+      current_page?(user_privileges_path(user)) ||
+      current_page?(mod_warning_log_path(user)) ||
+      current_page?(new_mod_warning_path(user)) ||
+      current_page?(mod_delete_path(user)) ||
+      current_page?(mod_delete_network_account_path(user)) ||
+      current_page?(mod_failban_path(user)) ||
+      current_page?(start_impersonating_path(user))
+    ) ? 'is-active' : ''}" do %>
```
ArtOfCode-: Should we consider moving this to a helper method instead?

app/views/users/full_log.html.erb
```diff
@@ -1,44 +1,71 @@
...
-        Show all events
-    </a>
-    <% if @interesting > 0 %>
-    <a class="button is-muted is-outlined <%= params[:filter] == 'interesting' ? 'is-active' : '' %>" href="?filter=interesting" role="button">
-        Negative
-        <span class="badge is-status"><%= @interesting %></span>
-    </a>
+    <p class="is-lead">
+      This is a filterable log for all activity by the user. You can consult it for moderation decisions. Do not share
+      this information to people, who do not have access to it.
```
ArtOfCode-: Stray comma! I know this is just copy/paste but let's take the chance to fix them while we're here.

ArtOfCode- (changes requested):

app/views/users/_tabs.html.erb
trichoplax: Since these pages are scattered over several controllers, I've added a helper to the application helpers.

ArtOfCode- (approved):

cellio (approved):
I ran through the new tools pages and everything looks good.  I'll file follow-on issues for the things we decided (in the previous PR) to defer.

app/helpers/application_helper.rb
```diff
@@ -394,4 +394,23 @@ def safe_page(params)
...
+      current_page?(mod_vote_summary_url(user)) ||
+      current_page?(user_privileges_url(user)) ||
+      current_page?(mod_warning_log_url(user)) ||
+      current_page?(new_mod_warning_url(user)) ||
+      current_page?(mod_delete_url(user)) ||
+      current_page?(mod_pii_correlation_url(user)) ||
+      current_page?(mod_delete_network_account_url(user)) ||
+      current_page?(mod_failban_url(user)) ||
+      current_page?(start_impersonating_url(user))
+  end
```
Oaphi: This looks like a good candidate for a loop, this is way too many ORs for one method

Oaphi (dismissed):

app/helpers/application_helper.rb
trichoplax: Oh good point. I did some searching and I've found 2 approaches. I've implemented one, but the following also works:

```rb
  def user_mod_tools_tab_active?(user)
    user_mod_tools_tab_pages = [
      mod_user_url(user),
      full_user_log_url(user),
      user_annotations_url(user),
      mod_vote_summary_url(user),
      user_privileges_url(user),
      mod_warning_log_url(user),
      new_mod_warning_url(user),
      mod_delete_url(user),
      mod_pii_correlation_url(user),
      mod_delete_network_account_url(user),
      mod_failban_url(user),
      start_impersonating_url(user)
    ]

    user_mod_tools_tab_pages.any? { |page| current_page?(page) }
  end
```

I'm not sure which would be more readable, so let me know if you'd prefer the other approach (or if there's a better approach I haven't considered).

Oaphi: I was thinking about something a bit more succinct - what'd you say if took this approach instead:
```ruby
def user_mod_tools_tab_active?(user)
    methods = [
      :mod_user,
      :full_user_log,
      :user_annotations,
      :mod_vote_summary,
      :user_privileges,
      :mod_warning_log,
      :new_mod_warning,
      :mod_delete,
      :mod_pii_correlation,
      :mod_delete_network_account,
      :mod_failban,
      :start_impersonating
    ]

    methods.any? { |name| current_page?(send("#{name}_url", user)) }
  end
```

Oaphi: [`current_page?`](https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-current_page-3F) also accepts `controller` and `action` options so we could use that form to simplify even further.

Oaphi:
I'd like to check out the PR locally before we merge it, but do not block on me if I don't get to it in the next couple of days

trichoplax:
~~I recommend waiting until #2098 is merged as this will allow simplifying this pull request.~~ (Note that #2098 has now been merged and this pull request has been simplified.)

Oaphi:
Screenshot from 2026-07-24 14-48-10

@trichoplax - how about we color the dangerous actions red so as it's obvious from the start that choosing one of those should not be treated lightly? Not RC'ing over this, just a suggestion

Oaphi (approved):
As we've agreed to move the "red-coloring" step to a standalone issue+PR pair, LGTM

PR #2069, opened 2026-06-18
Add dnf prerequisite deps

SuperCowProducts:


ArtOfCode- (approved):

INSTALLATION.md
```diff
@@ -26,6 +26,14 @@ sudo apt install autoconf bison build-essential libssl-dev libyaml-dev libreadli
 sudo apt install mysql-server libmysqlclient-dev
 ```
 
+For Red Hat-Based Linux:
+
+```
+sudo dnf install gcc make pkg-config
+sudo dnf install autoconf bison @development-tools openssl-libs libyaml-devel readline-devel zlibrary-devel ncurses-devel libffi-devel gdbm-devel
```
Oaphi, line 33: I am not familiar with Red Hat's distros, unfortunately, and am struggling to find any docs on that matter, hence a quick question - what's `@development-tools`? Looks good otherwise, thanks for helping improve the installation docs!

SuperCowProducts: The issue is there's no real one-to-one correspondence with `build-essential`, so generally that is what is used. In any case I thought it would be useful to at least give `dnf` users a 'headstart': with dependency management there is always room for improvement :smile:

Oaphi: Oh yeah, for sure 😅 Don't mind me, just managed to find it's the new-ish group / environment spec-related syntax from dnf5 (I was primarily puzzled by the `@` prefix).

PR #2073, opened 2026-06-22
Remove reimplementations of String#truncate

### LINKED ISSUE #1420: Code-only improvement: avoiding reimplementing the built in truncate method
**Describe the bug**
In the file `app/views/users/_activity_items.html.erb` there are 4 occurrences of truncating a string using a line of code, which has identical behaviour to the built in `truncate` method:
- line 22: `<%= i.body_plain[0..300] + ((i.body_plain.length > 300) ? "..." : "") %><br>`
- line 34: `<%= i.content[0..300] + ((i.content.length > 300) ? "..." : "") %><br>`
- line 73: `<%= i.body[0..300] + ((i.body.length > 300) ? "..." : "") %>`
- line 88: `<%= i.reason[0..300] + ((i.reason.length > 300) ? "..." : "") %><br>`

For example usage of the `truncate` method, see `app/views/posts/show.html.erb` line 1.

We could avoid this duplicated code by replacing these 4 occurrences with calls to `truncate`.

ArtOfCode-:
Closes #1420.

Removes instances where we've semi-re-implemented `String#truncate` and uses that instead. Also deprecates `ApplicationHelper#split_words_max_length` in favour of the former.

Unrelated small change: adds additional environment variables for dev environments to control email settings.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -44,11 +44,15 @@ Rails.application.configure do
   config.hosts << /[a-z0-9\-.]+\.ngrok-free\.app/
 
   # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
-  config.action_mailer.delivery_method = :ses
+  raise_delivery_errors = ActiveRecord::Type::Boolean.new.cast(ENV['RAISE_DELIVERY_ERRORS'])
+  config.action_mailer.raise_delivery_errors = raise_delivery_errors || false
   config.action_mailer.asset_host = 'https://meta.codidact.com'
-
   config.action_mailer.perform_caching = false
+  delivery_method = ENV['MAILER_DELIVERY_METHOD']
+  config.action_mailer.delivery_method = delivery_method&.to_sym || :letter_opener_web
+  config.action_mailer.default_url_options = {
+    host: 'meta.codidact.com', protocol: ENV['MAILER_PROTOCOL'] || 'https'
+  }
 
   # Print deprecation notices to the Rails logger.
   config.active_support.deprecation = :log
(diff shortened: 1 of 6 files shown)


config/environments/development.rb
```diff
@@ -44,11 +44,15 @@
   config.hosts << /[a-z0-9\-.]+\.ngrok-free\.app/
 
   # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
-  config.action_mailer.delivery_method = :ses
+  raise_delivery_errors = ActiveRecord::Type::Boolean.new.cast(ENV['RAISE_DELIVERY_ERRORS'])
```
trichoplax, line 47: How sure are we that the ENV variable will be consistently upper or lower case (thinking of other instances in particular - can they manually change it)? Is it worth a defensive conversion to lowercase here?

In the Rails console:

```txt
:001> ActiveRecord::Type::Boolean.new.cast('False')
=> true
:002> ActiveRecord::Type::Boolean.new.cast('false')
=> false
:003> ActiveRecord::Type::Boolean.new.cast('FALSE')
=> false
```

It appears this is considered correct behaviour and [won't be fixed](https://github.com/rails/rails/pull/39071).

trichoplax (commented):
The replacements with `truncate` and the deprecation of `split_words_max_length` look good.

I don't feel qualified to approve the email config changes so I've just made this a comment. I've added one question in case it happens to apply.

config/environments/development.rb
ArtOfCode-: I'm not too fussed about this - it's just for quick dev reconfigurations (I used it to test the error with email sending that we're having at the moment), not long-term production config.

PR #2074, opened 2026-06-22
Update to Rails 8

ArtOfCode-:
Updates us to Rails 8.1. Also fixes SES emails by changing to AWS' official gems and config setup for this purpose.


Oaphi:
Quick note: don't merge before I get the chance to look through it if possible (likely not today, unfortunately) - I'd like to comb through what does updating to the next major version of Rails ([migration docs](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-7-2-to-rails-8-0) for reference - ATTOW 7.2 -> 8.0 migration is just a pointer to the changelog) means for us. Especially in light of the following change:
> [Propshaft](https://github.com/rails/propshaft) is now the default asset pipeline, replacing the old Sprockets system.

P.S. Don't block on me if I am not available, though, we can deal with arising issues later (hopefully)

Oaphi (dismissed):
I'll be addressing a few things we need to do before bumping to v8 shortly

PR #2085, opened 2026-07-10
Set network-specific 2FA token issuers

### LINKED ISSUE #2084: 2FA tokens always use codidact.com rather than the network name
I set up 2FA on an account on the dev server at dev.codidact.org, but the entry in my 2FA app shows `codidact.com` and `users-2fa.codidact.com`.  I spent several minutes thinking I'd lost track of my test account because of the mistaken domain.

The entry in the app should use the actual network URL.  I'm assuming that each qpixel instance handles its own 2FA and this is only a "wrong info" problem and not that everything actually depends on one service hosted by codidact.com, but let's confirm that.

ArtOfCode-:
Closes #2084

### DIFF
--- a/app/controllers/two_factor_controller.rb
+++ b/app/controllers/two_factor_controller.rb
@@ -16,8 +16,9 @@ class TwoFactorController < ApplicationController
       @secret = ROTP::Base32.random
       current_user.update(two_factor_token: @secret, two_factor_method: 'app',
                           backup_2fa_code: SecureRandom.alphanumeric(24))
-      totp = ROTP::TOTP.new(@secret, issuer: 'codidact.com')
-      uri = totp.provisioning_uri("#{current_user.id}@users-2fa.codidact.com")
+      base_domain = AppConfig.server_settings['network_base_domain']
+      totp = ROTP::TOTP.new(@secret, issuer: base_domain)
+      uri = totp.provisioning_uri("#{current_user.id}@users-2fa.#{base_domain}")
       qr_svg = RQRCode::QRCode.new(uri).as_svg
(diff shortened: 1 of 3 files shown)


app/controllers/two_factor_controller.rb
```diff
@@ -16,8 +16,9 @@ def enable_2fa
       @secret = ROTP::Base32.random
       current_user.update(two_factor_token: @secret, two_factor_method: 'app',
                           backup_2fa_code: SecureRandom.alphanumeric(24))
-      totp = ROTP::TOTP.new(@secret, issuer: 'codidact.com')
-      uri = totp.provisioning_uri("#{current_user.id}@users-2fa.codidact.com")
+      base_domain = AppConfig.server_settings['network_base_domain']
```
Oaphi, line 19: Any particular reason to choose app config over a site setting here? We already havea `Enable2FAForSsoUsers` global site setting, so there's precendent for relying on site settings in this controller.
Plus we already keep several network-related customization settings there - what do you think, @ArtOfCode-?

ArtOfCode-: I tend to prefer config for things that are generally set once and forget, vs site settings for things that might change more frequently.

Oaphi: I suppose, I am just a bit worried about having something that an instance _should_ change unless it's us in app config. That said, I am fine with us just adding a couple of notes in README (not in this PR - it's probably a good idea to document those in one go).

Oaphi (approved):
@ArtOfCode- I've added a basic test to improve coverage of (well, more like to start covering) the controller, leaving it up to you to merge if the test looks good to you (LGTM, tested the new setting manually as well).

PR #2086, opened 2026-07-10
Notify flagger of custom responses

### LINKED ISSUE #1924: Notify flagger of custom responses
When handling flags, moderators can write responses, which the flaggers can see on their own flags pages if they happen to go look.  If a moderator took the time to write a custom message, I'd like to increase the chance that users will see it -- "no response needed" is the default case, so most people aren't going to check for responses on their own.

Can handled flags with custom responses produce notifications?  At least to link to that page, if a more specific link is impractical?

ArtOfCode-:
Closes #1924.

### DIFF
--- a/app/controllers/flags_controller.rb
+++ b/app/controllers/flags_controller.rb
@@ -76,9 +76,7 @@ class FlagsController < ApplicationController
   end
 
   def resolve
-    if @flag.update(status: params[:result], message: params[:message], handled_by: current_user,
-                    handled_at: DateTime.now)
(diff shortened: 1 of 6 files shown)


Oaphi (changes requested):
@ArtOfCode- let's simplify AbilityQueue#add & AbilityQueue#add! methods and approach them the same way Rails does - without extra paramters. I've made a [suggestions PR](https://github.com/codidact/qpixel/pull/2087) to yours - what'd you say (feel free to dismiss the CR review)?

PR #2094, opened 2026-07-15
User profile undelete + ability view fix for non-existent profiles

### LINKED ISSUE #2130: ability pages error out for non-users
URLs of the form `abilities/edit_posts?for=NNN` where NNN is a user number produce this error (from the log in admin tools):

```
NoMethodError: undefined method 'privilege?' for nil

/var/apps/qpixel/app/views/abilities/show.html.erb:27:in '_app_views_abilities_show_html_erb___2642789021255159210_38936'
```
We get a lot of these, for all abilities and various user numbers -- smells like a crawler rather than organic user activity.  While reviewing logs, I also found requests like `...?for=7j9q2y`.

Can we ensure that the argument is (a) a number and (b) for a non-deleted user before attempting to fetch the ability page?

Oaphi:
This PR adds a reversal moderation action of restoring soft-deleted community profiles with the same button that deletes them (might come in handy as we don't have any form of confirmation of such actions). Restoration is audited, so I don't see any issues with having it.

In addition (and what is the main purpose of the PR) is to fix server error when trying to view an ability page for a user with the prorifle not present on the community. That's not supposed to happen anymore but data errors can lead to this + we expect that in other checks in the view, it's just that this specific one is outdated.


cellio (changes requested):
I wondered how this would interact with deleted users (as opposed to deleted community users).  As a moderator I can't restore a user that was deleted network-wide (good), *but* I'm invited to delete the community profile instead, which is a little confusing.  As an admin it's even more confusing (more below).

Sequence:
- admin deletes user network-wide
- mod goes to the user page on the local community (which shows as deleted)
- mod goes to "show all tools" and sees the "danger zone" notice and a button to delete the community profile

Clicking the button appears to work but is nonsensical, and also does not update the timestamp or delete attribution on the profile page, which could be confusing.

If a user was deleted network-wide, can we replace the button and danger-zone notice with text to the effect that this user was deleted network-wide and can't be locally restored, but contact an admin if you think this should be reversed?

Now, switching from a local mod to an admin, I go to the same profile (deleted network-wide), go to "show all tools", and see three buttons:
- undelete community profile (I guess that's because the mod deleted it)
- delete user network-wide (it's already deleted)
- feed to stat (not related to this work, but mentioned for completeness)

For now we should show the same "deleted network-wide, can't be restored" message.  We might want to allow admins to restore deleted users, but that's a more complicated case and I'm not sure it's possible as we clobbered the email address.  Plus we'd have to decide what to do about community users.  So that's more complicated and let's not try to handle it in this PR.

Summary:  in this PR let's
- [x] prevent mods from acting on users who were deleted network-wide
- [ ] not offer restoration buttons to admins through the UI in this case
- [x] add some tests that use network-deleted users

Oaphi:
> We might want to allow admins to restore deleted users, but that's a more complicated case and I'm not sure it's possible as we clobbered the email address. 

Let's keep treating network-wide user deletion as irreversible, at least for the time being - we do clobber both the email and the password, after all. 

As for the rest - yea, won't hurt to handle the "deleted everywhere, not just on this community" case better while I am at it, will do

Oaphi:
Screenshot from 2026-07-20 03-03-37

@cellio I've opted to disable (not remove) the buttons and add a hover title explaining the situation. Let's follow the principle of always showing buttons that the user has access to but can't use right now due to a specifc set circumstances (in this case the user being deleted network-wide).

If I get some time to spare, I'll also add proper translation strings for these actions

cellio (approved):
Tested and LGTM.  Non-blocking: do people using phones/tablets see hover text?

Oaphi:
> Tested and LGTM. Non-blocking: do people using phones/tablets see hover text?

Nope, but that's, unfortunately, true for every single tooltip of ours, not much we can do about it. We could add the explanations to the always visible description, I suppose - that would alleviate the issue

cellio:
> We could add the explanations to the always visible description, I suppose - that would alleviate the issue

That's what I was wondering about -- we do that on the self-delete page if you're not allowed to, so I wonder if we should do it here too.  Admins will probably know, but mods could be puzzled.  On the other hand, most moderation probably doesn't happen on phones.  On the third hand, is it easy?  But as I said, non-blocking because this is just one case of a widespread issue, as you said.

Oaphi:
Needs to be updated now that the mod tools PR is merged in - since it's a low priority issue, I'll deal with it later

ArtOfCode- (approved):

PR #2098, opened 2026-07-16
Pass active tab to tabs partial view

trichoplax:
## Main change
This passes the name of the active tab to the partial view `app/views/users/_tabs.html.erb` to remove the need for each tab to check the current page to decide if it should show as active.

This is a fairly trivial change at this stage, but will become more useful in #2063 where a Moderator Tools tab will be introduced that needs to show as active for 11 different URLs.

## Additional changes
This pull request also includes other changes triggered by working on the tabs:
- The Subscriptions page (for emails) had the tabs along the top, but wasn't listed among them, resulting in an odd look with no active tab. I've added a tab for Subscriptions to fix this, and removed the right hand panel for consistency with the other tabs.
- The tabs now take up enough horizontal space that reading through them takes a while and they run on to two lines on smaller screens or at higher zoom levels. I've reduced their width with the following changes:
  - Removed the icon from the "Profile" tab.
  - Changed "Vote Summary" to "Votes".
  - Changed "Notifications" to "Inbox".
  - Changed "Subscriptions" to "Email".
  - Changed "All Communities" to "Network".

Let me know if you'd prefer the Subscriptions page to be an independent page with no tabs along the top, or if you'd prefer to avoid any of these width reducing changes.

### DIFF
--- a/app/views/users/_tabs.html.erb
+++ b/app/views/users/_tabs.html.erb
@@ -1,31 +1,31 @@
 <div class="tabs is-classic-tabs">
-  <%= link_to user_path(user), class: "tabs--item #{current_page?(user_path(user)) ? 'is-active' : ''}" do %>
+  <%= link_to user_path(user), class: "tabs--item #{active_tab == "profile" ? 'is-active' : ''}" do %>
     <i class="fas fa-user"></i> Profile
   <% end %>
-  <%= link_to user_activity_path(user), class: "tabs--item #{current_page?(user_activity_path(user)) ? 'is-active' : ''}" do %>
+  <%= link_to user_activity_path(user), class: "tabs--item #{active_tab == "activity" ? 'is-active' : ''}" do %>
     Activity
   <% end %>
-  <%= link_to vote_summary_path(user), class: "tabs--item #{current_page?(vote_summary_path(user)) ? 'is-active' : ''}" do %>
+  <%= link_to vote_summary_path(user), class: "tabs--item #{active_tab == "vote summary" ? 'is-active' : ''}" do %>
     Vote Summary
   <% end %>
   <% if user == current_user %>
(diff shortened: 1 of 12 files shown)


cellio (approved):
LGTM. Ran through the tabs at both desktop and phone sizes.

ArtOfCode- (approved):

app/views/users/_tabs.html.erb
```diff
@@ -1,31 +1,34 @@
 <div class="tabs is-classic-tabs">
-  <%= link_to user_path(user), class: "tabs--item #{current_page?(user_path(user)) ? 'is-active' : ''}" do %>
-    <i class="fas fa-user"></i> Profile
+  <%= link_to user_path(user), class: "tabs--item #{active_tab == 'profile' ? 'is-active' : ''}" do %>
```
Oaphi: If you can change it to fit the new, more compact, approach, it'd be great (note the lack of ternary and the empty string default - it's handled under the hood):
```ruby
'is-active' if active_tab == 'profile'
```
Example usage:
https://github.com/codidact/qpixel/blob/1aa43790c52e547ad6bc4146b9bd3aa575fa8865/app/views/layouts/osa_training.html.erb#L32-L44

Oaphi (changes requested):

app/views/users/_tabs.html.erb
Oaphi: Once again, this is not critical, so feel free to override this RC review, it's mostly to draw attention

trichoplax: Thanks for letting me know about that. I'm familiar with the similar form in Python but I didn't realise Ruby had that too. I've now applied the change throughout the file.

Oaphi (approved):

PR #2100, opened 2026-07-16
Fix regression preventing users from emptying their profile text

### LINKED ISSUE #2099: Emptying profile text fails but claims success
**Describe the bug**
After deleting all of the text in my user's Profile field, and pressing Save, a message is displayed saying "Your profile details were updated" but the previous text is still present. Attempting to edit again shows the Profile field empty during the edit, even though it is not empty after saving.

**To Reproduce**
1. Go to your own user page and ensure there is already profile text.
2. Click on the "Edit" tab.
3. Delete all of the profile text.
4. Press Save.
5. See "Your profile details were updated" in a green box, but no change to the profile.

**Expected behavior**
After saving, the profile should be empty (showing the default message for empty profiles).

Oaphi:
closes #2099

### DIFF
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -167,12 +167,19 @@ class UsersControllerTest < ActionController::TestCase
     assert_equal users(:standard_user).id, assigns(:user).id
   end
 
-  test 'should update profile text' do
+  test 'should correctly update profile text' do
     sign_in users(:standard_user)
-    patch :update_profile, params: {
-      user: { profile_markdown: 'ABCDEF GHIJKL' }
-    }
-    assert_equal assigns(:user).profile.strip, '<p>ABCDEF GHIJKL</p>'
+
+    ['Non-empty profile text', ''].each do |text|
+      patch :update_profile, params: {
+        user: { profile_markdown: text }
+      }
+
+      @user = assigns(:user)
+      assert_not_nil @user
+
+      assert_equal @user.profile.strip, text.present? ? "<p>#{text}</p>" : ''
+    end
   end
 
   test 'should update websites' do
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -170,7 +170,7 @@ class UsersControllerTest < ActionController::TestCase
   test 'should correctly update profile text' do
     sign_in users(:standard_user)
 
-    ['Non-empty profile text', ''].each do |text|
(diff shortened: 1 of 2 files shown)


test/controllers/users_controller_test.rb
```diff
@@ -167,12 +167,19 @@ class UsersControllerTest < ActionController::TestCase
...
 
-  test 'should update profile text' do
+  test 'should correctly update profile text' do
     sign_in users(:standard_user)
-    patch :update_profile, params: {
-      user: { profile_markdown: 'ABCDEF GHIJKL' }
-    }
-    assert_equal assigns(:user).profile.strip, '<p>ABCDEF GHIJKL</p>'
+
+    ['Non-empty profile text', ''].each do |text|
```
trichoplax: I recommend adding a test here for a whitespace only profile text. Perhaps something like:

```rb
['Non-empty profile text', '', '    '].each do |text|
```

trichoplax (changes requested):
I think there's an edge case where this will still fail (leaving the profile text rendered as if there were no change, while leaving the profile field in the "Edit" tab still appearing empty to the user). I've suggested an extra test to catch this.

test/controllers/users_controller_test.rb
trichoplax: Having tried that myself, it turns out it still passes due to both `''` and `'    '` counting as empty. I got a failing test by making the profile text non-empty again between testing `''` and `'    '`:

```rb
['Non-empty profile text', '', 'More non-empty profile text', '    '].each do |text|
```

trichoplax: Note that this isn't a quirk of the test - as a user I can edit my profile to have text, and then edit it to have `'    '`, and the previous text continues to show.

Oaphi: hmm, interesting. Yeah, it's a quirk of `blank?` (of which `present?` is the inverse) that's causing it - whitespace-only strings count as "blank?" (thanks, Rails) but not `empty?`, which only considers zero-length strings (thanks, Ruby). As a result, whitespace-only strings do not match any of the branches. Your first check should've failed too, it's just that the test uses `strip` + `present?`, which makes the empty and whitespace-only strings equal. 

Good catch, let's just trim the text when saving - it'll also prevent users from testing mod patience by submitting giant whitespace-only profiles.

trichoplax (approved):
Looks good, and saving the profile works now, for empty, whitespace, and text.

cellio (approved):
Tested and confirmed fix.  LGTM.

PR #2113, opened 2026-07-28
Tests for post body / body_markdown length validation

Oaphi:
Followup for #2111


test/models/post_test.rb
```diff
@@ -77,6 +77,38 @@ class PostTest < ActiveSupport::TestCase
...
+                      user: user }
+
+    [
+      ['', '', false],
+      ['', '<p></p>', false],
+      ['  ', '<p>  </p>', false],
+      ['# Heading', '<h1>Heading</h1>', true],
+      ['a' * 30_000, 'a' * 65_535, true],
+      ['a' * 30_001, '<p>body_markdown is too long</p>', false],
+      ['body is too long', 'a' * 65_536, false]
```
trichoplax, line 103: We could have an extra case to cover #2115. Maybe something like:

```rb
["a\r\n" * 15_000, '<p>body_markdown is too long before converting newlines to \n</p>',  true]
```

trichoplax: On second thought, let's get this merged as is, and then I'll add such tests to #2115 so this doesn't have to wait for that one to be merged.

trichoplax (approved):

PR #2116, opened 2026-07-28
fix voting access for restricted users

### LINKED ISSUE #2114: Users without Participate Everywhere can still vote freely
meta:296429

The Participate ability is supposed to allow you to vote only on answers to your own questions, and comment only on those answers and your own posts.  The Participate Everywhere ability lifts those restrictions so you can vote and comment on all posts.

We are not enforcing the restriction on which posts you can vote on (but we are on comments).  This means users with Participate (only) can vote freely but can't explain those votes.

I did some looking around in the code.  It looks like there's nothing in the backend that blocks either votes or comments in this case.  For comments, the view grays out the button if you're not allowed (I haven't dug into how it decides that yet).  For votes, we apparently don't do anything in the view to check.  If you try to vote on your own post you get a red-box error saying you can't do that; ideally we would catch out-of-scope votes similarly and explain the ability restriction in the message.

cellio:
Fixes https://github.com/codidact/qpixel/issues/2114.  After checking for trying to vote on your own post, the votes controller now also checks if you are missing the "unrestricted" ability and this isn't an answer to your question.  It's important that this is the second check, because we're calling `owns_post_or_parent?` instead of writing a new very similar check.  If it's your post we've already bailed, so this is really just checking whether you own the parent.

I tested this manually with restricted and unrestricted users voting on their own posts, answers to their questions, and unrelated posts.  I also pre-emptively ran rubocop before committing, so I'm not sure what's up with that test failure here.


app/controllers/votes_controller.rb
```diff
@@ -7,7 +7,14 @@ def create
     post = Post.find(params[:post_id])
 
     if post.user == current_user && !SiteSetting['AllowSelfVotes']
-      render(json: { status: 'failed', message: 'You may not vote on your own posts.' }, status: :forbidden)
+      render(json: { status: 'failed', message: I18n.t('votes.limits.own_post') },
+             status: :forbidden)
+      return
+    end
+
+    if !current_user.privilege?('unrestricted') && !current_user.owns_post_or_parent?(post)
```
Oaphi: Let's make it a proper access control method on the user model. Something along the lines of `can_vote_on?(post)`

Oaphi (changes requested):

app/controllers/votes_controller.rb
cellio: I moved the new test into a method on the model as you suggested.  I did not also do the site-setting check there, because the `can_something?` methods by convention return booleans, but here we have two different error messages for the two reasons you might not be allowed to vote.

cellio: According to rubocop, the user model is now too long by two lines.  I didn't find any obvious opportunities for consolidating that don't compromise legibility.

Oaphi (approved):

PR #2118, opened 2026-07-30
Show dangerous user mod tools menu items in red

trichoplax:
Following a suggestion during review of #2063 this applies a red colour to the links in the left hand panel of the user moderator tools tab of the user page.

Menu items showing grey or red for dangerous items or blue background for the active item

This uses the `.is-red` class, which for links is darkened by 25%. I've left this as is for now, but I'd like feedback on which of the following approaches to take:

- Use this same darkened red (as pictured) for menu item links as any other links.
- Make the red brighter just for menu items.
- Make the red brighter for all links.

The red text is only applied for menu items that are not the active one, since once active there is red text in the specific tab content. Active menu items show with a blue background regardless of whether dangerous. Would you prefer this or to change to a red background?

### DIFF
--- a/app/views/shared/_user_mod_sidebar.html.erb
+++ b/app/views/shared/_user_mod_sidebar.html.erb
@@ -67,7 +67,7 @@
           Compare PII
         <% end %>
         <%= link_to mod_delete_path(user),
-                    class: "menu--item #{current_page?(mod_delete_path(user)) ? 'is-active' : ''}" do %>
+                    class: "menu--item #{current_page?(mod_delete_path(user)) ? 'is-active' : 'is-red'}" do %>
           <i class="fas fa-trash fa-fw"></i>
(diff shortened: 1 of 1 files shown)


Oaphi:
> Would you prefer this or to change to a red background?

Keeping the blue background is fine by me - after all, one has to actually navigate to the section for this to happen, right? At which point there's a giant banner on the right side explaining that the action is dangerous.

cellio (approved):
LGTM in both desktop and phone layouts.  From the screenshot I wondered if it would be a little *too* attention-grabbing, but in context it looks fine to me.  I don't see a need to add bespoke CSS here.

Oaphi:
> This uses the .is-red class, which for links is darkened by 25%.

Fine by me too, all other "danger" link buttons have the same color (comment tools, for example)

PR #2119, opened 2026-07-30
Normalize post body_markdown newlines

### LINKED ISSUE #2115: Newlines converted to Windows style on saving post
**Describe the bug**
Following on from #2111

See also [discussion on Discord](https://discord.com/channels/634104110131445811/668890340199235613/1531679877638656130).

When a post is submitted (a new post or an edit) all of its newlines are converted from `\n` to `\r\n`. This means a post that is below the limit of 30,000 characters can become over the limit and be rejected.

**To Reproduce**
1. Make a new post or edit an existing one.
2. Make sure it contains at least 1 newline.
3. On the server, see that drafts are saved with `\n` newlines.
4. Save the post.
5. On the server, see that the submitted post has `\r\n` newlines.

**Expected behavior**
Submitting a post should not change the length of its raw Markdown.

Ideally `\r\n` would be converted to `\n` so that Windows users get the full 30000 limit rather than losing a character per line, and Mac and Linux users get consistent numbers when editing regardless of the operating system a post was created on.

trichoplax:
Fixes #2115 

Normalizes the `body_markdown` field in the post model so that newlines are all `\n`.

Applied only to this one field that was raised in #2115 for now, but let me know any other places you would also like to see this normalization, or if there's a preferred general approach. Provided your suggested approach doesn't clash with this, I'm hoping to get the specific fix merged and deployed first, then apply anything further separately once the bug is no longer holding up a question being posted.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -31,6 +31,10 @@ class Post < ApplicationRecord
 
   serialize :tags_cache, coder: YAML, type: Array
 
+  normalizes :body_markdown, with: -> body_markdown {
+    body_markdown.encode(body_markdown.encoding, universal_newline: true)
+  }
+
   validates :body, presence: true, length: { maximum: 65_535 }
   validates :body_markdown, presence: true, length: { maximum: 30_000 }
   validates :doc_slug, uniqueness: { scope: [:community_id], case_sensitive: false }, if: -> { doc_slug.present? }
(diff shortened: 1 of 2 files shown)


cellio (approved):
Tested locally with the post that was failing in prod.

Oaphi (approved):

PR #2120, opened 2026-07-30
YARD fixes + UserSortable concern 

Oaphi:
Additional manual testing requested: sorting of logs and posts (just that it still works as expected)

### DIFF
--- a/app/models/concerns/user_sortable.rb
+++ b/app/models/concerns/user_sortable.rb
@@ -1,0 +1,27 @@
+module UserSortable
+  extend ActiveSupport::Concern
+
+  class_methods do
+    # Sort a collection according to a user selection, by mapping user selectable values to column names.
+    # SQL injection safe.
+    # @param term_opts Hash of search term options
+    # @param field_mappings Hash of user-selectable values to column names: +{ age: :created_at, rep: :threshold }+
+    # @option term_opts :term [String] A user-provided search term to apply - usually from +params+, e.g. +params[:sort]+.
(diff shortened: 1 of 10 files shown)


app/models/concerns/user_sortable.rb
```diff
@@ -0,0 +1,27 @@
+module UserSortable
+  extend ActiveSupport::Concern
+
+  class_methods do
+    # Sort a collection according to a user selection, by mapping user selectable values to column names.
```
cellio, line 5: I'm not super-familiar with how all this is wired together, but should this have turned into doc at https://docs.dev.codidact.org/docs/UserSortable?  Or am I looking in the wrong place?

cellio (commented):
Don't rely just on me, but this looks good to me.  That weird "align=center" thing on the landing page was there before too; I don't know what's causing it, but it's not a regression.

I reviewed only the doc, not the part where you moved UserSortable into a concern.

app/models/concerns/user_sortable.rb
Oaphi: No clue, those are concerns with `class_methods` instead of `included` - that's likely a limitation of yardoc. I can take a look just in case, but pretty sure that's not an issue on our side

cellio: Ah, different type of method.  I don't think we need to fuss over it; I was just trying to figure out if there was an error in either the content or the build setup that we were tripping over.  Not that, so fine with me.

ArtOfCode- (approved):

PR #2127, opened 2026-07-31
Add test to fail when an edit is only line ending style changes

### LINKED ISSUE #2126: Empty edits can be made
**Describe the bug**
Following #2119 or #2123 it is possible to save an edit that makes no changes. Previously this was impossible, and we have a test to ensure this, but the test had body_markdown containing no newlines, so did not detect that a line ending style only change is detected as a change and allows the edit.

**To Reproduce**
I've [added a test to detect an edit being accepted with only line ending style changes](https://github.com/codidact/qpixel/pull/2127) that is correctly failing due to this issue. It can be merged once a fix makes it pass.

See also [this example in production](https://meta.codidact.com/posts/296432/history#7). I caused this simply by pressing Edit under the question and pressing Save.

**Expected behavior**
An edit that contains only line ending style changes should be treated as an empty edit and rejected.

trichoplax:
Fixes #2126 

A post edit that only changes line endings will now be rejected. I've aimed to make this compatible with existing posts that may have `body_markdown` using any type of line endings, and newer posts that will have exclusively normalized `\n` (LF) line endings.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -446,7 +446,7 @@ class ApplicationController < ActionController::Base
   end
 
   def normalize_if_string(maybe_string)
-    return maybe_string unless maybe_string.is_a? String
+    return maybe_string unless maybe_string.is_a?(String)
 
     maybe_string.encode(maybe_string.encoding, universal_newline: true).strip
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -444,4 +444,12 @@ class ApplicationController < ActionController::Base
   def request_authenticity_tokens
     super << csrf_token_storage_strategy.fetch(request)
   end
+
+  def normalize_if_string(potential_string)
+    if potential_string.is_a? String
+      potential_string.encode(potential_string.encoding, universal_newline: true).strip
+    else
+      potential_string
+    end
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -446,10 +446,8 @@ class ApplicationController < ActionController::Base
   end
 
   def normalize_if_string(maybe_string)
-    if maybe_string.is_a? String
-      maybe_string.encode(maybe_string.encoding, universal_newline: true).strip
-    else
-      maybe_string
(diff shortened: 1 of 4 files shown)


app/controllers/posts_controller.rb
```diff
@@ -712,6 +712,14 @@ def legal_delete
 
   private
 
+  def normalize_if_string(potential_string)
```
Oaphi: Style: we already have at least one `maybe_`-prefixed param for situations like this - let's standardize on it before our approaches diverge?

Oaphi: This doesn't seem like it belongs in a controller - let's make it an application_helper - should be useful for much more than just posts

Oaphi (changes requested):

app/controllers/posts_controller.rb
```diff
@@ -712,6 +712,14 @@ def legal_delete
...
+    if potential_string.is_a? String
```
Oaphi: Let's exit early for success critera - we tend to do that with other helpers (note that the below is a typed in suggestion):

```ruby
return maybe_string unless maybe_string.is_a?(String)
```
Then it's just a matter of making the normalization logic the last line of the helper

Oaphi (changes requested):

app/controllers/application_controller.rb
```diff
@@ -444,4 +444,10 @@ def require_sudo
   def request_authenticity_tokens
     super << csrf_token_storage_strategy.fetch(request)
   end
+
+  def normalize_if_string(maybe_string)
```
Oaphi: And a final thing - let's just call it `normalize_content` or something along these lines () - `if_string` is an implementation detail. This would align the helper with Rails's naming convention: `normalize_attribute`, `normalize_path`, etc

Oaphi (changes requested):

Oaphi (approved):

Oaphi (changes requested):
Didn't realize the `normalize_if_string` -> `normalize` change is not pushed out yet, oops

Oaphi (approved):

PR #2128, opened 2026-08-02
Manually exclude line ending changes from diffs

trichoplax:
It seems Diffy's `ignore_crlf` option does not work consistently so this removes its use and instead passes normalized `before` and `after` strings to Diffy, removing the need for the option.

### DIFF
--- a/app/views/post_history/_diff.html.erb
+++ b/app/views/post_history/_diff.html.erb
@@ -1,8 +1,8 @@
 <div class="diff">
   <% if before.present? && after.present? %>
     <% if before.is_a?(String) && after.is_a?(String) %>
-      <% before = post.normalize_newlines(before) %>
-      <% after = post.normalize_newlines(after) %>
+      <% before = Post.normalize_newlines(before) %>
+      <% after = Post.normalize_newlines(after) %>
       <% diff = Diffy::SplitDiff.new(before, after, format: :html) %>
       <div class="diff-section">
         <div class="diff-old is-changed raw-markdown">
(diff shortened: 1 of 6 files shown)


cellio (approved):
Tested in my dev environment and the diffs are as expected.  I tested on old posts, posts created between the normalization fix and `develop`, and a post newly created while on this branch.  Edit diffs and history diffs are aligned and highlight the expected parts, not the entire post.  Sometimes when you add a paragraph the previous one is highlighted because of the added newline; it doesn't always happen as shown in this screenshot, but this is existing behavior, not a regression.

diff screenshot

app/views/post_history/_diff.html.erb
```diff
@@ -1,7 +1,9 @@
 <div class="diff">
   <% if before.present? && after.present? %>
     <% if before.is_a?(String) && after.is_a?(String) %>
-      <% diff = Diffy::SplitDiff.new(before, after, format: :html, ignore_crlf: true) %>
+      <% before = before.encode(before.encoding, universal_newline: true).strip %>
+      <% after = after.encode(after.encoding, universal_newline: true).strip %>
```
Oaphi: At this point let's make it at least a method on the model, otherwise we are guaranteed to start diverging at some point. Ideally it should be a shared concern, but that can wait

Oaphi (changes requested):

app/views/post_history/_diff.html.erb
trichoplax: I'm very new to concerns, but I've had a go at adding a simple one. It just provides one method to be used for the validation in the post history model, and to allow this view to access it from the post model.

Oaphi (approved):

PR #2133, opened 2026-08-10
Fix recalc_abilities script

Oaphi:
Say hi to our first (rather basic, but still) test for scripts as well!

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -39,3 +39,7 @@ component_management:
       name: tasks
       paths:
         - app/tasks/**
+    - component_id: module_scripts
+      name: scripts
+      paths:
+        - app/scripts/**
--- a/scripts/recalc_abilities.rb
+++ b/scripts/recalc_abilities.rb
@@ -12,4 +12,4 @@ OptionParser.new do |opts|
   end
 end.parse!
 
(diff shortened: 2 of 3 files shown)


cellio (commented):
Before this fix, running the script via `rails runner` produced an error spew.  With the fix, it completes with output like from another script I tried (data dump).  In that sense it's fixed and is in any case a big improvement!

I'm still having a problem where the script doesn't *actually* seem to recalculate abilities, though hitting the "recalc abilities" button for a user directly in the mod tools does work.  I'm guessing there's some difference in the paths taken to the "actually do it" code from the button vs from the script, and therein lurks a problem.

cellio:
> I'm still having a problem where the script doesn't _actually_ seem to recalculate abilities, though hitting the "recalc abilities" button for a user directly in the mod tools does work. I'm guessing there's some difference in the paths taken to the "actually do it" code from the button vs from the script, and therein lurks a problem.

Retracted.  The script runs based on a queue and the mod tool doesn't count as a queuable event.  When I did something to actually force something into the queue, the script worked as expected.  (Filed https://github.com/codidact/qpixel/issues/2134 for the queue/tool mismatch.)

cellio (approved):

ArtOfCode- (approved):

PR #2137, opened 2026-08-12
Fix category seeds getting recreated on every deploy

### LINKED ISSUE #2110: Prevent seeds from creating default categories in existing communities
Pushing updates to a production system includes rerunning seeds to pick up changes.  Our seeds include creating a category named Q&A if it doesn't exist with that exact name.  Some existing communities don't use that name, though.  Sometime in the last few months, we started seeing Q&A appear in the Proposals community.  The first time it happened we deleted it, but the next time we updated prod it came back.  Currently we have worked around the problem by keeping the category but making it visible to staff only, which avoids confusing most of our users but is still a speed bump for admins.

We want to seed Q&A (and Meta) for *new* communities, but if a community already exists, it would be better for updating to skip running these seeds.  Admins can create categories manually if a community wants to add something that was renamed or removed in the past.

I had assumed that the same process that seeds a Q&A category also seeds a Meta category, except that Meta in prod does not have a Meta of its own, even for staff.  But Collab does.  I don't know how we suppressed Meta Meta in a way that stuck, but that's a path to investigate.

Oaphi:
Prevents category seeds from being run if a given community already has categories - this allows new installations to get the default ones while preventing the seeded ones from being constantly recreated if anything about the seeded ones changes.

\+ new `MAILER_HOST` env variable for local development and non-Codidact instances. Matches `MAILER_PROTOCOL`
\+ adds a new admin "categories" tool and merges it with the "create category" one - technically, it's just a proxy to what we already have and simply not expose directly.
\+ "email all users" admin tool is no longer a developer-only tool (we wanted to align it with the actual controller check for a while now - threw it in here for good measure);
\+ fixes a corner-case where non-logged in users are not redirected to signin upon clicking on the post creation button if the category has more than one type to choose from;

closes #2110


ArtOfCode- (approved):

test/application_system_test_case.rb
```diff
@@ -19,6 +19,7 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   driven_by :selenium, using: DRIVER, screen_size: [1920, 1080]
 
   setup do
+    Devise.sign_out_all_scopes
```
Oaphi, line 22: This should hopefully solve the intermittent "you are already signed in" issue with system tests. cc @trichoplax

cellio (approved):
Tested almost everything and LGTM.  I can't tell where the MAILER_HOST change shows up, so I didn't verify that (but this is at worst no worse than before, so not blocking).

Issue #1545, opened 2025-02-27
Redux: Posting sometimes doesn't clear the saved post


cellio:
meta:293215

https://github.com/codidact/qpixel/issues/538 seems to be happening again, non-deterministically.  Sometimes a user posts an answer successfully, but the draft of that answer is still in the answer box on the page.  The problem has been seen in multiple browsers by multiple people.

We don't want to discard drafts prematurely of course, but *after* the post is made it would be safe to discard the draft.  Maybe we have an ND timing issue here?

trichoplax:
[Saved text not cleared when posted](https://meta.codidact.com/posts/293948)

This is a new report, stating that this now happens pretty much every time. I'm seeing the same myself, so I'm linking to this in case it affects priority.

trichoplax:
In searching for this issue, I spotted other Meta posts not yet linked from here:
- [Text entry field for answer stays open (once) after submitting](https://meta.codidact.com/posts/288674)
- [Going back in browser encourages accidental duplicate posting.](https://meta.codidact.com/posts/290845) (probably not the same root cause but similar experience for the users commenting there)
- [Question posted twice despite following the question posting process only once](https://meta.codidact.com/posts/293638)

trichoplax:
I also see the draft of my previous question when trying to post a new question. Highlighting this because this was initially raised as solely an answer problem. I see this often.

PR #2138, opened 2026-08-14
Spam protection tools

ArtOfCode-:
* [x] Automatically add spammers to STAT on deletion (with exceptions so we don't automatically block _every_ deleted account).
 * [x] IP (v6?) CIDR range flagging?
 * [x] Identical post check: block posting a post identical to any of your previous posts for users with <10 non-deleted posts.
 * [x] Potential spammer block: users with <10 non-deleted posts are blocked from posting if there's an active or helpful spam flag against any of their posts.


app/controllers/categories_controller.rb
```diff
@@ -138,12 +138,12 @@ def rss_feed
 
   def post_types
     @post_types = @category.top_level_post_types
-    if @post_types.one?
+    if !user_signed_in?
```
ArtOfCode-, line 141: Apparently one of our system tests was relying on this redirecting, which was order-dependent, so I've moved it to always redirect when not signed in.

test/controllers/concerns/users/user_test_helpers.rb
```diff
@@ -0,0 +1,34 @@
+module UserTestHelpers
```
ArtOfCode-, line 1: This whole file is just the existing helpers extracted from users_controller_test.rb to allow for a restructure.

test/controllers/users/user_mod_tools_test.rb
```diff
@@ -0,0 +1,252 @@
+require 'test_helper'
```
ArtOfCode-, line 1: This file is the existing tests relating to user mod tools extracted into their own file, because the users_controller_test.rb was getting excessively long.

cellio:
A couple initial questions (haven't started testing/reviewing yet):

- How and when does the user find out about the block on identical posts?  (I don't see any view changes in this PR.)
- Identical posts across the network, right?
- For the potential spammer block, what feedback does the user get?  If a legit user got caught in it, would there be a clear path to appeal the block?
- We have a bug where sometimes state doesn't get cleared when a post is made, and it still shows as you having a pending draft.  Do we have to worry about interactions with that here, or is that a separate issue?  (Want to make sure that wouldn't count as a dupe.)  I've never been able to reproduce that bug in my dev environment, so I don't know how to test it.

ArtOfCode-:
> How and when does the user find out about the block on identical posts? (I don't see any view changes in this PR.)

They don't find out _specifically_ what blocked the post - in line with the obscurity in the rest of our spam tools, we spit out a ["useful" error message](https://github.com/codidact/qpixel/blob/develop/app/models/application_record.rb#L62) and leave it at that. I don't see many (any?) legitimate reasons for a new user to post _identical_ post content either multiple times on one site, or several times across the network, so I'm reasonably comfortable with that.

There's more of a chance of an established user/mod/admin posting identical posts across the network (I think we've done it as staff), but anyone with >=10 (configurable) non-deleted posts network-wide is exempt from this completely.

> Identical posts across the network, right?

Yes.

> For the potential spammer block, what feedback does the user get? If a legit user got caught in it, would there be a clear path to appeal the block?

Same feedback - a "useful" error message. They also wouldn't be able to post on Meta so would have to contact support for help. However:

* **Helpful** spam flags only block posting for 7 (configurable) days, after which they're considered expired for this purpose, for exactly this reason. 7 days might be too long - 1 would probably be enough, I just wanted a period where the flag could be marked as helpful and _not_ drop the blocking effect, because...
* **Active** i.e. unreviewed spam flags block posting indefinitely, but **only** until they're reviewed.

The theory is that a potential spammer would be blocked from posting until a moderator reviews the spam flag. On review, they'd then mark the flag as helpful and nuke the user, obviating the need for a post block, and the new deletion job will apply a STAT block. A legitimate user who was wrongly spam-flagged would also be blocked until review, but would then be released once the flag is declined.

Likewise, users with >=10 (configurable) non-deleted posts network-wide are completely exempt.

> We have a bug where sometimes state doesn't get cleared when a post is made, and it still shows as you having a pending draft. Do we have to worry about interactions with that here, or is that a separate issue?

Pending drafts aren't saved as posts (they're actually not in the DB at all, but in Redis), so won't interact with this at all unless the user tries to post it a second time without modification - which would be rejected.

cellio:
That all sounds good to me.  Thanks for explaining!

cellio:
Duplicate posts:

I tested with a user who had exactly 10 posts, 3 of them deleted.  I copied the content of an existing post and was able to repost it.  (Tried with both questions and answers.)

I then tried with a user with no posts -- added an answer (ok), then cut/pasted that on a different question and got the error as expected.  Does this mean we're only checking against posts created after this change is in place?  (That's fine, but I want to check that it's intentional and not a bug.)

I tested duplicate posts across communities; worked as expected.  Adding a carriage return was enough to make the posts non-identical (post succeeded); do we want to normalize or are we not worried about this case?

--

Spam flags:

Tested with a user with < 10 posts and an active spam flag, both local community and elsewhere on the network.  Both produced errors as expected, and declining the flag enabled posting again.  Note: I tested this with the same user from my first paragraph: the flag block worked but the duplicate content block didn't.  Are we accidentally counting deleted posts in one place but not the other?

Tested with a user with > 10 posts and an active spam flag.  Posting permitted, as expected.

--

Error message:

The actual error message looks like one of our 418s -- "The heisenbug vanished when we opened the logs. Please look away and try again."  I hadn't followed your "useful" link before; while your comment seemed fine when I read it, seeing the cryptic error on the post page "live" feels different.  I'd be more comfortable with a vague error than a cryptic one, especially in the case of spam flags because sometimes people are wrong about that.  Can we just say something like "post rejected"?  A real person would then come to meta to ask, and meanwhile a spammer doesn't get useful information.

--

Deleted spammers:

I'm a little afraid to IP-block myself on my dev environment, so I haven't tested deleting spammers.

app/jobs/delete_user_job.rb
```diff
@@ -0,0 +1,28 @@
...
+    if perform_spam_check
+      # Can't use model helper methods very easily here, because we want network-wide flags and that doesn't play
+      # nicely with default scopes.
+      flag_query = Post.unscoped
+                       .joins(Arel.sql("INNER JOIN flags ON flags.post_type = 'Post' AND flags.post_id = posts.id"))
+                       .joins(Arel.sql('INNER JOIN post_flag_types ON flags.post_flag_type_id = post_flag_types.id'))
+                       .where(flags: { status: 'helpful' },
+                              post_flag_types: { name: "it's spam" },
+                              posts: { user_id: user.id })
+      if flag_query.any?
```
cellio, line 21: I don't see any qualifiers here for productivity of the account or age of the flags.  If somebody had a validated spam flag years ago but wasn't deleted (e.g. borderline case fixed with an edit), and that user went on to be productive, and then much later requested account deletion, would we be tarnishing that user's IP address (household, university, whatever)?  I'm not yet sure how this code is reached in the first place; if this is for the "suspected spammers" deletion page then that can't happen and this is fine, but I don't have all the context so I'm asking.

app/models/concerns/post_creation_validations.rb
```diff
@@ -0,0 +1,81 @@
...
+  # rubocop:disable Metrics/BlockLength
+  included do
+    validate :no_mathjax_in_title, on: :create
+    validate :post_type_requires_parent, on: :create
+    validate :post_type_has_category, on: :create
+    validate :can_post_in_category, on: :create
+    validate :identical_post_spam, on: :create
+    validate :no_active_spam_flags, on: :create
+
+    after_create :escalate_suspicious_cidr
```
cellio, line 13: After creating what?  I mean, instances of this class, I guess, but I'm not clear on how concerns work and most of these validations, like mathjax titles, are not about things where we'd consider IP blocks.  I'm a little lost here.

app/models/user.rb
```diff
@@ -453,7 +453,7 @@ def send_welcome_tour_message
                         'how this site works.', '/tour')
   end
 
-  def block(reason, length: 180.days, automatic: true)
+  def block(reason, length: 10.years, automatic: true)
```
cellio, line 456: Should this duration be a setting like the ones in other files?  (I realize it was already hard-coded and you've just changed the number, so this is definitely not a blocker just a question.)

config/config/server_settings.yml
```diff
@@ -1,4 +1,11 @@
-registration_rate_limit: 300
+# Minimum number of seconds between registration attempts from the same IP.
+registration_rate_limit: 3600
```
cellio, line 2: Any way to easily override this in a dev environment?  (Without maintaining local changes to a file in source control, I mean?)  I sometimes need batches of users and hit the existing 5-minute limit...

trichoplax:
> Pending drafts aren't saved as posts (they're actually not in the DB at all, but in Redis), so won't interact with this at all unless the user tries to post it a second time without modification - which would be rejected.

Does "rejected" here mean that an accidental second posting will not work, leaving the user unharmed, or that they will be flagged and blocked from posting anything?

Some users on seeing their draft question or answer still present after posting (due to #1545) assume that posting failed, and post again. Previously this has led to them noticing the unintended duplicate soon after and raising it on Meta (or commenting under an existing Meta post). Will this still be possible with this pull request or will such a user be blocked from posting on Meta?

I can imagine a new user caught by this might not realise that emailing support is even an option.

app/jobs/delete_user_job.rb
ArtOfCode-: There's a threshold of 100 rep network-wide that disables this check completely.

app/models/concerns/post_creation_validations.rb
ArtOfCode-: Concerns are just extracted blocks of related code that can be mixed in to other classes. This one is mixed into the Post class, so the validations and `after_create`	are applied to `Post`. I just extracted the other validations we do on create from the controller and moved them here as well.

app/models/user.rb
ArtOfCode-: Not quite that easy - we'd still need a default here because it forms part of the method signature, so removing it would mean going through all callers and ensuring they set the parameter themselves.

ArtOfCode-:
> Does "rejected" here mean that an accidental second posting will not work, leaving the user unharmed, or that they will be flagged and blocked from posting anything?

@trichoplax the former - the identical post check _just_ blocks that post, and doesn't apply any ongoing consequences to the account.

config/config/server_settings.yml
ArtOfCode-: I've added an override for dev environments.

ArtOfCode-:
@cellio 

> I tested with a user who had exactly 10 posts, 3 of them deleted. I copied the content of an existing post and was able to repost it. (Tried with both questions and answers.)

No repro - I have a user with 10 posts, 2 deleted. I copied a post's Markdown and got the expected error. Did you copy the Markdown or the post's rendered appearance? The latter might have introduced changes in the Markdown, which would mean the post passes the check.

> I'd be more comfortable with a vague error than a cryptic one, especially in the case of spam flags because sometimes people are wrong about that.

Done.

> I'm a little afraid to IP-block myself on my dev environment, so I haven't tested deleting spammers.

If you want to test it, you can always go into the DB manually afterwards to unblock yourself - `DELETE FROM blocked_items;`

app/models/concerns/post_creation_validations.rb
cellio: Ah, after creating a *post* -- now it makes sense, thanks.

cellio:
> No repro - I have a user with 10 posts, 2 deleted. I copied a post's Markdown and got the expected error. Did you copy the Markdown or the post's rendered appearance? The latter might have introduced changes in the Markdown, which would mean the post passes the check.

I did copy the markdown (went into edit to get it), but now I can't repro either with the same user and same post -- maybe I missed a terminal newline and that made the difference?

> > I'd be more comfortable with a vague error than a cryptic one, especially in the case of spam flags because sometimes people are wrong about that.
> 
> Done.

Thank you.

app/models/concerns/post_creation_validations.rb
```diff
@@ -44,7 +44,7 @@ def identical_post_spam
       unless prev_non_deleted_count >= threshold
         identical_posts = Post.unscoped.where(user: user, body_markdown: body_markdown).where.not(id: id)
         if identical_posts.any?
-          errors.add(:base, ApplicationRecord.useful_err_msg.sample)
+          errors.add(:base, I18n.t('posts.spam_blocked'))
```
cellio, line 47: Do we need to update expected output in tests?

cellio (commented):
LGTM. Someone else should review code (particularly the queries and CIDR).  I didn't test CIDR/auto-escalated-flag.

app/controllers/users/registrations_controller.rb
```diff
@@ -8,7 +8,11 @@ class Users::RegistrationsController < Devise::RegistrationsController
   def create
     super do |user|
       unless user.errors.any?
-        rate_limit = AppConfig.server_settings['registration_rate_limit']
+        rate_limit = if Rails.env.development?
+                       0
+                     else
+                       AppConfig.server_settings['registration_rate_limit']
+                     end
```
Oaphi, line 15: Non-blocking, but I'd rather we had a way of overriding app config settings instead of hardcoding this.
Using an env var falling back to the "deffined once" server setting seems reasonable to me here (if we go this route, note that env vars are usually documented in `INSTALLATION.md`):

```ruby
rate_limit = ENV.fetch("REGISTRATION_RATE_LIMIT") { AppConfig.server_settings['registration_rate_limit'] }
```

This way we'll get the best of both worlds: configurability and a default that's guaranteed to be set at runtime.

app/controllers/users_controller.rb
```diff
@@ -342,7 +342,9 @@ def soft_delete
         return
       end
 
-      @user.soft_delete(current_user)
+      max_rep = @user.community_users.maximum(:reputation)
+      check_threshold = AppConfig.spam_protection['deletion_block_max_rep']
+      DeleteUserJob.perform_later(@user, current_user, perform_spam_check: max_rep < check_threshold)
```
Oaphi, line 347: Let's move the threshold check to the job itself - it's not impossible to get the job scheduled when the user is not over the threshold but for it to execute when the user is actually over it. Non-blocking

Oaphi:
@ArtOfCode- I took the liberty of adding the missing tests myself, feel free to mend to your liking. None of the added tests are related to the change directly, just covering reported gaps

PR #2139, opened 2026-08-19
Change editor controls to properly edit block markdown

### LINKED ISSUE #2051: Editor quote button should quote entire selection, not just first line
meta:294961

In the post editor, if you select text and then apply one of the formatting buttons, it formats the selected content (bold, italic, etc).  However, the Quote button doesn't interact well with multi-paragraph selection; it adds a `>` at the beginning but not throughout.  An answer there suggests adding ">" to the beginning of each line, which sounds like it'd work.  (We should test with paragraphs containing embedded returns, as might happen if you draft your post in an external tool and then paste it into the editor.)

MoshiKoi:
Fixes #2051

Block markdown should place the markdown symbols at the start of each line, rather than in the middle, even if that's where the selection is.

```md
line 1

line 2

line 3
```

Blockquoted

```md
> line 1
> 
> line 2
> 
> line 3
```

Numbered list

```md
1. line 1

2. line 2

3. line 3
```

### DIFF
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -1,7 +1,5 @@
 $(() => {
-  /** @typedef {{apply: function(JQuery<HTMLTextAreaElement | HTMLInputElement>): void}} Action */
-
-  /** @implements {Action} */
+  /** @implements {MarkdownAction} */
   class InlineAction {
     constructor(start, end = '') {
       this.start = start;
@@ -20,7 +18,7 @@ $(() => {
     }
   }
 
-  /** @implements {Action} */
+  /** @implements {MarkdownAction} */
   class BlockAction {
     /**
      * @param {function(string, number): string} callback
@@ -78,7 +76,7 @@ $(() => {
   }
 
   /**
-   * @type {{[key: string]: Action}}
+   * @type {{[key: string]: MarkdownAction}}
    */
   const actions = {
     bold: new InlineAction('**', '**'),
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -64,10 +64,12 @@ $(() => {
       for (const line of val.split('\n')) {
         const begin = idx;
         const end = begin + line.length;
-        if (begin <= preSelectionStart && preSelectionStart <= end)
+        if (begin <= preSelectionStart && preSelectionStart <= end) {
           startPos = begin;
-        if (begin <= preSelectionEnd && preSelectionEnd <= end)
+        }
+        if (begin <= preSelectionEnd && preSelectionEnd <= end) {
           endPos = end;
+        }
         // include newline
         idx = end + 1;
       }
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -107,15 +107,13 @@ $(() => {
   $(document).on('click', '.js-markdown-tool', (ev) => {
     const $tgt = $(ev.target);
     const $button = $tgt.is('a') ? $tgt : $tgt.parents('a');
-    const action = $button.attr('data-action');
+    const action = /** @type {string} */ ($button.attr('data-action'));
 
     /** @type {JQuery<HTMLTextAreaElement | HTMLInputElement>} */
     const $field = $('.js-post-field');
 
-    if (action && action in actions) {
-      actions[action].apply($field);
-      $field.trigger("focus");
-    }
+    actions[action].apply($field);
+    $field.trigger("focus");
   });
 
   QPixel.DOM?.addSelectorListener('keypress', '#markdown-link-name, #markdown-link-url', (ev) => {
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -1,6 +1,13 @@
 $(() => {
-  /** @implements {MarkdownAction} */
+  /**
+   * A markdown action which places markings before and after the current selection
+   * @implements {MarkdownAction}
+   */
   class InlineAction {
+    /**
+     * @param {string} start Text to place before the selection
+     * @param {string} end Text to place after the selection
+     */
     constructor(start, end = '') {
       this.start = start;
       this.end = end;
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/markdown.js
```diff
@@ -1,4 +1,96 @@
 $(() => {
+  /** @typedef {{apply: function(JQuery<HTMLTextAreaElement | HTMLInputElement>): void}} Action */
+
+  /** @implements {Action} */
+  class InlineAction {
```
ArtOfCode-: Rather than relying on JSDoc only, for a pure JS approach would it be better to define `Action` explicitly here as an abstract class and have `InlineAction` and `BlockAction` extend it?

app/assets/javascripts/markdown.js
```diff
@@ -1,4 +1,96 @@
...
+    /**
+     * @param {JQuery<HTMLTextAreaElement | HTMLInputElement>} $field 
+     */
+    apply($field) {
+      // Override the selection range so it encompasses full lines
+      const [startPos, endPos] = this.#getBlockSelection($field);
+      $field[0].setSelectionRange(startPos, endPos);
+      const lines = $field.val().substring(startPos, endPos).split('\n');
+      let jdx = 0;
+      for (let idx = 0; idx < lines.length; ++idx) {
```
ArtOfCode-, line 51: Does a `for`-`let` block work here?

app/assets/javascripts/markdown.js
```diff
@@ -1,4 +1,96 @@
...
+          lines[idx] = this.callback(lines[idx], jdx);
+          ++jdx;
+        }
+      }
+      $field[0].setRangeText(lines.join('\n'));
+    }
+
+
+    /**
+     * @param {JQuery<HTMLTextAreaElement | HTMLInputElement>} $field
```
ArtOfCode-, line 65: It would be useful to include some documentation here about what this function is intended to do. What are the numbers it returns?

app/assets/javascripts/markdown.js
```diff
@@ -1,4 +1,96 @@
...
+      const preSelectionEnd = $field[0].selectionEnd;
+      const val = $field.val();
+      // Not particularly efficient, but seems to work fine
+      let startPos, endPos;
+      let idx = 0;
+      for (const line of val.split('\n')) {
+        const begin = idx;
+        const end = begin + line.length;
+        if (begin <= preSelectionStart && preSelectionStart <= end)
+          startPos = begin;
```
ArtOfCode-:
```suggestion
        if (begin <= preSelectionStart && preSelectionStart <= end) {
          startPos = begin;
        }
```

app/assets/javascripts/markdown.js
```diff
@@ -1,4 +1,96 @@
...
+        if (begin <= preSelectionEnd && preSelectionEnd <= end)
+          endPos = end;
```
ArtOfCode-:
```suggestion
        if (begin <= preSelectionEnd && preSelectionEnd <= end) {
          endPos = end;
        }
```

ArtOfCode- (changes requested):

app/assets/javascripts/markdown.js
Oaphi: @ArtOfCode- not sure we need a base class here, tbh? Doesn't seem like they have any shared implementation, even when it comes to the constructor. Given that the only common thing about them is that they both have the `apply` method, `implements Action` (and, in our case, its JSDoc equivalent) seems reasonable to me.

One thing I'd ask to change (that one is non-blocking, just a request), though, @MoshiKoi: can you move the definition to the `globals.d.ts` declaration file? I've been collecting our assets' definitions there for a while now with the aim of transitioning to a proper TypeScript-based build pipeline for assets.

app/assets/javascripts/markdown.js
```diff
@@ -7,25 +99,9 @@ $(() => {
...
-      mathjax: ['$', '$'],
-    };
-
-    if (Object.keys(actions).indexOf(action) !== -1) {
-      const preSelection = [$field[0].selectionStart, $field[0].selectionEnd];
-      QPixel.MD.insertIntoField($field, actions[action][0], actions[action][1]);
-      $field.focus();
-      $field[0].selectionStart = preSelection[0] + actions[action][0].length;
-      $field[0].selectionEnd = preSelection[1] + actions[action][0].length;
+    if (action && action in actions) {
```
Oaphi: Let's just do `actions[action]?.apply($field)` - we can safely use optional chaining these days, and `actions[action]` with a non-existent `action` will do the same thing as `in` (that said, `in` is definitely an improvement over `Object.keys(actions).indexOf(action) !== -1`) by trying to find it in the `actions` prototype chain.

We could also be a bit more explicit here and do (just an example):

```js
const handler = actions[action]

if (handler) {
  handler.apply($field);
  $field.trigger("focus");
}
```

Oaphi (changes requested):

app/assets/javascripts/markdown.js
Oaphi: Doesn't seem like we should be worried about untrusted input in `data-action` here but do correct me if I am wrong and we should (the worst thing that could happen here is a self-inflicted uncaught error).

app/assets/javascripts/markdown.js
MoshiKoi: @ArtOfCode- JS doesn't have abstract classes, so it'd really just be an empty implementation

@Oaphi Sound good to me

app/assets/javascripts/markdown.js
MoshiKoi: That's true, `action` or `actions[action]` being null shouldn't actually occur. I might as well remove the check; I was slightly unsure because my type checker was complaining.

app/assets/javascripts/markdown.js
MoshiKoi: No, (assuming you meant `for (let line of lines)`. The way it works is that it has to actually reassign the lines. A range for does not do reassignment like that; I can't write `line = this.callback(line)` to change the actual array item.

app/assets/javascripts/markdown.js
Oaphi: Was it complaining about it being nullable or `action` being unusable to index the `actions` map, btw?

MoshiKoi: The latter - the former can be suppressed with optional chaining like you suggested.

Oaphi: Ah, yeah, quick guess is that it inferred the `actions` map's type as a constant. Widening the type to `Record<string, Action>` to allow `string` to be used to index it should alleviate the issue, I believe?

MoshiKoi: Opted to type cast it. If it somehow is null at runtime, this means the buttons are hooked up incorrectly and it should just fail instead of silently ignoring the error.

MoshiKoi:
Closing in favor of #2140 for CI purposes

Issue #1993, opened 2026-02-23
On dashboard, collapse uninteresting communities


cellio:
meta:294770

This is a narrower change while waiting for https://github.com/codidact/qpixel/issues/1323.  The dashboard takes a lot of space to show a community that the reader might not be interested in (full list of categories for each, plus logo).  Order can't currently be changed (it's by age), but can we add a way for users to "unpin" (collapse and deprioritize) ones they don't care about *and persist that state*?  I'm imagining a "collapse" icon next to each entry, which collapses that community to a single label with the name and moves it to the bottom of the list.  It's ok if the move doesn't happen until the next page load.  Collapsed communities should have an icon to do the reverse -- expand and restore to its original position (move can wait for reload).

On page load, we'd show all of the expanded (normal state) communities in their current order followed by the collapsed views of all the unpinned ones, again in deterministic order.  Ultimately we want user-controlled ordering that applies everywhere communities are listed, so this is a stopgap -- but because the dashboard entries are large particularly on a phone, I'd like to do some sort of stopgap if we can.

trichoplax:
Before someone starts work on this, I'd like to see a complexity assessment of both this and #1323. It isn't obvious to me that this would be significantly quicker or easier than implementing ordering.

trichoplax:
It might be worth considering whether this change would still be wanted after #1323 is implemented. If so, it's worth doing this one first, but if not, the additional work might be better spent on going straight to implementing ordering.

cellio:
Thanks for raising that.  I had been assuming that 1323 (the end goal) would be a lot harder, but if not, there's no point in doing this interim work.  Let's assess 1323 and then decide what to do with this one.

Issue #1323, opened 2024-03-03
Customize order of communities in community switcher


cellio:
Depends on: https://github.com/codidact/qpixel/issues/284 .

After we expose a network profile, allow users to change the order of communities on their own profiles, and show communities in that order both on the network profile and in the community switcher.  This would allow users to put the communities they care about most early in the list, easing navigation.

This would address some of the requests in meta:290090.

cellio:
A basic network profile is now in place, unblocking this.

MoshiKoi:
Would it be better to put this in the network profile tab or in preferences? (We already have a "Network Preferences" heading there)

cellio:
We'll have to figure out what the UI should look like.  Should the network profile page have controls to reorder communities, either drag/drop or reordering arrows?  Were you thinking of something different that would fit in with preferences @MoshiKoi ?

MoshiKoi:
I wasn't really thinking about the control itself, more so where a user would expect to find such a setting. To me, it feels more natural to add such a customization option to the Preferences tab, where we already centralize "things you can change about the site". 

That being said, if you're worried about making the control "fit in" with preferences, in my opinion, there is no reason to limit preferences to the simple key-value controls we have currently. That is just a consequence of how we currently architect preferences, not a reason to split up different settings into different tabs.

cellio:
If we can make it work somehow on the preferences page, I agree that's a more logical place for people to look for it.  Any thoughts on what would be a good UI for this?

MoshiKoi:
Drag and drop or reordering arrows is probably the most common UI for a reorderable list, perhaps with a checkbox or something to hide those a user isn't interested in (#1993).

PR #2143, opened 2026-08-21
Network preference for the order of communities in the dashboard and dropdown

MoshiKoi:
First take at #1323, testing and feedback welcome

This PR adds a new preference type for the community preference, as well as a separate "block preference" for styling purposes since it didn't make much sense trying to cram it into the horizontal layout the other preferences use. The styling is very much a work in progress, but it is functional right now.

image

You can reorder the communities by dragging. ~~I pulled in JQuery-UI for this - not entirely sure if it's worth bringing it in as a dependency, but it can be replaced if needed.~~ I have switch to SortableJS to avoid depending on jQuery.

Might try to also implement "collapsed" communities (#1993). Not sure if that has any real added value if it just sends communities to the bottom since users can do that anyway, but it might be useful as a foundation for later work if we want users to, for instance, be able to set up an email subscription of "posts from focused communities" or similar mechanism.


trichoplax:
I like the draggable handles. I think for many people that will be the easiest way to rearrange the communities. Could we keep the handles and also add up and down buttons (that shift that community one place in that direction)?

Some people may find up and down buttons more convenient, but also some people cannot drag (for various reasons).

See [Dragging Movements](https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements) accessibility guidance.

If both draggable handles and up and down buttons take up too much space, there's also the possibility of having up and down buttons only appear when a draggable handle is clicked/tapped rather than dragged. Personally I'd prefer up and down buttons that are always visible to aid discoverability.

PR #2148, opened 2026-08-23
Add a blank line after generated <summary> tag

### LINKED ISSUE #2144: Insert blank line in editor tools spoiler text
Following #2142 the post editor toolbar allows inserting a `<details>` section for spoiler text. This will work fine for plain text but will fail to render Markdown inside it, as described in [Details tag requires an internal blank line to render markdown](https://meta.codidact.com/posts/287279).

The text added by the button could include the blank line automatically to reduce the risk of the user running into this problem. The blank line will have no effect on the rendered result.

MoshiKoi:
Closes #2144

### DIFF
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -102,7 +102,7 @@ $(() => {
     hr: new InlineAction('\n\n-----\n\n'),
     table: new InlineAction('\n\n| Title1 | Title2 |\n|- | - |\n| row1_1 | row1_2 |\n\n'),
     mathjax: new InlineAction('$', '$'),
-    spoiler: new InlineAction('\n\n<details>\n<summary>Spoiler</summary>\n', '\n</details>\n\n'),
+    spoiler: new InlineAction('\n\n<details>\n<summary>Spoiler</summary>\n\n', '\n</details>\n\n'),
   };
 
   $(document).on('click', '.js-markdown-tool', (ev) => {
--- a/app/assets/javascripts/markdown.js
+++ b/app/assets/javascripts/markdown.js
@@ -102,7 +102,11 @@ $(() => {
     hr: new InlineAction('\n\n-----\n\n'),
     table: new InlineAction('\n\n| Title1 | Title2 |\n|- | - |\n| row1_1 | row1_2 |\n\n'),
     mathjax: new InlineAction('$', '$'),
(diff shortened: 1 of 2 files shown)


Oaphi (changes requested):
Let's add an explanatory comment before this empty line (wording is just an example) - I can guarantee that users will just start typing over that empty line or even delete it outright:

```html
<details>
<summary>Spoiler</summary>
<!--
An empty line below is necessary for Markdown to render.
This HTML comment is a hint and can be safely deleted
-->

</details>
```

MoshiKoi:
Sounds good - I was on the fence about it since we've already had complaints about Codidact being too explicit and in the way, but considering that you won't ever need to actually delete the comment, I guess it's fine here.

Oaphi:
I was (initially) thinking of just automatically stripping it out on submission, but that's probably too much work for little to no benefit

Oaphi (approved):

trichoplax:
> I was (initially) thinking of just automatically stripping it out on submission, but that's probably too much work for little to no benefit

I like the idea of it staying in for future editors. It might even be worth omitting "can be safely deleted" to avoid encouraging this.

cellio (commented):
Tested and LGTM.  I agree with the earlier suggestion to shorten the comment.

Question: using this puts the cursor after the `<` in `<spoiler>`, which is (a) not a good place and (b) peculiar.  If it had been *before* it, I would have thought that that's just how these tools go.  But if I use the bold tool, it puts the cursor between the `**` and `**`, right where you want to start typing.  Can we put the cursor at the place where you start typing the hidden text -- after the newline, on the line before `</details>`?

cellio:
The tooltip on the tools button says "headings, lines" -- let's add ", spoilers".

Oaphi:
> The tooltip on the tools button says "headings, lines" -- let's add ", spoilers".

@cellio if we are doing this in the PR, let's make the button's `title` list-agnostic and just say `Markdown tools` (or something like that). Every item in the modal has its own `title` and `aria-label`, so explicitly listing the tools in the tooltip is just unnecessary maintenance burden.

cellio:
> @cellio if we are doing this in the PR, let's make the button's `title` list-agnostic and just say `Markdown tools` (or something like that). Every item in the modal has its own `title` and `aria-label`, so explicitly listing the tools in the tooltip is just unnecessary maintenance burden.

That's fine too.  I just want to avoid the inconsistency of listing two of three items in the tooltip; I don't care which direction we go to fix it.

cellio (approved):
LGTM now.

PR #2154, opened 2026-09-06
General improvements for donations handling

Oaphi:
- `:intent` with `:amount` less than 0.5 no longer causes a runtime error due to a typo in a local variable name (`@symbol`);
- `return_to` parameter is now dropped unless it's an HTTP, HTTP(S), or relative URI (security);
- `:intent` and `:success ` have tests now;

If you want to test donations manually, you'll only need an API key for a [sandbox](https://docs.stripe.com/sandboxes).

**Do not merge unless at least one requested reviewer manually confirms that the full flow works.**

### DIFF
--- a/app/controllers/donations_controller.rb
+++ b/app/controllers/donations_controller.rb
@@ -30,8 +30,10 @@ class DonationsController < ApplicationController
 
     # amount * 100 because Stripe takes amounts in pence
     @amount = amount
-    @intent = Stripe::PaymentIntent.create({ amount: (amount * 100).to_i, currency: @currency,
-                                             metadata: { user_id: current_user&.id }, description: params[:desc] },
+    @intent = Stripe::PaymentIntent.create({ amount: (amount * 100).to_i,
+                                             currency: @currency,
+                                             metadata: { user_id: current_user&.id },
+                                             description: params[:desc] },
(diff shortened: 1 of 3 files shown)


app/controllers/donations_controller.rb
```diff
@@ -22,28 +24,34 @@ def intent
     end
 
     if amount < 0.50
-      flash[:danger] = "Sorry, we can't accept amounts below #{symbol}0.50. We appreciate your generosity, but the " \
+      flash[:danger] = "Sorry, we can't accept amounts below #{@symbol}0.50. We appreciate your generosity, but the " \
```
cellio, line 27: I haven't been able to trigger this message; an amount below 0.5 gets intercepted in the view with a message that just says "can't be below 0.5" without the explanation.  This PR doesn't change that behavior, but I just wanted to note that in case it's testing error on my part and I should be able to make this show up.

cellio (commented):
I set up a Stripe sandbox.  Donations above the minimum go through (correct UI for us, shows up in Stripe, sends email).  Donations below the minimum are caught in the form so they're never submitted.  I don't yet know what we do with webhooks or how to test that.

ArtOfCode- (approved):

PR #2159, opened 2026-09-18
Prevent anonymous users from viewing abilities for specific users

Oaphi:
As I still believe in a future where we will no longer have to resort to this, the restriction is configurable as a new site setting called `RequireSignInToViewUserAbilities`, even if, by default, anonymous access is now blocked.

As a bonus, the PR allows seeds to update categories for exisitng site settings - I wanted to move `RestrictedAccess*` settings to the same category as the new one:

req_sign_in_to_view_user_ab_site_setting


cellio (approved):
LGTM.  I tested {setting true, setting false} x {signed in, anonymous} and all worked as expected.  Also confirmed that the other settings have moved to the new category (thanks for that).

